Issue #187, opened 2015-01-13
Support pre-push


asottile:
(Maybe generically support more hooks as well)

This seems pretty straightforward to implement. Should be able to read the spec on pre-push and use the revision information to run the pre-commit hooks.

We'll need to extend run, install, and uninstall to support this as well.

guykisel:
You're going to have to rename the entire project now.

asottile:
lol

asottile:
#189

shihoonhootsuite:
Was there documentation on how to enable pre-push or is this just enabled by default?

asottile:
Docs for pre-push are http://pre-commit.com/#advanced  section "pre-commit
during push"

You'll need a sufficiently new version of git (I believe >1.8) which
supports pre-push as a hook

mechastorm:
Thanks!

PR #188, opened 2015-01-13
Make pre_commit also support pre-push book

dongweiming:
finished #187

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -9,7 +9,6 @@ import stat
 import sys
 
 from pre_commit.logging_handler import LoggingHandler
-from pre_commit.util import resource_filename
 
 
 logger = logging.getLogger('pre_commit')
@@ -42,37 +41,55 @@ def make_executable(filename):
     )
 
 
-def install(runner, overwrite=False, hooks=False):
+def get_hook_path(runner, hook_type):
+    if hook_type == 'pre-commit':
+        hook_path = runner.pre_commit_path
+        legacy_path = runner.pre_commit_legacy_path
+    else:
+        hook_path = runner.pre_push_path
+        legacy_path = runner.pre_push_legacy_path
+    return hook_path, legacy_path
+
+
+def install(runner, overwrite=False, hooks=False, hook_type='pre-commit'):
     """Install the pre-commit hooks."""
-    pre_commit_file = resource_filename('pre-commit-hook')
+    hook_path, legacy_path = get_hook_path(runner, hook_type)
 
     # If we have an existing hook, move it to pre-commit.legacy
     if (
-        os.path.exists(runner.pre_commit_path) and
-        not is_our_pre_commit(runner.pre_commit_path) and
-        not is_previous_pre_commit(runner.pre_commit_path)
+        os.path.exists(hook_path) and
+        not is_our_pre_commit(hook_path) and
+        not is_previous_pre_commit(hook_path)
     ):
-        os.rename(runner.pre_commit_path, runner.pre_commit_legacy_path)
+        os.rename(hook_path, legacy_path)
 
     # If we specify overwrite, we simply delete the legacy file
-    if overwrite and os.path.exists(runner.pre_commit_legacy_path):
-        os.remove(runner.pre_commit_legacy_path)
-    elif os.path.exists(runner.pre_commit_legacy_path):
+    if overwrite and os.path.exists(legacy_path):
+        os.remove(legacy_path)
+    elif os.path.exists(legacy_path):
         print(
             'Running in migration mode with existing hooks at {0}\n'
             'Use -f to use only pre-commit.'.format(
-                runner.pre_commit_legacy_path,
+                legacy_path,
             )
         )
 
-    with io.open(runner.pre_commit_path, 'w') as pre_commit_file_obj:
-        contents = io.open(pre_commit_file).read().format(
+    with io.open(hook_path, 'w') as pre_commit_file_obj:
+        if hook_type == 'pre-push':
+            with io.open(runner.pre_push_template) as fp:
+                pre_push_contents = fp.read()
+        else:
+            pre_push_contents = ''
+
+        contents = io.open(runner.pre_template).read().format(
             sys_executable=sys.executable,
+            hook_type=hook_type,
+            pre_push=pre_push_contents,
         )
         pre_commit_file_obj.write(contents)
-    make_executable(runner.pre_commit_path)
+    make_executable(hook_path)
 
-    print('pre-commit installed at {0}'.format(runner.pre_commit_path))
+    print('pre-commit installed at {0}'.format(hook_path))
 
     # If they requested we install all of the hooks, do so.
     if hooks:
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
--   repo: git@github.com:pre-commit/pre-commit-hooks
```
dongweiming: Do not use git@github.com:xx/xx. we have no permission to access this address

asottile: You should set up ssh keys!

pre_commit/commands/install_uninstall.py
```diff
@@ -97,7 +97,7 @@ def uninstall(runner):
         return 0
 
     os.remove(runner.pre_commit_path)
-    print('pre-commit uninstalled')
+    print('pre-commit: the {0} hook is uninstalled'.format(hook_type))
```
asottile: I think this can more simply say:

```
print('{0} uninstalled'.format(hook_type))
```

pre_commit/commands/run.py
```diff
@@ -48,8 +49,21 @@ def _print_user_skipped(hook, write, args):
     ))
 
 
+def get_changed_files(new, old):
+    changed_files = cmd_output(
+        'git', 'show', '--pretty=format:', '--name-only',
+        '{0}..{1}'.format(old, new))[1].splitlines()
```
asottile: we prefer this style when making multiline invocations:

```
func(
    params...,
)
```

So this one would look like:

```
changed_files = cmd_output(
    'git', 'diff', '--name-only', '{0}..{1}'.format(old, new),
)[0].splitlines()
```

pre_commit/commands/run.py
```diff
@@ -48,8 +49,21 @@ def _print_user_skipped(hook, write, args):
...
+    for f in changed_files:
+        if not f:
+            changed_files.remove(f)
```
asottile: why is this branch necessary?

pre_commit/main.py
```diff
@@ -44,8 +44,26 @@ def main(argv=None):
             'in the config file.'
         ),
     )
+    install_parser.add_argument(
+        '-t', '--hook-type', choices=('pre-commit', 'pre-push'),
+        default='pre-commit',
+        help=(
```
asottile: I think you can remove the help message here, argparse basically duplicates what you've said here

pre_commit/main.py
```diff
@@ -44,8 +44,26 @@ def main(argv=None):
...
+    )
 
-    subparsers.add_parser('uninstall', help='Uninstall the pre-commit script.')
+    uninstall_parser = subparsers.add_parser(
+        'uninstall', help='Uninstall the pre-commit script.',
+    )
+    uninstall_parser.add_argument(
+        '-t', '--hook-type', choices=('pre-commit', 'pre-push'),
+        default='pre-commit',
+        help=(
```
asottile: same here

pre_commit/main.py
```diff
@@ -67,6 +85,14 @@ def main(argv=None):
...
         '--verbose', '-v', action='store_true', default=False,
     )
+    run_parser.add_argument(
+        '--origin', '-o', default='',
+        help='The origin branch/commit_id when use `git push`',
+    )
+    run_parser.add_argument(
+        '--source', '-s', default='',
+        help='The remote branch/commit_id when use `git push`',
+    )
```
asottile: I think there's some way in argparse to require both or neither here, would be nice to have that

pre_commit/main.py
```diff
@@ -67,6 +85,14 @@ def main(argv=None):
     run_parser.add_argument(
         '--verbose', '-v', action='store_true', default=False,
     )
+    run_parser.add_argument(
+        '--origin', '-o', default='',
+        help='The origin branch/commit_id when use `git push`',
+    )
+    run_parser.add_argument(
+        '--source', '-s', default='',
+        help='The remote branch/commit_id when use `git push`',
```
asottile: s/use/using/

pre_commit/main.py
```diff
@@ -67,6 +85,14 @@ def main(argv=None):
     run_parser.add_argument(
         '--verbose', '-v', action='store_true', default=False,
     )
+    run_parser.add_argument(
+        '--origin', '-o', default='',
+        help='The origin branch/commit_id when use `git push`',
+    )
+    run_parser.add_argument(
+        '--source', '-s', default='',
```
asottile: I think `default` is unnecessary here

pre_commit/resources/pre-push-hook
```diff
@@ -0,0 +1,69 @@
...
+ENV_PYTHON_RETV=$?
+python -c 'import pre_commit.main' >& /dev/null
+PYTHON_RETV=$?
+
+
+if ((
+            (WHICH_RETV != 0) &&
+            (ENV_PYTHON_RETV != 0) &&
+            (PYTHON_RETV != 0)
+            )); then
```
asottile: this doesn't match `pre-commit-hook`, I wonder if there's a way we could combine the duplicated code somehow

pre_commit/resources/pre-push-hook
```diff
@@ -0,0 +1,69 @@
...
+    echo '`pre-commit` not found.  Did you forget to activate your virtualenv?'
+    exit 1
+fi
+
+
+# Run the legacy pre-commit if it exists
+if [ -x "$HERE"/pre-commit.legacy ]; then
```
asottile: hmmm, we should probably handle the legacy scripts as well for pre-push

pre_commit/runner.py
```diff
@@ -17,18 +17,19 @@ class Runner(object):
     repository under test.
     """
 
-    def __init__(self, git_root):
+    def __init__(self, git_root, hook_type='pre-commit'):
```
asottile: I'd rather we not add another optional argument to this class and instead change the property below to a function that takes the hook type.

Or better yet, make two properties, one for pre_commit_path, one for pre_push_path

tests/commands/run_test.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: UTF-8 -*-
+# -*- coding: utf-8 -*-
```
asottile: please revert this line

UTF-8 is actually marginally faster at parsing than utf-8

asottile:
Overall seems pretty good.

Please add some integration tests for pre-push hooks, and restore coverage to 100% :)

tests/commands/run_test.py
dongweiming: Can you tell me the banchmark url, or some reason?

pre_commit/main.py
dongweiming: I think this is necessary, because there are other options. they are not required

tests/commands/run_test.py
asottile: I forget where I saw the benchmark but the basic reason is when looking up the codecs it tries the value first and then normalizes it, the speed difference is negligible, but changing this line is unnecessary to your patch

pre_commit/main.py
asottile: without default you'll get `None` which follows the same codepath

pre_commit/commands/install_uninstall.py
```diff
@@ -42,37 +41,55 @@ def make_executable(filename):
...
 
 
-def install(runner, overwrite=False, hooks=False):
+def get_hook_path(runner, hook_type):
+    if hook_type == 'pre-commit':
+        hook_path = runner.pre_commit_path
+        legacy_path = runner.pre_commit_legacy_path
+    else:
+        hook_path = runner.pre_push_path
+        legacy_path = runner.pre_push_legacy_path
```
asottile: might be able to simplify this since legacy path will always be hook_path + '.legacy'

pre_commit/runner.py
```diff
@@ -43,17 +44,39 @@ def repositories(self):
...
+    def pre_push_path(self):
+        return self.get_hook_path('pre-push')
+
+    @cached_property
+    def pre_template(self):
+        return resource_filename('hook-tmpl')
 
     @cached_property
+    def pre_push_template(self):
+        return resource_filename('pre-push-tmpl')
```
asottile: these don't really belong in the runner, I'd just keep these in the place that installs

pre_commit/commands/install_uninstall.py
asottile: hmmm this appears to be existing, you can probably just leave it this way

pre_commit/resources/hook-tmpl
```diff
@@ -7,6 +7,7 @@ HERE=`pwd`
 popd > /dev/null
```
asottile: since we're changing this file, we'll need to bump the magical number at the top.  I usually just md5 setup.py and put that there (you'll need to update the list of these as well)

pre_commit/runner.py
dongweiming: I appreciate your pursuit of the code, especially for the requirement of coverage. But I think you with a lot of personal subjective for ths project. I open the [pre-push](https://github.com/dongweiming/pre-push) is because I feel your can hard accepte others opinion.

These cost me a lot of time. i think you can merge it and modify somethings you want. or you can do this part by yourself. Thank you so much for review!

asottile:
Cool, thanks for the PR :D

I'll write some tests, modify some formatting, and merge it from here.

asottile:
I'll be merging this in #189

mechastorm:
Was there documentation on how to enable pre-push or is this just enabled by default?

dongweiming:
run

    pre-commit install [-t pre-commit]  # pre-commit will now run on every commit.

or

    pre-commit install -t pre-push   # pre-push will now run on every push.

you only need install one of them. and also can uninstall it:

    pre-commit uninstall -t pre-push   # pre-push will now run on every push.

Issue #159, opened 2014-09-02
Won't run on Windows


quentindemetz:
I'm getting the following traceback:
```
(trojan) C:\trojan>Scripts\pre-commit.exe run
Traceback (most recent call last):
  File "C:\trojan\Scripts\pre-commit-script.py", line 9, in <module>
    load_entry_point('pre-commit==0.2.8', 'console_scripts', 'pre-commit')()
  File "C:\trojan\lib\site-packages\pkg_resources.py", line 356, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "C:\trojan\lib\site-packages\pkg_resources.py", line 2431, in load_entry_point
    return ep.load()
  File "C:\trojan\lib\site-packages\pkg_resources.py", line 2147, in load
    ['__name__'])
  File "C:\trojan\lib\site-packages\pre_commit\main.py", line 12, in <module>
    from pre_commit.commands.run import run
  File "C:\trojan\lib\site-packages\pre_commit\commands\run.py", line 11, in <module>
    from pre_commit.output import get_hook_message
  File "C:\trojan\lib\site-packages\pre_commit\output.py", line 11, in <module>
    ['tput', 'cols'], stdout=subprocess.PIPE
  File "C:\Python27\Lib\subprocess.py", line 710, in __init__
    errread, errwrite)
  File "C:\Python27\Lib\subprocess.py", line 958, in _execute_child
    startupinfo)
WindowsError: [Error 2] The system cannot find the file specified
```

asottile:
I started trying to get some of tests to pass here: https://github.com/pre-commit/pre-commit/tree/windoze

The first wall I'm hitting is symlinks, which require elevated privileges by default under windows (wtf).

I might rethink the one place I'm using symlinks and swap in something like sqlite instead (which might actually make everything simpler?).

For xargs, etc.  I think I can depend on powershell (who would develop on windows without powershell amirite?).  Here's some recipes I found from googling: http://five.agency/powershell-tips-for-bash-users/

asottile:
So I don't forget next time: pytest does not handle coloring well under powershell, use this in pytest.ini:

```
[pytest]
addopts = --color=no
```

asottile:
Using git-shell (from github's windows downloads), I have pre-commit running!  A bunch of the tests fail however and I'm not convinced that using github's tools constitutes supporting windows.  But worst-case you can start using some of the hooks.  I'm going to keep working on this branch and get more things passing.

[...]

PR #190, opened 2015-01-19
Use sqlite3 instead of symlinks for managing repositories

asottile:
Necessary for #159


asottile:
Note: as written, an upgrade to this version will not break, but will cause the user to need to reinstall all hooks.  I think this is ok? Could write an upgrade pathway as well... hmmm...

asottile:
#196 includes this

PR #216, opened 2015-03-29
Use sys.executable when executing virtualenv.  Resolves #208.

### LINKED ISSUE #208: pre-commit potentially uses the wrong `virtualenv` when building environments
It should use `sys.executable, '-m', 'virtualenv'` instead of `'virtualenv'`

asottile:


asottile:
This fix is included in 0.4.4

PR #222, opened 2015-05-05
Care less about user installs during test.  Resolves #221

### LINKED ISSUE #221: py.test / tox fails when running tests on a machine where pre-commit is installed
Here is the culprit:
```
$ py.test tests/commands/install_uninstall_test.py -k test_environment_not_sourced                
========================================================================== test session starts ===========================================================================
platform linux2 -- Python 2.7.6 -- py-1.4.26 -- pytest-2.6.4
plugins: ipdb, xdist
collected 24 items 

tests/commands/install_uninstall_test.py F

================================================================================ FAILURES ================================================================================
______________________________________________________________________ test_environment_not_sourced ______________________________________________________________________

tmpdir_factory = <tests.conftest.TmpdirFactory object at 0x7f0a4cfa3890>

    def test_environment_not_sourced(tmpdir_factory):
        path = make_consuming_repo(tmpdir_factory, 'script_hooks_repo')
        with cwd(path):
            # Patch the executable to simulate rming virtualenv
            with mock.patch.object(sys, 'executable', '/bin/false'):
                assert install(Runner(path)) == 0
    
            ret, stdout, stderr = cmd_output(
                'git', 'commit', '--allow-empty', '-m', 'foo',
                env={'HOME': os.path.expanduser('~')},
                retcode=None,
            )
>           assert ret == 1
E           assert 0 == 1

tests/commands/install_uninstall_test.py:170: AssertionError
-------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------
pre-commit installed at /tmp/user/1000/pytest-22/test_environment_not_sourced0/1/.git/hooks/pre-commit
======================================================== 23 tests deselected by '-ktest_environment_not_sourced' =========================================================
================================================================ 1 failed, 23 deselected in 1.31 seconds =================================================================
```

It looks like `mock.patch.object(sys, 'executable', '/bin/false')` is not enough: `pre-commit` can still be found as an executable in the path or a python lib by `.git/hooks/pre-commit`. That's probably because I installed it with with `pip install --user` and not used any virtualenv.

I'd suggest just removing this test as the balance usefulness / complexity to make it pass does not look interesting to me. What kind of regression could it catch anyway ?

asottile:


Lucas-C:
Looks good, great job !
A bit verbose for a single test though :)

Issue #219, opened 2015-05-01
Add support for pcre / scripts / system hooks definition in .pre-commit-config.yaml


Lucas-C:
Everything is in the title :)

_Rationale:_ a `pre-commit` user shouldn't have to setup a git repository to configure a pre-commit check that can be defined in 5 lines or less.

_Example:_ taken the from [do_not_commit](https://github.com/pricematch/pricematch-pre-commit-hooks/blob/master/hooks.yaml) hook:

    - id: do_not_commit
      name: Block if "DO NOT COMMIT" is found
      entry: DO NOT COMMIT
      language: pcre
      files: ^(.*)$
 
_Suggested solutions:_

1. Allow for pcre / scripts / system hooks definition in _.pre-commit-config.yaml_
2. Allow for the `repo` field in _.pre-commit-config.yaml_ to point to a subfolder (of the git repo configured with `pre-commit`) that contains a _hooks.yaml_ .
This currently crashes because `pre-commit` expect to find git repository in the folder point by `repo`.

asottile:
I think this is along similar lines of this: https://github.com/pre-commit/pre-commit/issues/173

And is kind-of possible: https://github.com/Yelp/venv-update/blob/a5960acab7101a1e70c57945b2038fef9d005aed/.pre-commit-config.yaml#L15-L22

I'd certainly like to make this easier though :)

Lucas-C:
Yes ! The `Yelp/venv-update` example is exactly what I wanted !
If you support this, maybe mention it in the docs ?
Although maybe allowing a `repo: local` value would be cleaner than having to point it to an existing dummy github repo.

asottile:
Yeah that way *works* currently, I wouldn't call it supported yet since it's kinda clunky and I want a more straightforward way of doing that.  That method is unlikely to break though.  I imagine the actual implementation will be similar

Lucas-C:
Ok. Then you can resolve this ticket if you want to track this feature elsewhere.
And thank you for your detailed answers !

Lucas-C:
I went on and hacked around it: what do you think of the solution described in #223 ?

asottile:
I'm going to clarify here since I think I'm a bit confused :)

I'd like to make @bukzor's config work when it changes from:

```yaml
-   repo: git://github.com/bukzor/pre-commit-system-hook.git
    sha: 64a3e3f0d11f74ccd0498ea3149c8e177dd9989c
    hooks:
    -   id: system
[...]

Issue #157, opened 2014-08-30
Don't stash .pre-commit-config.yaml


asottile:
Followup for #156

caffodian:
Dropping in to share my experience with this... 

I spent an hour or so doing a deep dive as the result of the following cryptic output (JSON schema validation failures on the yaml.)  I guess in previous projects I'd always committed the yaml file before actually trying to run hooks.  

```
An error has occurred: InvalidConfigError: Invalid content: .pre-commit-config.yaml
None is not of type u'array'

Failed validating u'type' in schema:
    {u'items': {u'properties': {u'hooks': {u'items': {u'properties': {u'args': {u'items': {u'type': u'string'},
                                                                                u'type': u'array'},
                                                                      u'exclude': {u'default': u'^$',
                                                                                   u'type': u'string'},
                                                                      u'files': {u'type': u'string'},
                                                                      u'id': {u'type': u'string'},
                                                                      u'language_version': {u'type': u'string'}},
                                                      u'required': [u'id'],
                                                      u'type': u'object'},
                                           u'minItems': 1,
                                           u'type': u'array'},
                                u'repo': {u'type': u'string'},
                                u'sha': {u'type': u'string'}},
                u'required': [u'repo', u'sha', u'hooks'],
                u'type': u'object'},
     u'minItems': 1,
     u'type': u'array'}

On instance:
    None
Check the log at ~/.pre-commit/pre-commit.log
```

Eventually figuring out that this was the result of it not being able to actually load the yaml file, because it was being stashed.  

I've since kicked myself a ton, but if we don't want to change it to not stash this yaml file, we could use a friendlier warning/error message and maybe a documentation PR as well.  I'm willing to take a look.  Thoughts?

asottile:
I've been meaning to implement this, just haven't found the time.  I also think that jsonschema needs dropping, it was useful at first when the schema was simple but as you can see the error messages suck :(.  Though the custom validation will be more complicated, it'll probably be easier to add more complicated features and provide better error messages.

caffodian:
Hmm...  I haven't looked at the code that does the git stash (I have a feeling that it's probably more annoying to say "stash everything except this one file" than might be expected) but should be able to pre-detect if the file doesn't exist before trying to run the json validation...which might lead to less confusion.  

And then of course, there's always the least technical but probably quickest thing to do of shoving a giant warning into the docs :p

asottile:
[...]

PR #223, opened 2015-05-05
Preparatory work & clean-up for #219

Lucas-C:
This is an attempt to address #219.
As it is a first draft, it misses some tests and a bit of code cleaning, but I'd like an initial design review before fixing the final glitches.

**Example of valid _ .pre-commit-config.yaml_ using this feature :**
```
- local-hooks:
  - id: no-todo
    name: No TODO allowed in commited files
    entry: grep -iI todo
    expected_return_value: 1
    language: system
    files: ''
```

**Pros / Cons of the solution choosed:**
+ + end-user-friendly: the new "local-hooks" field schema is the same as _hooks.yaml_ schema
+ + nice reuse of MANIFEST_JSON_SCHEMA
+ + no more `_run_hook` / `_run_hooks` code duplication
+ + all the regressions tests pass
- - "local-hooks" and classical repo/sha/hooks objects are mixed together in _.pre-commit-config.yaml_
- - the `hooks_and_invokers` list of tuples is not entirely satisfying. Maybe it'd be worth having a proper `Hook` class ?

### DIFF
--- a/pre_commit/clientlib/validate_config.py
+++ b/pre_commit/clientlib/validate_config.py
@@ -37,7 +37,7 @@ CONFIG_JSON_SCHEMA = {
                 }
             }
         },
-        'required': ['repo', 'sha', 'hooks'],
+        'required': ['repo', 'hooks'],
     }
 }
 
@@ -53,6 +53,10 @@ def try_regex(repo, hook, value, field_name):
 
 def validate_config_extra(config):
     for repo in config:
+        if 'sha' not in repo:
+            raise InvalidConfigError(
+                'Missing "sha" field for repository {0}'.format(repo['repo'])
+            )
         for hook in repo['hooks']:
             try_regex(repo, hook['id'], hook.get('files', ''), 'files')
             try_regex(repo, hook['id'], hook['exclude'], 'exclude')
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -18,6 +18,15 @@ from pre_commit.util import noop_context
 logger = logging.getLogger('pre_commit')
 
 
+class HookExecutor(object):
+    def __init__(self, hook, invoker):
+        self.hook = hook
+        self._invoker = invoker
+
+    def invoke(self, filenames):
+        return self._invoker(self.hook, filenames)
+
+
 def _get_skips(environ):
     skips = environ.get('SKIP', '')
     return set(skip.strip() for skip in skips.split(',') if skip.strip())
@@ -55,21 +64,25 @@ def get_changed_files(new, old):
     )[1].splitlines()
 
 
-def _run_single_hook(runner, repository, hook, args, write, skips=set()):
+def get_filenames(args, include_expr, exclude_expr):
     if args.origin and args.source:
-        get_filenames = git.get_files_matching(
+        getter = git.get_files_matching(
             lambda: get_changed_files(args.origin, args.source),
         )
     elif args.files:
-        get_filenames = git.get_files_matching(lambda: args.files)
+        getter = git.get_files_matching(lambda: args.files)
     elif args.all_files:
-        get_filenames = git.get_all_files_matching
+        getter = git.get_all_files_matching
     elif git.is_in_merge_conflict():
-        get_filenames = git.get_conflicted_files_matching
+        getter = git.get_conflicted_files_matching
     else:
-        get_filenames = git.get_staged_files_matching
+        getter = git.get_staged_files_matching
+    return getter(include_expr, exclude_expr)
 
-    filenames = get_filenames(hook['files'], hook['exclude'])
+
+def _run_single_hook(hook_executor, args, write, skips=frozenset()):
+    hook = hook_executor.hook
+    filenames = get_filenames(args, hook['files'], hook['exclude'])
     if hook['id'] in skips:
         _print_user_skipped(hook, write, args)
         return 0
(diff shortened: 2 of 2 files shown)


asottile:
I like the idea.

I wonder if we can simplify this further and reuse existing logic by making local-hooks slightly less special to configure:

- Make a special value for `repo` (for instance `pre-commit-locally-configured` or maybe something less verbose)
- If that special repo is encountered, its backing manifest is a list which always contains all hook ids with basically a Null configuration
- The code already allows overriding of all the keys presented in the backing configuration so the rest would just work

I also like the cleanup work that was done for `_run_hooks` it might be easier to think about this review if we split this into two PRs as well

:thumbsup:

Lucas-C:
Agreed, I'll split it in to commits.

Your solution sounds good, but it would make the `sha` field no more mandatory. In the JSON schema. But we could add an extra validation step so it's not a blocker.

I'm ok to move to this solution, but don't you think the current JSON schemas definition kind of overlap ? There is some code duplication for the allowed overrides, I'll try to make it more DRY.

asottile:
Yeah the json schemas do kind of overlap.  I really hope to get rid of them eventually (as we've discussed in the other issue) and replace them with a bunch of functions.

imo aim for correctness and readability instead of working hard at DRY, I've found that in a lot of cases expressiveness / correctness kind of work against DRY-ness.  Though I think replacing the jsonchemas will actually make a lot of things cleaner / more flexible but that's definitely out of scope for this change I would think :)

I'm cool with making sha optional as long as we require it in the normal case

Lucas-C:
A comment on something you wrote "The code already allows overriding of all the keys presented in the backing configuration so the rest would just work" : actually the hooks definition in the manifest & config schemas differ quite a bit : some fields actually cannot be overridden, only `id` is required in the config, and it also has an additional `exclude` field, etc. Due to this differences (manly the missing fields) I feel like we cannot actually implement "local hooks" without modifying the schema.
What would you thing of something like this in `validate_config.py` :
```
from pre_commit.clientlib.validate_manifest import MANIFEST_JSON_SCHEMA
CONFIG_JSON_SCHEMA = {
    'type': 'array',
    'minItems': 1,
    'items': {
        'type': 'object',
        'properties': {
            'repo': {'type': 'string'},
            'sha': {'type': 'string'},
            'hooks': MANIFEST_JSON_SCHEMA
        },
        'required': ['repo', 'hooks'],
    }
}
```

With possibly more tweaking (like adding the `exclude` field to the manifest).

Lucas-C:
I updated the pull request to only contains preliminary changes.
Please don't merge it until you answer my question above :)

Also you wrote "I really hope to get rid of [JSON schemas] eventually, as we've discussed in the other issue". I don't think you mentioned that before in an issue I was part of. Are you refering to issue #157 or something else ?

asottile:
I'll comment on the second thing when I'm not on the phone :D

The first however: though the fields are not specified in the config schema,  the way the runtime hook dictionary is made is by merging the config dict on top of the manifest dict.  JSON schema is a bit permissive in that it let's you add keys even if they aren't in the schema.

pre_commit/runner.py
```diff
@@ -35,10 +35,14 @@ def config_file_path(self):
...
+    def config(self):
+        return load_config(self.config_file_path)
+
+    @cached_property
     def repositories(self):
         """Returns a tuple of the configured repositories."""
-        config = load_config(self.config_file_path)
-        repositories = tuple(Repository.create(x, self.store) for x in config)
+        repositories = tuple(Repository.create(hook, self.store)
+                             for hook in self.config)
```
asottile: Are the changes to this file necessary? I don't see them used elsewhere

pre_commit/clientlib/validate_config.py
```diff
@@ -53,6 +53,9 @@ def try_regex(repo, hook, value, field_name):
 
 def validate_config_extra(config):
     for repo in config:
+        if 'sha' not in repo:
+            raise InvalidConfigError('Missing "sha" field'
+                                     'for repository {0}'.format(repo['repo']))
```
asottile: minor nitpick (sorry!):

The predominant style in this repository is to do function calls:

```
like_this(
    with_args,
    here,
)
```

Mostly because it slightly reduces merge conflicts when adding arguments etc.

I can just fix this up after merging if you don't want to bother though :)

pre_commit/commands/run.py
```diff
@@ -18,6 +18,15 @@
 logger = logging.getLogger('pre_commit')
 
 
+class HookExecuter(object):
```
asottile: executer or executor?

firefox's spellcheck at least prefers executor

asottile:
Address comment for pre_commit/runner.py and I'll merge

:+1:

asottile:
Ah yeah as for jsonschema, I totally got you confused with another person.  I'm terrible sorry about that >.<  Here's the comment I mentioned it in: https://github.com/pre-commit/pre-commit/issues/157#issuecomment-98570101

pre_commit/runner.py
Lucas-C: Ok, I'll remove them from this commit

pre_commit/clientlib/validate_config.py
Lucas-C: It's ok, done

pre_commit/commands/run.py
Lucas-C: Executor sounds better :)
Done

Lucas-C:
> the way the runtime hook dictionary is made is by merging the config dict on top of the manifest dict.
Are you refering to the following code snippet ?
https://github.com/pre-commit/pre-commit/blob/master/pre_commit/repository.py#L48

Then ok, but we won't be able to reuse that code for the local hooks, as it's nested in the `Repository` class.
But we can simply copy that merge logic.

asottile:
Oh forgot to mention, please PR to add your name on http://github.com/pre-commit/pre-commit.github.io :D

Thanks for contributing!

PR #225, opened 2015-05-08
--allow-unstaged-config arg

caffodian:
Change the behaviour for unstaged config per #157 discussion

- if allow unstaged config, puts a warning message that the config will still be stashed, but continues
- if not provided, errors out

stuff I'd expect we might want to change would be the actual content of the messages and perhaps what the "default" is here.  

thanks for the support! @asottile

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -129,6 +129,18 @@ def _has_unmerged_paths(runner):
     return bool(stdout.strip())
 
 
+def _has_unstaged_config(runner):
+    retcode, stdout, stderr = runner.cmd_runner.run(
+        [
+            'git', 'diff', '--exit-code', runner.config_file_path
+        ],
+        retcode=None,
+        encoding=None,
+    )
+    # be explicit, other git errors don't mean it has an unstaged config.
+    return retcode == 1
+
+
 def run(runner, args, write=sys_stdout_write_wrapper, environ=os.environ):
     # Set up our logging handler
     logger.addHandler(LoggingHandler(args.color, write=write))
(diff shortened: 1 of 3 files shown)


caffodian:
Ha.  Ironically, I somehow missed actually running the hooks on one of the files while messing about :p   should be resolved now.

pre_commit/commands/run.py
```diff
@@ -151,6 +163,16 @@ def run(runner, args, write=sys_stdout_write_wrapper, environ=os.environ):
...
+    if _has_unstaged_config(runner):
+        if args.allow_unstaged_config:
+            logger.warn('You have an unstaged config file and have '
+                        'specified the --allow-unstaged-config option.\n'
+                        'Note that your config will be stashed before the '
+                        'config is parsed unless --no-stash is specified.')
+        else:
+            logger.error('You have an unstaged config file and have not '
+                         'specified the --allow-unstaged-config option.\n')
+            return 1
```
asottile: Let's make `--no-stash` silence both of these errors

pre_commit/main.py
```diff
@@ -94,6 +94,11 @@ def main(argv=None):
         '--source', '-s',
         help='The remote branch"s commit_id when using `git push`',
     )
+    run_parser.add_argument(
+        '--allow-unstaged-config', default=False, action='store_true',
+        help='Allow an unstaged config to be present.  Note that this will \
```
asottile: nit: prefer paren'd continuation to `\`s

I'll fix it when I merge if you don't want to bother :)

tests/commands/run_test.py
```diff
@@ -334,3 +336,39 @@ def test_lots_of_files(mock_out_store_directory, tmpdir_factory):
...
+
+    args = _get_opts(allow_unstaged_config=allow_unstaged_config)
+    ret, printed = _do_run(repo_with_passing_hook, args)
+    common_msg = 'You have an unstaged config file'
+    warning_msg = 'have specified the --allow-unstaged-config option.'
+
+    error_msg = 'have not specified the --allow-unstaged-config option.\n'
+
+    assert common_msg in printed
+    if allow_unstaged_config:
```
asottile: I've found when there's two disparate branches in a test that it's often clearer as two separate tests

Here's a nice article I like to reference about the topic: http://googletesting.blogspot.com/2014/07/testing-on-toilet-dont-put-logic-in.html

In this case it might be a good idea to split the repeated bits into a helper and call that

asottile:
Overall looks great!  Fix up those few little things, squash the commits and I'll merge it \o/

caffodian:
fixed and squashed.  not sure if appveyor is just going nuts because of the squash or...

asottile:
Appveyor seems to be mad because it merge-conflicts with master now

asottile:
:+1: Thanks for the contribution! Please PR http://github.com/pre-commit/pre-commit.github.io to add yourself :D

asottile:

PR #226, opened 2015-05-10
Adding support for locally-defined hooks

### LINKED ISSUE #219: Add support for pcre / scripts / system hooks definition in .pre-commit-config.yaml

Lucas-C:
Closes #219
All tests pass and coverage for new file `pre_commit/hook_executor.py` is 100%

### DIFF
--- a/pre_commit/clientlib/validate_config.py
+++ b/pre_commit/clientlib/validate_config.py
@@ -6,6 +6,13 @@ from pre_commit.clientlib.validate_base import is_regex_valid
 from pre_commit.errors import FatalError
 
 
+_LOCAL_HOOKS_MAGIC_REPO_STRING = 'local'
+
+
+def is_local_hooks(repo_entry):
+    return repo_entry['repo'] == _LOCAL_HOOKS_MAGIC_REPO_STRING
+
+
 class InvalidConfigError(FatalError):
     pass
 
@@ -53,7 +60,12 @@ def try_regex(repo, hook, value, field_name):
 
 def validate_config_extra(config):
     for repo in config:
-        if 'sha' not in repo:
+        if is_local_hooks(repo):
+            if 'sha' in repo:
+                raise InvalidConfigError(
+                    '"sha" property provided for local hooks'
+                )
+        elif 'sha' not in repo:
             raise InvalidConfigError(
                 'Missing "sha" field for repository {0}'.format(repo['repo'])
             )
--- a/pre_commit/repository.py
+++ b/pre_commit/repository.py
@@ -5,6 +5,10 @@ import shutil
 
 from cached_property import cached_property
 
+from pre_commit import git
+from pre_commit.clientlib.validate_config import is_local_hooks
+from pre_commit.clientlib.validate_manifest import MANIFEST_JSON_SCHEMA
+from pre_commit.jsonschema_extensions import apply_defaults
 from pre_commit.languages.all import languages
 from pre_commit.manifest import Manifest
 from pre_commit.prefixed_command_runner import PrefixedCommandRunner
@@ -21,10 +25,13 @@ class Repository(object):
 
     @classmethod
     def create(cls, config, store):
-        repo_path_getter = store.get_repo_path_getter(
-            config['repo'], config['sha']
-        )
-        return cls(config, repo_path_getter)
+        if is_local_hooks(config):
+            return LocalRepository(config)
+        else:
+            repo_path_getter = store.get_repo_path_getter(
+                config['repo'], config['sha']
+            )
+            return cls(config, repo_path_getter)
 
     @cached_property
     def repo_url(self):
@@ -111,3 +118,28 @@ class Repository(object):
         return languages[hook['language']].run_hook(
             self.cmd_runner, hook, file_args,
         )
+
+
+class LocalRepository(Repository):
+    def __init__(self, repo_config, repo_path_getter=None):
+        repo_path_getter = None
+        super(LocalRepository, self).__init__(repo_config, repo_path_getter)
+
+    @cached_property
+    def hooks(self):
+        return tuple(
+            (hook['id'], apply_defaults(hook, MANIFEST_JSON_SCHEMA['items']))
+            for hook in self.repo_config['hooks']
+        )
+
+    @cached_property
+    def cmd_runner(self):
+        return PrefixedCommandRunner(git.get_root())
+
+    @cached_property
+    def sha(self):
+        raise NotImplementedError
+
+    @cached_property
+    def manifest(self):
+        raise NotImplementedError
--- a/tests/clientlib/validate_config_test.py
+++ b/tests/clientlib/validate_config_test.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import jsonschema
 import pytest
 
 from pre_commit.clientlib.validate_config import CONFIG_JSON_SCHEMA
@@ -25,7 +26,7 @@ def test_run(input, expected_output):
     assert run(input) == expected_output
 
 
-@pytest.mark.parametrize(('manifest_obj', 'expected'), (
+@pytest.mark.parametrize(('config_obj', 'expected'), (
     ([], False),
     (
         [{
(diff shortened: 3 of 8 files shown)


tests/commands/run_test.py
```diff
@@ -14,11 +14,12 @@
 from pre_commit.commands.run import _has_unmerged_paths
 from pre_commit.commands.run import get_changed_files
 from pre_commit.commands.run import run
+from pre_commit.ordereddict import OrderedDict
 from pre_commit.runner import Runner
 from pre_commit.util import cmd_output
 from pre_commit.util import cwd
 from testing.auto_namedtuple import auto_namedtuple
-from testing.fixtures import make_consuming_repo
+from testing.fixtures import add_config_to_repo, make_consuming_repo
```
asottile: weird, there's a pre-commit hook that should split this line, does this line change when `pre-commit run --all-files` ? (might be a side-effect of rebase?)

tests/commands/run_test.py
```diff
@@ -81,7 +82,7 @@ def _test_run(repo, options, expected_outputs, expected_ret, stage):
         stage_a_file()
     args = _get_opts(**options)
     ret, printed = _do_run(repo, args)
-    assert ret == expected_ret
+    assert ret == expected_ret, printed
```
asottile: I'd imagine:

`assert ret == expected_ret, (ret, expected_ret, printed)`

would be more useful

tests/commands/run_test.py
```diff
@@ -334,3 +333,68 @@ def test_lots_of_files(mock_out_store_directory, tmpdir_factory):
...
+            ('entry', 'grep -iI todo'),
+            ('expected_return_value', 1),
+            ('language', 'system'),
+            ('files', ''),
+        ))])
+    ))
+    add_config_to_repo(repo_with_passing_hook, config)
+    assert Runner(repo_with_passing_hook).local_hooks[0]['id'] == 'no-todo'
+
+    with io.open('dummy', 'w') as config_file:
```
asottile: `config_file`?

tests/commands/run_test.py
Lucas-C: Agreed, fixing that

pre_commit/hook_executor.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import unicode_literals
```
asottile: so, my main issue with this is the `Repository` is supposed to be the hook runner, now we'd have two things that have that responsibility...

I think we should maybe instead push the responsibility into `Repository.create`?

tests/clientlib/validate_config_test.py
```diff
@@ -66,8 +67,8 @@ def test_run(input, expected_output):
         False,
     ),
 ))
-def test_is_valid_according_to_schema(manifest_obj, expected):
-    ret = is_valid_according_to_schema(manifest_obj, CONFIG_JSON_SCHEMA)
+def test_is_valid_according_to_schema(config_obj, expected):
+    ret = is_valid_according_to_schema(config_obj, CONFIG_JSON_SCHEMA)
```
asottile: oops my bad! ++

tests/commands/run_test.py
Lucas-C: My bad, I hadn't install pre-commit properly on this repo ! Fixed

tests/commands/run_test.py
Lucas-C: Fixed

pre_commit/hook_executor.py
Lucas-C: Yes, that was also my main concern. I don't really like adding those classes.
Glad you suggest a different approach.
Are you ok if I drop all the HookExecuter classes, but create a LocalRepository subclass of Repository ?

asottile: Sounds good to me :)

pre_commit/repository.py
```diff
@@ -111,3 +117,28 @@ def run_hook(self, hook, file_args):
...
+    def cmd_runner(self):
+        return PrefixedCommandRunner(git.get_root())
+
+    @cached_property
+    def sha(self):
+        return None
+
+    @cached_property
+    def manifest(self):
+        return None
```
asottile: I'd just change these to `raise NotImplementedError`

pre_commit/clientlib/validate_config.py
```diff
@@ -52,8 +59,19 @@ def try_regex(repo, hook, value, field_name):
 
 
 def validate_config_extra(config):
+    local_repos_found = False
     for repo in config:
-        if 'sha' not in repo:
+        if is_local_hooks(repo):
+            if local_repos_found:
+                raise InvalidConfigError(
+                    'Multiple local hooks (where repo = null) found'
```
asottile: oops, this is off now, are we going with `null` or `'local'`?

pre_commit/repository.py
```diff
@@ -21,6 +25,8 @@ def __init__(self, repo_config, repo_path_getter):
 
     @classmethod
     def create(cls, config, store):
+        if is_local_hooks(config):
+            return LocalRepository(config)
```
asottile: let's make this `if...: ... else: ...` so it's slightly more explicit

tests/clientlib/validate_config_test.py
```diff
@@ -121,3 +122,74 @@ def test_config_with_ok_exclude_regex_passes():
...
+            'name': 'Block if "DO NOT COMMIT" is found',
+            'entry': 'DO NOT COMMIT',
+            'language': 'pcre',
+            'files': '^(.*)$',
+        }],
+    }],
+))
+def test_config_with_local_hooks_definition_fails(config_obj):
+    # Catching jsonschema.exceptions.ValidationError or InvalidConfigError
+    with pytest.raises(Exception):
```
asottile: I think you can explicitly include both of those exceptions:

```
with pytest.raises((jsonschema.exceptions.ValidationError, InvalidConfigError)):
    # ...
```

asottile:
:+1: Looks great, address my last couple comments and I'll merge it

Lucas-C:
Thanks for all your comments !
Everything should be fixed now :)

Lucas-C:
I guess one last to do will be to update the docs :smile:

pre_commit/clientlib/validate_config.py
```diff
@@ -52,8 +59,19 @@ def try_regex(repo, hook, value, field_name):
...
 
 def validate_config_extra(config):
+    local_repos_found = False
     for repo in config:
-        if 'sha' not in repo:
+        if is_local_hooks(repo):
+            if local_repos_found:
+                raise InvalidConfigError(
+                    'Multiple local hooks found'
+                )
```
asottile: one last thing (sorry, didn't notice this on the first passes), is there a reason to raise this error? it should just work now that we're doing it at the repository right? (it's a bit silly I agree, but it should work just fine if the check is removed?)

Lucas-C: Yes, but I prefer to be strict at this early stage of config parsing, instead of allowing weird anomalies later.
E.g. if one needs a list of all the local hooks and assume there can be only one entry for them in the config.
Better avoid ambiguity, don't you think ?

asottile: I feel that if it's not broken, why go through the extra trouble to exclude the case?

Also, since hooks run in the order they are defined, it might be desirable to run some hooks out-of-order.  We currently don't forbid repeated repos at the moment so I see no reason to make `local` repos special in that regard.

Lucas-C: Ok, this last argument convinced me. I'm going to remove the check.

asottile:

asottile:
local hooks are now documented here: http://pre-commit.com/#repository-local-hooks

PR #243, opened 2015-06-14
Fixing bug with local hooks that disappeared during autoupdate

Lucas-C:
I added a test, `test_autoupdate_local_hooks_with_out_of_date_repo`, that is currently failing without this patch.

The issue typically araise when mixing local & remote hooks : on a successful remote hooks autoupdate, local hooks are simply ignored when the new YAML config is writen to file.


asottile:
Thanks for the patch!  I'll bump the version hopefully later today

asottile:
This is available in 0.5.3

PR #247, opened 2015-07-20
fix for issue 246

dkunitsk:


tests/commands/install_uninstall_test.py
```diff
@@ -87,6 +87,13 @@ def test_install_hooks_directory_not_present(tmpdir_factory):
     install(runner)
     assert os.path.exists(runner.pre_commit_path)
 
+def test_install_hooks_dead_symlink(tmpdir_factory):
```
asottile: this'll probably require a skip for windows due to symlinks? we'll find out :)

asottile:
Thanks for the patch!  Please PR https://github.com/pre-commit/pre-commit.github.io to add yourself to the contributors :D

PR #249, opened 2015-07-26
Make `pre-push-tmpl` non-executable

chriskuehl:
Howdy strangers of pre-commit,

This makes `pre_commit/resources/pre-push-tmpl` non-executable.

This is pretty trivial and probably seems unnecessary. The reasoning behind it is that Debian complains about executable files which are neither binaries nor scripts with a shebang:

```
Now running lintian...
W: pre-commit: executable-not-elf-or-script usr/lib/python3/dist-packages/pre_commit/resources/pre-push-tmpl
Finished running lintian.
```

In this case it looks like it makes more sense to make `pre-push-tmpl` non-executable (rather than adding a shebang) since it ends up inserted in the middle of `hook-tmpl` (which has a shebang).

I don't think there are any cases where we need this to be executable, but maybe I'm wrong?


asottile:
Correct, in fact the other template is non executable as well!

PR #250, opened 2015-07-26
Add --version option to validator binaries

chriskuehl:
It would be nice if these binaries had a `--version` option as well as the `--help` option they already have.

They could use the same copy-pasted-from-stackoverflow snippet we use for the main pre-commit.

One benefit is that you can use help2man to generate man pages automatically for these validators.

(Normally these are only used as pre-commit hooks, but they can be used standalone as well. Debian wants manpages for any binaries, and we can generate them easily given `--help` and `--version`. The alternative is to not ship them with the Debian package, or to apply this patch only in the Debian package and not upstream.)

### DIFF
--- a/pre_commit/clientlib/validate_base.py
+++ b/pre_commit/clientlib/validate_base.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import argparse
 import os.path
+import pkg_resources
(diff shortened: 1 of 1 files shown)


asottile:
Thanks!

PR #252, opened 2015-07-26
Run this repo's pre-commit hooks in tox (and thus travis)

chriskuehl:
seems like a good idea. not sure if there's some edge case where this doesn't make sense (or where it needs to be a separate target)?

### DIFF
--- a/pre_commit/clientlib/validate_base.py
+++ b/pre_commit/clientlib/validate_base.py
@@ -3,12 +3,12 @@ from __future__ import unicode_literals
 
 import argparse
 import os.path
-import pkg_resources
 import re
 import sys
 
 import jsonschema
 import jsonschema.exceptions
+import pkg_resources
 import yaml
 
 from pre_commit.jsonschema_extensions import apply_defaults
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,3 +20,4 @@
     sha: 3d86483455ab5bd06cc1069fdd5ac57be5463f10
     hooks:
     -   id: reorder-python-imports
+        language_version: python2.7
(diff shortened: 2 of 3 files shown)


chriskuehl:
retest this please

chriskuehl:
Fails because argparse isn't part of Python 2.6:
```
diff --git a/pre_commit/clientlib/validate_base.py b/pre_commit/clientlib/validate_base.py
index 707bdde..af18725 100644
--- a/pre_commit/clientlib/validate_base.py
+++ b/pre_commit/clientlib/validate_base.py
@@ -1,11 +1,11 @@
 
-import argparse
 import os.path
 import re
 import sys
 
+import argparse
 import jsonschema
 import jsonschema.exceptions
 import yaml
diff --git a/pre_commit/main.py b/pre_commit/main.py
index aed6688..1f52c33 100644
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -1,9 +1,9 @@
 
-import argparse
 import os
 import sys
 
+import argparse
 import pkg_resources
 
 from pre_commit import color
diff --git a/tests/main_test.py b/tests/main_test.py
index 140b587..ac1674f 100644
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import
 
 import argparse
-
 import mock
 import pytest
```

Probably need to just run this on Python 2.7, I'll follow up to this later. Spent some time perusing @asottile's aspy.refactor_imports, but looks like it would be somewhat complicated to add a way to fake classify 3rd-party modules as part of the standard library for old Python versions.

asottile:
@chriskuehl if you put `language_version: python2.7` on reorder imports I think this is merge able

pre_commit/clientlib/validate_base.py
```diff
@@ -3,12 +3,12 @@
 
 import argparse
 import os.path
-import pkg_resources
```
chriskuehl: heh, this was definitely my fault. at least i recognized that "someone" would make a dumb mistake like this!

chriskuehl:
fyi the commit you merged was fine even though it failed appveyor. I think I angered it by pushing too frequently.

asottile:
Also apparently "you cannot merge a request with failing checks from a phone". Switch to desktop version works though.  Dumb feature github :/

Issue #220, opened 2015-05-01
Feature request: a special "<text>" value for the hooks.yaml "files" field


Lucas-C:
Hi,

I had to implement this kind of "binary-files-filtering-out" today:
https://github.com/Lucas-C/pre-commit-hooks/blob/master/pre_commit_hooks/utils.py#L10

So I made a quick overview of the existing checkers, and I found only 2 that are targeting **any** text file: [end-of-file-fixer](https://github.com/pre-commit/pre-commit-hooks/blob/master/hooks.yaml#L76) and [trailing-whitespaces](https://github.com/pre-commit/pre-commit-hooks/blob/master/hooks.yaml#L106). Both use a long list of known text files suffixes.

If ever you think this use case could come in handy, I'd suggest to provide a magic value like `<any text file>` to the `files` parameter in _hooks.yaml_, so that hooks developers don't have to re-implement this filtering each time.

You can freely re-use the code I mentioned above if it comes handy to you.

Also, sorry for the 3-issues-opened-in-a-row :) I got really excited by this tool and ideas for improvements just flowed in ! But they are really just humble features suggestions.

Cheers

asottile:
+1 this would be great, we already do some file filtering so I imagine we could just add another filtering function to do this. If we didn't support windows I'd probably use something like libmagic, but shouldn't be too hard to find printables / unprintables

Lucas-C:
The solution I used is very portable, take a look :)

We can do even more simple by using the heuristic used by `git`: http://stackoverflow.com/a/7112964/636849
(i.e. just checking for \0 characters)

dupuy:
@asottile wrote:
> If we didn't support windows I'd probably use something like libmagic

In the shell-scripted pre-commit hooks I used before switching to this package, I used `file` magic to select files - this was especially useful for running checks against scripts where there was no file extension, but a simple hashbang check could tell that a particular file was Python, or Perl, or *sh...
I would love it if there were some simple anchored PCRE filter that could select files based on content (of, usually, the first line), so that I could do something like this in `hooks.yaml`:
```
-   id: autopep8-wrapper
    name: autopep8 wrapper
    description: "Runs autopep8 over python source..."
    entry: autopep8-wrapper
    language: python
    files: \.py$
    file-content: #! *?/.*?\bpython\d?\b
    args: [-i]
```
[...]

Issue #191, opened 2015-02-04
pre-commit should probably ignore symlinks?


asottile:
Right now some hooks do the wrong thing with symlinks.  Maybe the hooks should be written such that they ignore symlinks?

kentwills:
+1

asottile:
Also probably should ignore submodules too

asottile:
pre-commit 0.15.0 will not lint symlinks or submodules by default.

To explicitly lint symlinks use:
```yaml
    types: [symlink]
```
To explicitly lint submodules use:
```yaml
    types: [directory]
```

PR #254, opened 2015-07-29
[WIP] Don't run hooks on submodule roots

chriskuehl:
git tracks submodules like files. In particular, the roots of submodules appear in the output of `git ls-files`. This means we get weird errors like:

```
ckuehl@anthrax:~/tmp$ git init derp && cd derp
ckuehl@anthrax:~/tmp/derp$ git submodule add git@github.com:chriskuehl/dotfiles herp
ckuehl@anthrax:~/tmp/derp$ cat > .pre-commit-config.yaml
-   repo: https://github.com/pre-commit/pre-commit-hooks.git
    sha: 5dd2605fbe26937c721a8ecc0e53949d3eea4f12
    hooks:
    -   id: detect-private-key
ckuehl@anthrax:~/tmp/derp$ pre-commit run --all-files
Detect Private Key.............................................................Failed
  File "/home/c/ck/ckuehl/.pre-commit/repokqqh2jfj/py_env-default/bin/detect-private-key", line 9, in <module>
    load_entry_point('pre-commit-hooks==0.4.2', 'console_scripts', 'detect-private-key')()
  File "/home/c/ck/ckuehl/.pre-commit/repokqqh2jfj/py_env-default/local/lib/python2.7/site-packages/pre_commit_hooks/detect_private_key.py", line 15, in detect_private_key
    content = open(filename, 'rb').read()
IOError: [Errno 21] Is a directory: 'herp'
```

The only way to avoid this right now is to explicitly exclude submodule paths in you `.pre-commit-config.yaml` (or to modify every hook to check if an input file is actually a file, which seems bad).

We can exclude these with a slightly more complicated `ls-files` command.

I didn't test this extensively or even check if the tests still pass, mostly just looking to see if this looks like a reasonable solution. If it does I'll go ahead and write real tests (and test it on some older git versions to make sure it still works).

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -10,6 +10,11 @@ from pre_commit.errors import FatalError
 from pre_commit.util import cmd_output
 from pre_commit.util import memoize_by_cwd
 
+# octal constants for git file modes
+GIT_MODE_FILE = 0o100644
+GIT_MODE_EXECUTABLE = 0o100755
+GIT_MODE_SYMLINK = 0o120000
+GIT_MODE_SUBMODULE = 0o160000
 
 logger = logging.getLogger('pre_commit')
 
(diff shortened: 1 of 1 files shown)


pre_commit/git.py
```diff
@@ -69,7 +74,28 @@ def get_staged_files():
...
+
+    def split(line):
+        match = split_regex.match(line)
+        return int(match.group(1), 8), match.group(2)
+
+    output = cmd_output('git', 'ls-files', '--stage')[1]
+    return [
+        path for mode, path in [
+            split(line) for line in output.splitlines()
+        ] if mode != GIT_MODE_SUBMODULE
```
chriskuehl: Could also exclude symlinks here if desired. I didn't because I can imagine some useful hook (e.g. checking for symlinks pointing outside of the repo), but it could also potentially be dangerous (e.g. we could end up running a hook on a file outside of the git repo, transparently to the hook, which then "fixes" a file outside of the git repo).

chriskuehl: There are also _maybe_ useful hooks to run against submodules, but imo those should just read `.gitmodules` instead. Random hooks shouldn't have to worry that they'll be passed in a directory.

asottile:
This seems similar to the approach that I wanted to take.  One thing that I wanted though is a way for hooks to opt into checking submodules / symlinks in hooks.yaml

chriskuehl:
What would you think about adding a per-hook option (defined in hooks.yaml or overriden in .pre-commit-config.yaml) like:

    types: ['files', 'symlinks', 'submodules']

where the default (if omitted) is just `['files']`? Could also separate executable files but not sure that's really helpful.

asottile:
Yeah that sounds good, also makes it easy to implement #220

chriskuehl:
Abandoning this PR, I'll make another that takes care of #191, #220, and #257.

PR #258, opened 2015-08-07
Allow specifying python version relative to user's home directory.

meunierd:
The main use case for this PR is to more easily work with user-local python installations, particularly pyenv's. This way a user can specify `~/.pyenv/versions/3.4.3/bin/python`. This is particularly useful when working with projects across many python versions, where it doesn't make sense to have them all available on your PATH, while still keeping the .pre-commit-config.yaml portable by not specifying an absolute path.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -31,8 +31,11 @@ def in_env(repo_cmd_runner, language_version):
 
 
 def norm_version(version):
+    version = os.path.expanduser(version)
     if os.name == 'nt':  # pragma: no cover (windows)
         if not distutils.spawn.find_executable(version):
+            # expanduser introduces a leading slash
+            version = version.strip('\\')
             # The default place for python on windows is:
             # C:\PythonXX\python.exe
             version = r'C:\{0}\python.exe'.format(version.replace('.', ''))
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import io
+import os
 import os.path
 import shutil
 
@@ -12,6 +13,7 @@ from pre_commit import five
 from pre_commit.clientlib.validate_config import CONFIG_JSON_SCHEMA
 from pre_commit.clientlib.validate_config import validate_config_extra
 from pre_commit.jsonschema_extensions import apply_defaults
+from pre_commit.languages.python import norm_version
 from pre_commit.languages.python import PythonEnv
 from pre_commit.repository import Repository
 from pre_commit.util import cmd_output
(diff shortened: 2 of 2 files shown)


asottile:
Looks good, mind writing a test so this doesn't regress?

meunierd:
Yeah! I just wasn't sure where it should live in the test suite. Any suggestions?

asottile:
Probably this file would suffice: https://github.com/pre-commit/pre-commit/blob/master/tests/repository_test.py

Something similar to what test_switch_language_versions_doesnt_clobber does but set up a virtualenv in the home directory

meunierd:
Updated with a test. The intention wasn't to specify the target of the virtualenv itself but rather the python binary passed to with `-p`

tests/repository_test.py
```diff
@@ -414,3 +416,12 @@ def test_local_repository():
...
     assert len(local_repo.hooks) == 1
+
+
+def test_norm_version_expanduser():
+    old_home = os.environ.get('HOME', None)
+    test_home = '/Users/someuser'
+    os.environ['HOME'] = test_home
+    expected_path = '/Users/someuser/.pyenv/versions/3.4.3/bin/python'
+    assert norm_version('~/.pyenv/versions/3.4.3/bin/python') == expected_path
+    os.environ['HOME'] = old_home
```
asottile: you'll want to use:

```python

with mock.patch.dict(os.environ, {'HOME': test_home}):
    # run your assertion
```

right now when the assertion fails, the environment is polluted for the rest of the test.

I think you might also need to twiddle the file paths so the test passes on windows.

meunierd:
Finally updated this PR. Sorry about the delay/noise. Windows is a pain.

pre_commit/languages/python.py
```diff
@@ -31,11 +31,13 @@ def in_env(repo_cmd_runner, language_version):
 
 
 def norm_version(version):
+    version = os.path.expanduser(version)
     if os.name == 'nt':  # pragma: no cover (windows)
         if not distutils.spawn.find_executable(version):
             # The default place for python on windows is:
             # C:\PythonXX\python.exe
             version = r'C:\{0}\python.exe'.format(version.replace('.', ''))
+            version = version.replace(r'C:\\', 'C:\\')
```
asottile: hmm, why was this necessary?

Maybe move the expanduser after the windows branch?

tests/repository_test.py
```diff
@@ -414,3 +416,15 @@ def test_local_repository():
     with pytest.raises(NotImplementedError):
         local_repo.manifest
     assert len(local_repo.hooks) == 1
+
+
+def test_norm_version_expanduser():
+    home = os.path.expanduser('~')
+    if os.name == 'nt':
```
asottile: `# pragma: no cover` these two branches so coverage doesn't complain :)

asottile:
Sweet, looks good -- minor nitpicks left :D

pre_commit/languages/python.py
meunierd: The expansion of `~` on Windows has a leading slash for some reason.

meunierd:
Nit picked. I'm not sure of a better approach for the C:\\ replace though.

pre_commit/languages/python.py
asottile: Ah, maybe remove the C:\ part entirely from the format string?  iirc \foo\bar is the same as C:\foo\bar anyway and then pre-commit cares less about what your default drive is

meunierd:
Updated with a more explicit handling of the leading slash issue.

asottile:
Sweet! Thanks for the PR, be sure to add yourself on https://github.com/pre-commit/pre-commit.github.io

meunierd:
Thanks! Will do.

PR #259, opened 2015-08-12
Fix pre-commit when python-future is installed and a py3 virtualenv is being created

asottile:
Using http://github.com/asottile/virtualenv-hax in place of virtualenv to work around https://github.com/PythonCharmers/python-future/issues/148


chriskuehl:
please no

asottile:
Seems no longer necessary with latest virtualenv

asottile:
now to get yelp to upgrade...

Issue #255, opened 2015-07-29
Feature request: pre-commit or pre-push only hooks


ymilki:
I am writing a pre-commit hook that validates the author and committer emails of commits. However, these values can only be guaranteed to queried properly *after* the commit is created.

git does not provide utilities to retrieve this information in the pre-commit stage nor can you retrieve explicit information from the --author flag. (http://stackoverflow.com/questions/9627198/stop-a-git-commit-by-a-specific-author-using-pre-commit-hook)

I would like to ensure that this hook is only run at pre-push time. I can think of two implementations:

* yaml config file can specify pre-push or pre-commit only
* pre-push/pre-commit stage is made available via environmental variable when hook is run.

ymilki:
Turns out, this is partially blocked by #256 since commits can be made without changed files.

dnephin:
I have a use for this as well

barrysteyn:
I would also need something like this

asottile:
I think the way I'd like to see this implemented is a per-hook `stage` option (that defaults to a list of all the stages, currently `['commit', 'push']`).

Probably something like:

```yaml
-   repo: ...
    sha: ...
    hooks:
    -   id: check-added-large-files
        stages: [commit]
    -   id: check-merge-conflict
        stages: [push]
```

Does this seem reasonable?

barrysteyn:
Hi

I like this. But one thing... Can stages mimic the names of the hooks, e.g:

```
-   repo: ...
    sha: ...
    hooks:
    -   id: check-added-large-files
        stages: [commit]
    -   id: check-merge-conflict
        stages: [pre-commit,pre-push]
```

barrysteyn:
@asottile  Actually, after thinking about it some more, ignore my last comment. I like yours much better. stages will be commit or push.

barrysteyn:
This issue is now solved in #271

PR #271, opened 2015-10-02
Allow hooks to be applied to either pre-push or pre-commit

barrysteyn:
Solve #255

Allows hooks to be filtered to pre-commit or pre-push. The hooks section in .`pre-commit-config.yaml` can now have a `filter` key, which if defined, can be `pre-push` or `pre-commit`.

I have not written formal tests for this. If whoever is has control over this repo likes the effort, I will very happily add some unit tests and update documentation. Just to note that not only do I need this feature badly, but others (like #255) also need it.

### DIFF
--- a/pre_commit/clientlib/validate_manifest.py
+++ b/pre_commit/clientlib/validate_manifest.py
@@ -25,6 +25,13 @@ MANIFEST_JSON_SCHEMA = {
             'language_version': {'type': 'string', 'default': 'default'},
             'files': {'type': 'string'},
             'expected_return_value': {'type': 'number', 'default': 0},
+            'stages': {
+                'type': 'array',
+                'default': [],
+                'items': {
+                    'type': 'string',
+                },
+            },
             'args': {
                 'type': 'array',
                 'default': [],
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -175,6 +175,7 @@ def run(runner, args, write=sys_stdout_write_wrapper, environ=os.environ):
 
     with ctx:
         repo_hooks = list(get_repo_hooks(runner))
+
         if args.hook:
             repo_hooks = [
                 (repo, hook) for repo, hook in repo_hooks
@@ -183,4 +184,11 @@ def run(runner, args, write=sys_stdout_write_wrapper, environ=os.environ):
             if not repo_hooks:
                 write('No hook with id `{0}`\n'.format(args.hook))
                 return 1
+
+        # Filter hooks for stages
+        repo_hooks = [
+            (repo, hook) for repo, hook in repo_hooks
+            if not hook['stages'] or args.hook_stage in hook['stages']
+        ]
+
         return _run_hooks(repo_hooks, args, write, environ)
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -87,7 +87,6 @@ def main(argv=None):
     run_parser.add_argument(
         '--verbose', '-v', action='store_true', default=False,
     )
-
     run_parser.add_argument(
         '--origin', '-o',
         help='The origin branch\'s commit_id when using `git push`',
@@ -100,6 +99,10 @@ def main(argv=None):
         '--allow-unstaged-config', default=False, action='store_true',
         help='Allow an unstaged config to be present.  Note that this will'
         'be stashed before parsing unless --no-stash is specified'
+    )
+    run_parser.add_argument(
+        '--hook-stage', choices=('commit', 'push'), default='commit',
+        help='The stage during which the hook is fired e.g. commit or push'
     )
     run_mutex_group = run_parser.add_mutually_exclusive_group(required=False)
     run_mutex_group.add_argument(
--- a/pre_commit/resources/hook-tmpl
+++ b/pre_commit/resources/hook-tmpl
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # This is a randomish md5 to identify this script
-# 79f09a650522a87b0da915d0d983b2de
+# e358c9dae00eac5d06b38dfdb1e33a8c
 
 pushd `dirname $0` > /dev/null
 HERE=`pwd`
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -59,6 +59,7 @@ def _get_opts(
         origin='',
         source='',
         allow_unstaged_config=False,
+        hook_stage='commit'
 ):
     # These are mutually exclusive
     assert not (all_files and files)
@@ -68,6 +69,7 @@ def _get_opts(
         color=color,
         verbose=verbose,
         hook=hook,
+        hook_stage=hook_stage,
         no_stash=no_stash,
         origin=origin,
         source=source,
@@ -89,6 +91,7 @@ def _test_run(repo, options, expected_outputs, expected_ret, stage):
         stage_a_file()
     args = _get_opts(**options)
     ret, printed = _do_run(repo, args)
+
     assert ret == expected_ret, (ret, expected_ret, printed)
     for expected_output_part in expected_outputs:
         assert expected_output_part in printed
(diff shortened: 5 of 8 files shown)


pre_commit/resources/hook-tmpl
```diff
@@ -8,6 +8,7 @@ popd > /dev/null
 
 retv=0
 args=""
+hook_name="pre-commit"
```
chriskuehl: I think `hook_stage` would be a better name for this

pre_commit/main.py
```diff
@@ -101,6 +100,10 @@ def main(argv=None):
         help='Allow an unstaged config to be present.  Note that this will'
         'be stashed before parsing unless --no-stash is specified'
     )
+    run_parser.add_argument(
+        '--hook-name',
+        help='The name of the hook running the command e.g. pre-commit'
```
chriskuehl: similarly, maybe call it stage?

pre_commit/resources/hook-tmpl
barrysteyn: Sure thing.

pre_commit/main.py
barrysteyn: Okay, I will do this now.

barrysteyn:
Okay, I added your suggestions. I need to go to a meeting, but after that, I will add some unit tests and update the documentation.

barrysteyn:
@chriskuehl The tests are failing, but they are failing for things I have not touched. For instance, in the `install_uninstall_test.py`, there is a function `_get_commit_output` which has an exit code of 1. This is just an example.

Please can you run the tests from this pull requests and tell me what is wrong.

pre_commit/resources/hook-tmpl
```diff
@@ -40,6 +41,7 @@ fi
 {pre_push}
 
 # Run pre-commit
+args="$args --hook-stage $hook_stage"
```
asottile: this is not quite right, it'll end up calling:

`pre-commit --hook-stage commit`

when it needs to call

`pre-commit run --hook-stage commit`

pre_commit/resources/hook-tmpl
```diff
@@ -8,6 +8,7 @@ popd > /dev/null
```
asottile: I can't comment quite high enough in this file, but since you're changing the contents you'll need to update the md5 blob at the top and update the list in pre_commit/commands/install_uninstall.py

pre_commit/commands/run.py
```diff
@@ -183,4 +182,11 @@ def run(runner, args, write=sys_stdout_write_wrapper, environ=os.environ):
             if not repo_hooks:
                 write('No hook with id `{0}`\n'.format(args.hook))
                 return 1
+       
+        if args.hook_stage:
+            repo_hooks = [
+                (repo, hook) for repo, hook in repo_hooks
+                if not hook['stages'] or args.hook_stage in hook['stages']
+            ]
```
asottile: I'd like to see some sample output for how this changes things :)

tests/commands/run_test.py
```diff
@@ -59,6 +59,7 @@ def _get_opts(
         origin='',
         source='',
         allow_unstaged_config=False,
+        hook_stage=[]
```
asottile: probably not a bug in this case, but default values that are mutable can cause issues:

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

pre_commit/clientlib/validate_manifest.py
```diff
@@ -25,6 +25,13 @@ class InvalidManifestError(ValueError):
             'language_version': {'type': 'string', 'default': 'default'},
             'files': {'type': 'string'},
             'expected_return_value': {'type': 'number', 'default': 0},
+            'stages': {
+                'type': 'array',
+                'default': [],
```
asottile: shouldn't this default to `['commit', 'push']` ?

pre_commit/main.py
```diff
@@ -101,6 +100,10 @@ def main(argv=None):
         help='Allow an unstaged config to be present.  Note that this will'
         'be stashed before parsing unless --no-stash is specified'
     )
+    run_parser.add_argument(
+        '--hook-stage',
```
asottile: This'll break invocations that are just `pre-commit` and `pre-commit run`.  Maybe make it default to `'commit'`?  This might also make the bash changes easier.

pre_commit/clientlib/validate_manifest.py
barrysteyn: Well, the way the logic is written, if it is an empty array, it will always be run in every stage. To me, this makes it more future proof since if we make another stage (say rebase), then it would still work on every stage by default. Of course, it is far more expressive to default it like you have suggested. I really don't mind changing it (this is not my project after all). Is your take on this still the same after my explanation?

pre_commit/resources/hook-tmpl
barrysteyn: I don't understand why this would happen? Is it because if there are no arguments, then it defaults to run? If that is the case, then why is there an `args` variable in the first place?

barrysteyn: Also, I should note that when I run the code, `--hook-stage` variable is passed to the `run` method.

pre_commit/commands/run.py
barrysteyn: Right. And that sample output request would be in a test. The only problem now is that I can't get the previous tests to work. Once they are working, then I will be happy to test this new functionality.

pre_commit/main.py
barrysteyn: Good catch! Thanks @asottile. Dunno why I missed this.

tests/commands/run_test.py
barrysteyn: Nice. Thanks for this.

pre_commit/resources/hook-tmpl
barrysteyn: The file says it is a random(ish) md5. How is it created? How should I update it?

pre_commit/resources/hook-tmpl
asottile: Yeah `pre-commit` alone defaults to `pre-commit run`.  Mostly to make the normal case *just work*.  pre-push was added later.

pre_commit/resources/hook-tmpl
asottile: I usually `md5sum setup.py`, but it really could be anything, just needs to be unique :)

pre_commit/clientlib/validate_manifest.py
asottile: ah yeah I guess that's fair.  The tradeoff I see here is slightly easier to debug here but needs a test to keep it from regressing vs an extra branch in the other code.  I usually lean towards the first but this is fine as long as the code executes correctly.

pre_commit/resources/hook-tmpl
barrysteyn: Okay, but then regarding your comment above, do you mean to change it such that `run` is called explicitly:

`pre-commit run --hook-stage commit|push`

So I would add run?

asottile: Yeah, I think for pre-push this is already the case, you'll just need to add it for pre-commit (or if you make it default to `commit`, you can probably avoid needing to edit args in this file)

barrysteyn:
Thanks @asottile, tests are passing. I was wondering if you have any input regarding unit tests that I am about to add?

pre_commit/commands/run.py
```diff
@@ -183,4 +184,12 @@ def run(runner, args, write=sys_stdout_write_wrapper, environ=os.environ):
             if not repo_hooks:
                 write('No hook with id `{0}`\n'.format(args.hook))
                 return 1
+
+        # Filter hooks for stages
+        repo_hooks = [
+            (repo, hook) for repo, hook in repo_hooks
+            if (not hook['stages'] or
+                args.hook_stage in map(str.strip, hook['stages']))
```
asottile: why is strip needed here?

asottile:
Tests seem good :)  python3 is unhappy though

barrysteyn:
I'll fix Python 3 tomorrow. I used strip because I thought it wise to
remove whitespace from stage items, but it's really not needed

barrysteyn:
Hi

I believe the tests to be working now. They failed on travis dues to travis not being able to download a requirement (i.e. it is not failing because a test fails) for python 3.x

That said, all tests pass on my local machine. Is there anyway this can be resolved. The functionality is ready and waiting to be merged.

barrysteyn:
Okay, tests are working for the new functionality, both for python 3 and python 2. Please can I ask that this be merged - I would really need to start using the functionality. Also, once merged, I will update the documentation.

asottile:
@barrysteyn Thanks for the PR :D

I'll push a new version now

barrysteyn:
Thanks @asottile - can I update the documentation (pre-commit webpage) or is there someone dedicated to doing that (I really don't mind).

Also, I have some suggestions regarding the future of the project. Shout if you want me to chat about them.

Lastly, we need some documentation for journey-man python programmers like myself. Getting the tests to run locally was challenging due to lack of documentation. Maybe just a tiny addition to the README explaining how to run the tests. Which also brings me to another point: One of the tests (the repository tests) always failed on my local machine (it just seemed to stall). I never went into why this is happening, but it should be fixed.

asottile:
Sure, the docs are updated through https://github.com/pre-commit/pre-commit.github.io

As for the repository test, the end-to-end tests for ruby / javascript are painfully slow (building ruby from source when the binary distribution isn't available is hella slow) (probably looks like they've stalled).  Because of this I made a way to skip them (`slowtests=false make test`).  I can document this.  For now I've created #274

As for the future of the project, email is probably the best place for that :)

asottile:
fwiw #275 should make one of the tests faster if you're running 14.04

barrysteyn:
FYI, this issue has been solved in #271

PR #277, opened 2015-10-08
Fix bug - pushing on an empty changeset

barrysteyn:
There is a problem when pushing an empty changeset - the command `pre-commit --hook-stage push` will be run. But actually, if the changeset is empty, there is nothing to push (everything is up to date). So we should just exit.


barrysteyn:
@asottile If we try push an empty change set, `pre-commit --hook-stage push` will be run resulting in an error. I have submitted a PR to fix this. If you think there is a better way to do this, then please change.

asottile:
I PR'd your PR :)

asottile:
@barrysteyn  the nice bit about me PRing your PR is I can just merge it now :P

asottile:
This is available in version 0.6.1

Issue #278, opened 2015-10-14
Cannot install ruby hooks on OS X


magicmark:
I can install pre-commit fine, but when it comes to running it for the first time, I get this error:

```
[INFO] Installing environment for git@***/pre-commit/mirrors-scss-lint.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: Command: ['bash', '-c', '. /Users/markl/.pre-commit/repoLPZw60/rbenv-1.9.3-p484/bin/activate && cd /Users/markl/.pre-commit/repoLPZw60/ && gem build *.gemspec && gem install --no-document *.gem']
Return code: 1
Expected return code: 0
Output:
      Successfully built RubyGem
      Name: __fake_gem
      Version: 0.0.0
      File: __fake_gem-0.0.0.gem
   
Errors:
    WARNING:  no email specified
    WARNING:  no homepage specified
    WARNING:  description and summary are identical
    ERROR:  While executing gem ... (OptionParser::InvalidOption)
        invalid option: --no-document
   
 
Check the log at ~/.pre-commit/pre-commit.log
```

Environment details:
OS X 10.10.5
`gem --version`: 2.4.8
`ruby --version`: ruby 2.0.0p481 (2014-05-08 revision 45883) [universal.x86_64-darwin14]

asottile:
The version that rbenv hands us on osx is ancient:

```
(rbenv) ... $ gem --version
1.8.23
```

Seems we can switch from `--no-document` to `--no-ri` `--no-rdoc` which is less future-proof but reasonable

@Lucas-C introduced this and may have  opinions :)

Lucas-C:
Seems perfectly reasonable to me.

PR #279, opened 2015-10-14
Changed to use --no-ri --no-rdoc

magicmark:
Changed to use --no-ri --no-rdoc to fix gem installs on OS X


asottile:
Closes #278

PR #292, opened 2015-11-17
Allow args for pcre hook

tdeo:
This allows to specify a list of args for pcre hook. I'd like to use pcre hook for multiline match, which can be achieved by adding a `-z` option to `grep`.

### DIFF
--- a/pre_commit/languages/pcre.py
+++ b/pre_commit/languages/pcre.py
@@ -25,7 +25,9 @@ def run_hook(repo_cmd_runner, hook, file_args):
             'xargs', '-0', 'sh', '-c',
             # Grep usually returns 0 for matches, and nonzero for non-matches
             # so we flip it here.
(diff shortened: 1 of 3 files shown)


asottile:
Looks great, could you add a test for this?

I'll figure out the appveyor derpiness :/

tdeo:
I just amended the commit to add a test. However, because I'm using a Mac, it is highly painful to run the tests. I'll let travis check it up.

tdeo:
@asottile I got a new integration test passing using the `-z` option for grep. Is that enough?

asottile:
Thanks again for the PR!

tdeo:
Thanks! When do you plan to do next release?

asottile:
I'll do that today when I get to work :)

asottile:
This is available as 0.6.5

tdeo:
Thanks you!

PR #295, opened 2015-11-19
Added the additional_dependencies config parameter

blarghmatey:
Added the ability to specify additional dependencies to be installed in
the pre-commit environment.

### DIFF
--- a/pre_commit/languages/node.py
+++ b/pre_commit/languages/node.py
@@ -5,6 +5,7 @@ import sys
 
 from pre_commit.languages import helpers
 from pre_commit.util import clean_path_on_failure
+from pre_commit.util import shell_escape
 
 
 ENVIRONMENT_DIR = 'node_env'
@@ -44,8 +45,11 @@ def install_environment(repo_cmd_runner,
         with in_env(repo_cmd_runner, version) as node_env:
             node_env.run("cd '{prefix}' && npm install -g")
             if additional_dependencies:
-                node_env.run("cd '{prefix}' && npm install -g " +
-                             ' '.join(additional_dependencies))
+                node_env.run("cd '{prefix}' && npm install -g {deps}".format(
+                    ' '.join(
+                        [shell_escape(dep) for dep in additional_dependencies]
+                    )
+                ))
 
 
 def run_hook(repo_cmd_runner, hook, file_args):
--- a/pre_commit/languages/node.py
+++ b/pre_commit/languages/node.py
@@ -23,7 +23,8 @@ def in_env(repo_cmd_runner, language_version):
     yield NodeEnv(repo_cmd_runner, language_version)
 
 
-def install_environment(repo_cmd_runner, version='default'):
+def install_environment(repo_cmd_runner, version='default',
+                        additional_dependencies=None):
     assert repo_cmd_runner.exists('package.json')
     directory = helpers.environment_dir(ENVIRONMENT_DIR, version)
 
@@ -41,6 +42,9 @@ def install_environment(repo_cmd_runner, version='default'):
 
         with in_env(repo_cmd_runner, version) as node_env:
             node_env.run("cd '{prefix}' && npm install -g")
+            if additional_dependencies:
+                node_env.run("cd '{prefix}' && npm install -g " +
+                             (' ').join(additional_dependencies))
 
 
 def run_hook(repo_cmd_runner, hook, file_args):
--- a/pre_commit/languages/node.py
+++ b/pre_commit/languages/node.py
@@ -47,11 +47,13 @@ def install_environment(
         with in_env(repo_cmd_runner, version) as node_env:
             node_env.run("cd '{prefix}' && npm install -g")
             if additional_dependencies:
-                node_env.run("cd '{prefix}' && npm install -g {deps}".format(
+                node_env.run(
+                    "cd '{prefix}' && npm install -g " +
                     ' '.join(
-                        [shell_escape(dep) for dep in additional_dependencies]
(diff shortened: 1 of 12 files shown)


blarghmatey:
I'm opening this pull request now while I add tests in order to get some feedback before going much further with it.

pre_commit/languages/node.py
```diff
@@ -23,7 +23,8 @@ def in_env(repo_cmd_runner, language_version):
     yield NodeEnv(repo_cmd_runner, language_version)
 
 
-def install_environment(repo_cmd_runner, version='default'):
+def install_environment(repo_cmd_runner, version='default',
+                        additional_dependencies=None):
```
asottile: minor nit.  I'd format this like:

```python
def install_environment(
    repo_cmd_runner,
    version='default',
    additional_dependencies=(),
):
```

pre_commit/languages/node.py
```diff
@@ -41,6 +42,9 @@ def install_environment(repo_cmd_runner, version='default'):
 
         with in_env(repo_cmd_runner, version) as node_env:
             node_env.run("cd '{prefix}' && npm install -g")
+            if additional_dependencies:
+                node_env.run("cd '{prefix}' && npm install -g " +
+                             (' ').join(additional_dependencies))
```
asottile: The parens around the string are unnecessary :)

pre_commit/languages/python.py
```diff
@@ -57,6 +58,9 @@ def install_environment(repo_cmd_runner, version='default'):
         repo_cmd_runner.run(venv_cmd)
         with in_env(repo_cmd_runner, version) as env:
             env.run("cd '{prefix}' && pip install .")
+            if additional_dependencies:
+                env.run("cd '{prefix}' && pip install -U" +
```
asottile: I wouldn't use `--upgrade`, it has a tendency to cause incompatibilities with the libraries installed.  I think `pip install ...` should be sufficient

pre_commit/repository.py
```diff
@@ -50,6 +51,14 @@ def languages(self):
         )
 
     @cached_property
+    def additional_dependencies(self):
+        dep_dict = defaultdict(lambda: defaultdict(set))
+        for _, hook in self.hooks:
+            dep_dict[hook['language']][hook['language_version']].update(
+                hook.get('dependencies', []))
```
asottile: seems good to me :)

asottile:
Looks great, don't worry too much about appveyor, their stuff is broken right now :/

pre_commit/languages/python.py
blarghmatey: My thinking is that it would allow for updating dependencies on subsequent runs, but I realized that without adding extra methods to verify the state of installed dependencies and run upgrades/additional installs then that wouldn't make sense anyway. In the current state, adding new dependencies requires `pre-commit clean` followed by a subsequent run to rebuild the environment.

pre_commit/languages/python.py
```diff
@@ -57,6 +59,9 @@ def install_environment(repo_cmd_runner, version='default'):
         repo_cmd_runner.run(venv_cmd)
         with in_env(repo_cmd_runner, version) as env:
             env.run("cd '{prefix}' && pip install .")
+            if additional_dependencies:
+                env.run("cd '{prefix}' && pip install " +
+                        (' ').join(additional_dependencies))
```
asottile: just thought about this, these'll need quoting in the general case:

```python
env.run(
    'cd '{{prefix}}' && '
    'pip install {deps}'.format(
        ' '.join(pipes.quote(dep) for dep in additional_dependencies),
    ),
)

asottile: errr, `pre_commit.util.shell_escape` instead of `pipes.quote`

blarghmatey:
Any advice on where I should be looking at for adding tests to address the coverage issue on 2.6?

asottile:
@blarghmatey you'll want to add a test that triggers the `additional_dependencies` behaviour for each of the three languages (the coverage it is complaining about).

py26 is notably touchy due to some coverage bugs around context managers that are hard to avoid / factor out (which is why the minimum is set to 99% anyway).

You can probably add the tests to `tests/repository_test.py`.  You'll want to augment `def _test_hook_repo` to support additional_dependencies I think?  Or use the bits that that guy calls in order to write your own helper.

I also sent a PR for you: https://github.com/blarghmatey/pre-commit/pull/1

:)

blarghmatey:
@asottile Can you duplicate the test failures on your machine? They run successfully on my laptop so I'm not exactly sure how to fix them.

asottile:
Yep, I'll run through these :)

asottile:
I opened a PR against your branch: https://github.com/blarghmatey/pre-commit/pull/2

asottile:
One last thing, let's add these to the json schemas (pre_commit/clientlib/validate_{manifest,config}.py

And then :shipit:!

blarghmatey:
@asottile Hope you had a good weekend! I added the `addtional_dependencies` attribute to the config schema. I didn't add it to the manifest schema because it didn't seem relevant in that context since the `setup.py`, `package.json`, etc. files should handle the dependencies in those repos. Happy to add it in if you think it should be there.

asottile:
@blarghmatey sounds good.  I'll merge this when I get into work (I'll probably spec them in manifest anyway since I have some usecases where I'll want them there)

Thanks for the awesome PR!   Please add your self (if you'd like) on https://github.com/pre-commit/pre-commit.github.io

blarghmatey:
@asottile Awesome! I'll also add in the new attribute to the docs on the site.

PR #297, opened 2015-11-19
Don't run flake8 twice

asottile:
Looks like this started running twice since @chriskuehl added pre-commit running pre-commit in #252

fixd!


chriskuehl:
whoops



(shipit)

PR #305, opened 2015-11-25
Output a message when a hook fails due to file modification

jdb8:
For some linters such as eslint, there is no output on a successful --fix operation. This provides a basic message to the user which indicates that their hook failed due to a file modification.

Feel free to change the text to something better, but I wanted to keep it generic since perhaps some hooks may not be 'fixing' anything - ideally the text for eslint would say 'Files were fixed in this hook' or 'Fixing files...'. Another option here could be 'Modifying files...' to match similar text in hooks such as autopep8 (which uses the same style in 'Fixing ___').

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -89,8 +89,10 @@ def _run_single_hook(hook, repo, args, write, skips=frozenset()):
     retcode, stdout, stderr = repo.run_hook(hook, filenames)
     diff_after = cmd_output('git', 'diff', retcode=None, encoding=None)
 
+    file_modifications = diff_before != diff_after
+
     # If the hook makes changes, fail the commit
-    if diff_before != diff_after:
+    if file_modifications:
         retcode = 1
 
     if retcode:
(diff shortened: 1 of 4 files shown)


pre_commit/commands/run.py
```diff
@@ -104,9 +106,19 @@ def _run_single_hook(hook, repo, args, write, skips=frozenset()):
...
         write('hookid: {0}\n'.format(hook['id']))
         write('\n')
+
+        # Print a message if failing due to file modifications
+        if file_modifications:
+            extra_text = (
+                ' Additional output:\n' if (stdout or stderr) else ''
+            )
+
+            write('Files were modified by this hook.{0}'.format(extra_text))
```
asottile: instead of formatting, I'd just do

```python
if stdout or stderr:
    write(' Additional output:')
```

This'll avoid the slightly-awkward double linebreak in this case

pre_commit/commands/run.py
```diff
@@ -89,8 +89,10 @@ def _run_single_hook(hook, repo, args, write, skips=frozenset()):
     retcode, stdout, stderr = repo.run_hook(hook, filenames)
     diff_after = cmd_output('git', 'diff', retcode=None, encoding=None)
 
+    file_modifications = (diff_before != diff_after)
```
asottile: unnecessary parens

asottile:
fixnship

pre_commit/commands/run.py
jdb8: I agree that just writing a line is nicer than formatting, but the double linebreak was actually deliberate to ensure that if we have additional output, we have the same spacing at the end of the hook's full output regardless of whether `stdout` or `stderr` exist:

```
eslint......................................................................................Failed
hookid: hook

Files were modified by this hook. Additional output:

Modified foo.py!

```

and for no `stdout` or `stderr`:

```
eslint......................................................................................Failed
hookid: hook

Files were modified by this hook.

```

That's unfortunately why the code is a little weird, since without the newline stuff we'd have two linebreaks at the end of the second case (since the final `write('\n')` is not surrounded by a check for `stdout` or `stderr`). Let me know if I'm missing a better way to avoid that.

PR #306, opened 2015-11-26
Produce a useful error message when hook id is not present.  Resolves #194

### LINKED ISSUE #194: Improve error message when attempting to run non-existent hook
Hook id in `/.pre-commit-config.yaml` doesn't exist in the included repository

(From https://github.com/pre-commit/pre-commit-hooks/issues/37)

It should probably suggest updating hooks or checking the spelling of the hookid

asottile:


chriskuehl:
lg2m

PR #316, opened 2015-12-07
Add an option to require a specific pre-commit version

### LINKED ISSUE #307: Hooks should have a way of specifying minimum pre-commit version to run
Obviously this is a bit of a chicken-and-egg problem until the feature exists, but afterwards this'll be useful.

For instance, the feature added in 0.6.6 (additional dependencies) would be useful to gate against older pre-commit versions where installation would be nonsensical.

This would be a fatal error that would require the user to upgrade pre-commit to use that hook.

I'd like to implement this first, and then change all of the mirror repositories to use the additional-deps approach.  This would alleviate the pains in #282 as well by eliminating the hacky mirroring technique that is currently required for node hooks.

asottile:
Resolves #307


chriskuehl:
this is pretty cool

PR #319, opened 2015-12-10
Make additional_dependencies rollforward safe

### LINKED ISSUE #311: additonal_dependencies isn't "rollback safe"
Using old pre-commit + a hook repo with `additional_dependencies` it'll happily create the repo without installing the additional dependencies.  Upon upgrading to a newer pre-commit, it doesn't know that the additional dependencies aren't installed yet and will happily attempt to run in there (usually causing an executable to not be found).  We need some way to signify when these have been installed in order for this to be rollable.  A workaround is to `pre-commit clean` when upgrading, but that kinda is not the best (and especially confusing).

asottile:
Resolves #311

### DIFF
--- a/pre_commit/repository.py
+++ b/pre_commit/repository.py
@@ -1,12 +1,16 @@
 from __future__ import unicode_literals
 
+import io
+import json
 import logging
+import os
 import shutil
 from collections import defaultdict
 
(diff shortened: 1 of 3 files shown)


chriskuehl:
this is pretty nifty actually

Issue #213, opened 2015-03-23
Does not work within submodules


quentindemetz:
I'm getting: 
```
An unexpected error has occurred: NotADirectoryError: [Errno 20] Not a directory: '/home/quentin/chef-repo/cookbooks/ssmtp-cookbook/.git/hooks/pre-commit'
```
chef-repo is my primary repository and ssmtp-cookbook a git submodule of that. 

**ssmtp-cookbook/.git file contents:**
```
gitdir: ../../.git/modules/cookbooks/ssmtp-cookbook
```

asottile:
What is your pwd?

quentindemetz:
app-cookbook

asottile:
Well you're not really supposed to develop inside of submodules (they aren't quite full git repos) but I guess we can support this workflow. The fix is a regression test and to mkdirp the hooks directory I'd imagine. A workaround is to manually make the hooks directory yourself.

asottile:
Ah I think I'm wrong here,  in newer git submodules have an entirely separate git directory from the repository.  This means our trick of searching upwards for the git dope won't work.  Apparently there is `git rev-parse --git-dir` which will get us what we want maybe?

quentindemetz:
I concur, `git rev-parse --git-dir` does yield the correct path to the git directory, including in my current submodule configuration:
```
$ git rev-parse --git-dir
/home/quentin/chef-repo/.git/modules/cookbooks/ssmtp-cookbook
```

Good find !

quentindemetz:
(oops)

asottile:
So I took an initial look into this, it's slightly more difficult than I initially thought since we (erroneously) depend on the git directory existing at the top level of the repository for several things (such as where to chdir right before running hooks).  Therefore this needs to be split into two concepts: git-dir and git-root.

PR #320, opened 2015-12-18
Use rev-parse --show-toplevel

asottile:
A first step towards #213

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -7,6 +7,7 @@ import os.path
 import re
 
 from pre_commit.errors import FatalError
+from pre_commit.util import CalledProcessError
 from pre_commit.util import cmd_output
 from pre_commit.util import memoize_by_cwd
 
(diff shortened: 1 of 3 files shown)


asottile:
I have a sneaking suspicion that this broke something at some point, but I can't remember why...

https://github.com/pre-commit/pre-commit/commit/bf457349b3a1ca16c47b6878c05c9a3c11ee712d

asottile:
Actually, I bet that was plumbum's fault and I just didn't know it at the time

testing/util.py
```diff
@@ -41,6 +41,14 @@ def get_head_sha(dir):
         return cmd_output('git', 'rev-parse', 'HEAD')[1].strip()
 
 
+def normpath2(path):
+    """Like os.path.normpath, but does lower() for windows.  This is obsolete
+    in python3 which does this already.
+    """
+    path = os.path.normpath(path)
+    return path.lower() if os.name == 'nt' else path
```
chriskuehl: does `normcase` do the right thing?
https://docs.python.org/2/library/os.path.html#os.path.normcase

although...

> On Unix and Mac OS X, this returns the path unchanged; on case-insensitive filesystems, it converts the path to lowercase.

this seems a little dubious, doesn't OS X use a case-insensitive fs by default?

chriskuehl:
since you're in the midwest



shipit

testing/util.py
asottile: oh yeah I could probably use normcase, osx should be doing the same though!

pre_commit/git.py
```diff
@@ -15,16 +16,12 @@
...
-            path = os.path.normpath(os.path.join(path, '../'))
-    raise FatalError(
-        'Called from outside of the gits. '
-        'Please cd to a git repository.'
-    )
+    try:
+        return cmd_output('git', 'rev-parse', '--show-toplevel')[1].strip()
+    except CalledProcessError:
+        raise FatalError(
+            'Called from outside of the gits.  Please cd to a git repository.'
```
chriskuehl: two spaces here intentional?

chriskuehl:
otherwise shipit

pre_commit/git.py
asottile: yeah

PR #329, opened 2016-01-07
[WIP] Add ability to target files by "type"

chriskuehl:
This adds a small file classification package which gives file tags like `{'file', 'executable', 'python'}`. These tags are inferred based on file mode (non-executable vs exec. vs submodule vs symlink), extension, and shebang.

Hooks can now define a list of types they want to target.

For example:

``` patch
 -   id: end-of-file-fixer
     name: Fix End of Files
     description: Ensures that a file is either empty, or ends with one newline.
     entry: end-of-file-fixer
     language: python
-    files: \.(c|cpp|html|erb|slim|haml|ejs|jade|js|coffee|json|rb|md|py|css|scss|less|sh|tmpl|txt|yaml|yml|pp)$
+    types: [text]
 -   id: check-case-conflict
     name: Check for case conflicts
     description: Check for files that would conflict in case-insensitive filesystems
     entry: check-case-conflict
     language: python
-    # Match all files
-    files: ''
+    types: [file]
 -   id: check-docstring-first
     name: Check docstring is first
     description: Checks a common error of defining a docstring after code.
     entry: check-docstring-first
     language: python
-    files: \.py$
+    types: [python]
```

Key benefits:
- Can target files without extensions but which have shebangs
- Avoids potentially dangerous targeting of symlinks by hooks by default
- Fixes targeting submodules which causes some of the official hooks to break
- Can target `text` or `binary` files

There's a stupidly simple identify command you can use for now:

```
ckuehl@supernova:~/ocf-proj/ocfweb$ pre-commit identify setup.py
{'file', 'nonexecutable', 'text', 'python'}
ckuehl@supernova:~/ocf-proj/ocfweb$ pre-commit identify manage.py
{'file', 'executable', 'text', 'python'}
ckuehl@supernova:~/ocf-proj/ocfweb$ pre-commit identify docs
{'symlink'}
ckuehl@supernova:~/ocf-proj/ocfweb$ pre-commit identify bootstrap-scss
{'submodule'}
ckuehl@supernova:~/ocf-proj/ocfweb$ pre-commit identify ocfweb/static/img/about/lab.png
{'file', 'nonexecutable', 'binary', 'png'}
```

Things we still need:
- Better list of extensions and interpreters
- Some tests. It'll be easy to get 100% coverage but should add some higher-level integration tests
- Better identify command

The commits are pretty atomic, that's probably the best way to review?

You can `pip install ckuehl-pre-commit-types==0.7.3.dev2` if you want to test without cloning. I've modified a bunch of repos to use the new types:
- The offical hooks: https://github.com/chriskuehl/pre-commit-hooks/commit/cd3f867cafee3488a93a13f97bf7fdab329c996e
- @asottile's reorder-python-imports: https://github.com/chriskuehl/reorder_python_imports/commit/138f897f7d53c53dd4535dc377764fa3a58e310c
- @Lucas-C's pre-commit hooks: https://github.com/chriskuehl/pre-commit-hooks-1/commit/cb3b03ba29a70be8ff4702429ea331cfc61b87c8
- My puppet hooks: https://github.com/chriskuehl/puppet-pre-commit-hooks/commit/95af1e1e303c764f375abafa96702c605fccfdaa


pre_commit/commands/run.py
```diff
@@ -55,24 +55,37 @@ def get_changed_files(new, old):
...
+    return getter(include_expr, exclude_expr, types)
 
 
 def _run_single_hook(hook, repo, args, write, skips=frozenset()):
-    filenames = get_filenames(args, hook['files'], hook['exclude'])
+    filenames = get_filenames(
+        args,
+        hook['files'],
+        hook['exclude'],
+        frozenset(hook['types']) if 'types' in hook else frozenset(['file']),
```
asottile: since you specified a default, I think you don't need the branch here

pre_commit/file_classifier/classifier.py
```diff
@@ -0,0 +1,142 @@
+# encoding: utf-8
```
asottile: herp, can you add the `fix-encoding-pragma` hook from `pre-commit-hooks` and run on all the things?

pre_commit/file_classifier/classifier.py
```diff
@@ -0,0 +1,142 @@
...
+            if re.match(regex, interpreter)
+        )
+    else:
+        return []
+
+
+def _read_interpreter_from_shebang(path):
+    """Read an interpreter from a file's shebang.
+
+    The first line of a script is guaranteed to be ASCII, so we read ASCII
```
asottile: this is not entirely true, but it is the assumption we're going by to attempt to find the shebang :)

pre_commit/file_classifier/classifier.py
```diff
@@ -0,0 +1,142 @@
...
+    until we hit a newline (at which point we check if we read a valid shebang)
+    or a non-ASCII character (at which point we bail).
+
+    :param path: path to text file
+    :return: interpreter, or None if no shebang could be read
+    """
+    MAX_SHEBANG_LENGTH = 128  # Linux kernel limit on shebangs
```
asottile: cygwin lets higher than this (lol)

pre_commit/file_classifier/extensions.py
```diff
@@ -0,0 +1,28 @@
+"""List of known filename to file type mappings.
+
+The list consists of tuples of (filename regex, list of types).
+
+Most of these are extensions (e.g. *.py -> python), but some just use the
+filename (e.g. Makefile -> make).
+"""
+KNOWN_EXTENSIONS = [
```
asottile: I wonder if we can utilize mimetypes module at all:

```
>>> mimetypes.guess_type('foo.py')
('text/x-python', None)
```

tox.ini
```diff
@@ -10,7 +10,7 @@ commands =
     coverage erase
     coverage run -m pytest {posargs:tests}
     # TODO: when dropping py26, change to 100
-    coverage report --show-missing --fail-under 99
+    coverage report --show-missing
```
asottile: will have to fix ;)

asottile:
:D

pre_commit/commands/run.py
chriskuehl: good catch, fixed! this was from a previous iteration where i hadn't set a default

pre_commit/file_classifier/classifier.py
chriskuehl: yup

pre_commit/file_classifier/classifier.py
chriskuehl: clarified

Lucas-C:
Thanks for pinging me ont this.
I couldn't find time to review this over the week-end, but I'll try to get a look at it asap if it's not too late (this evening or tomorrow)

pre_commit/clientlib/validate_manifest.py
```diff
@@ -27,7 +28,11 @@ class InvalidManifestError(ValueError):
             'minimum_pre_commit_version': {
                 'type': 'string', 'default': '0.0.0',
             },
-            'files': {'type': 'string'},
+            'files': {
+                'type': 'string',
+                # empty regex to match all files
+                'default': '',
```
Lucas-C: Nice default

pre_commit/file_classifier/extensions.py
Lucas-C: I totally agree with this

Lucas-C:
OK, I finally took some time to review this. Sorry for the delay.
First: thanks for your efforts ! I think `pre-commit` really needs this feature.

Now, a couple of general considerations :
- why not using an existing file type detection library, like libmagic-based [python-magic](https://pypi.python.org/pypi/python-magic) and [filemagic](https://pypi.python.org/pypi/filemagic), or even the standard `mimetypes` module, instead of re-implementing a subset of that ?
- the idea to return a set of file-identification flags is great. I think it may be very helpful to also add the encoding: is it ASCII, utf8, latin1 ?

And a last thing : I'd like more tests for the `text` flag detection, especially with files using various encodings.

Overall this is really good work : nice job !

pre_commit/file_classifier/classifier.py
```diff
@@ -0,0 +1,143 @@
...
+    https://github.com/file/file/blob/master/src/encoding.c#L203-L228
+    """
+    text_chars = (
+        bytearray([7, 8, 9, 10, 12, 13, 27]) +
+        bytearray(range(0x20, 0x7F)) +
+        bytearray(range(0x80, 0X100))
+    )
+    with io.open(path, 'rb') as f:
+        b = f.read(1024)  # only read first KB
+    return bool(b.translate(None, text_chars))
```
Lucas-C: Why skipping 0x0b (VT) ?

Lucas-C:
I stumbled upon old issue #220 today.
You job in this PR would definitively solved it @chriskuehl : are still working on this ?

chriskuehl:
@Lucas-C yeah, would definitely still like to ship this at some point. I agree with the comments on this PR though, just need to implement them.

jakumeowski:
This will also fix handling of merge conflicts in submodules: https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/check_merge_conflict.py#L33
```
09:55 $ git commit
[INFO] Checking merge-conflict files only.
Check for merge conflicts......................................................................................................................................................Failed
hookid: check-merge-conflict

  File "/Users/tristan/.pre-commit/repoPTewkG/py_env-python2.7/bin/check-merge-conflict", line 9, in <module>
    load_entry_point('pre-commit-hooks==0.5.1', 'console_scripts', 'check-merge-conflict')()
  File "/Users/tristan/.pre-commit/repoPTewkG/py_env-python2.7/lib/python2.7/site-packages/pre_commit_hooks/check_merge_conflict.py", line 33, in detect_merge_conflict
    with open(filename) as inputfile:
IOError: [Errno 21] Is a directory: 'localizations'
```

Lucas-C:
@chriskuehl: I'd really like to see this issue closed, do you mind if I take over the work you started ?

chriskuehl:
@Lucas-C nope, I don't mind. I think it's pretty stable (me and a few other people have been using a version of the branch consistently for about a year), but I agree with the comments above.

chriskuehl:
I spoke with @asottile a little about how we might resolve the issues on this branch and get it shipped. There's a couple minor things to resolve, but I think the main design questions are:

**To reduce the need to map file extensions to file tags:**

We can use the `mimetypes` module. This will get us a mimetype like `text/x-python` which we can use some logic to try to simplify (e.g. removing the `<type>/` and `x-`). It might be a bit more complicated than that (especially for non-`text/`), but I think we can do a pretty good job here.

There are also some more complicated ones, e.g. `.c` is `text/x-csrc`.

The bigger concern will be things that `mimetypes` can't identify. There are some common extensions like `.go` that python3.5 stdlib doesn't seem to identify. Because of this, I think we're still going to have to maintain a mapping of extensions to tags for some extensions (or use something besides `mimetypes`).

We might also have a problem here where newer versions of stdlib can identify some extensions that older ones can't?


**To reduce the need to map shebang interpreters to file tags:**

This one we can't use `mimetypes` for, but we can probably use some logic to do the right thing in most cases. For example, if we strip off things that look like versions, that would take care of `python2.7 => python`.

There's still a couple special cases like `('^(ba|da|tc|[ckz])?sh$', ['shell']),` and `('^node(js)?$', ['javascript']),`. Are those important enough to still maintain a mapping? Not sure.

chriskuehl:
Will also want to remove the shebang parsing from this branch (it pre-dates the shebang parsing that pre-commit master now contains, we can just use that now).

chriskuehl:
There are two things we need from a "file identification library":

* Get tags for a file based on its name and git attributes
* Get tags for a file based on its interpreter name (read from the shebang)

To make this branch a little easier to merge, I'm going to start with the first and maybe do the second in a follow-up.

We talked about using the `mimetypes` module, but I think the consensus is that it's probably not going to work well. It would be a pretty confusing experience when different versions of Python recognize files differently, and even recent releases of Python 3.x can't identify common extensions like `.go`.

How about making a new package called `identify` or something and adding it as a dependency of pre-commit? This avoids maintaining a big mapping in pre-commit, and means we could accept updates to it without releasing new versions of pre-commit all the time.

@asottile sound reasonable? Should we have this library be "part" of pre-commit (maybe another repo under the pre-commit org) or just a regular dependency one of us happens to maintain?

asottile:
I had the same thought: a separate library that's a dependency.  In or out of the org is fine.

chriskuehl:
I put a draft of the API (just the README) at https://github.com/chriskuehl/identify/pull/1, lmk what you think @asottile @Lucas-C

chriskuehl:
I released version 0.0.1 of the identify library. The API is fully implemented and it understands a small number of shebangs and file extensions. This branch is hopelessly conflicting, rather than force-push an entirely new one on top and lose all the history, I'll just close this and open a new PR.

PR #333, opened 2016-01-15
include checking symlinks

bchess:
os.path.exists -> os.path.lexists so that we also check symlink files

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -89,7 +89,7 @@ def get_files_matching(all_file_list_strategy):
             if (
                 include_regex.search(filename) and
                 not exclude_regex.search(filename) and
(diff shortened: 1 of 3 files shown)


asottile:
:shipit: once appveyor is done I'll merge and cut a version for ya :)

chriskuehl:
this is kinda dangerous if you have symlinks pointing outside the repo, most hooks won't expect that :(

hopefully #329 will fix that, i'll try to finish up remaining issues soon

asottile:
It's no less dangerous than *working* symlinks which point outside the repo (status quo)

chriskuehl:
@asottile you're right, misinterpreted this change

asottile:
This is available in v0.7.5

PR #338, opened 2016-01-23
Support repos without hooks.yaml by reading a local config

shaleh:
If a repo does not have a hooks.yaml any hook ids are then looked for in
.pre-commit-hooks-config.yaml in the repo that is under test. It has the same
format as a standard hooks.yaml.

The purpose of this change is to avoid needing to mirror a linter repo just
to call it. Long term, tools could be reference by a URL or other source than
a git repo making having a mirror repo even more of a hassle.

### DIFF
--- a/pre_commit/constants.py
+++ b/pre_commit/constants.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 CONFIG_FILE = '.pre-commit-config.yaml'
 
+EXTERNAL_MANIFEST_FILE = '.pre-commit-hooks-config.yaml'
 MANIFEST_FILE = 'hooks.yaml'
 
 YAML_DUMP_KWARGS = {
--- a/pre_commit/manifest.py
+++ b/pre_commit/manifest.py
@@ -8,17 +8,34 @@ import pre_commit.constants as C
 from pre_commit.clientlib.validate_manifest import load_manifest
 
 
-class Manifest(object):
-    def __init__(self, repo_path_getter):
-        self.repo_path_getter = repo_path_getter
+class BaseManifest(object):
+    def __init__(self):
+        pass
 
     @cached_property
     def manifest_contents(self):
-        manifest_path = os.path.join(
-            self.repo_path_getter.repo_path, C.MANIFEST_FILE,
-        )
-        return load_manifest(manifest_path)
+        return load_manifest(self.manifest_path)
 
     @cached_property
     def hooks(self):
         return dict((hook['id'], hook) for hook in self.manifest_contents)
+
+    def exists(self):
+        return os.path.isfile(self.manifest_path)
+
+
+class Manifest(BaseManifest):
+    def __init__(self, repo_path_getter):
+        super(Manifest, self).__init__()
+
+        self.repo_path_getter = repo_path_getter
+        self.manifest_path = os.path.join(
+            self.repo_path_getter.repo_path, C.MANIFEST_FILE,
+        )
+
+
+class ExternalManifest(BaseManifest):
+    def __init__(self, path):
+        super(ExternalManifest, self).__init__()
+        self.path = path
+        self.manifest_path = os.path.join(path, C.EXTERNAL_MANIFEST_FILE)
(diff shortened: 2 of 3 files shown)


shaleh:
'''
-   id: github-openstack-bandit
    name: Bandit security checks
    description: "Runs bandit over python source."
    entry: bandit
    language: python
    files: \.py$
'''

That is a sample pre-commit-hooks-config.yaml for the Bandit security checker from OpenStack.

This pull request does not have any unit tests or other niceties. I wanted to start the conversation before spending a ton of time on it.

@asottile to address your concern of duplicated info that could have been in hooks.yaml one solution would be to have a central hooks.yaml that could be downloaded. The location of this file could be specified as a yaml entry in the ``.pre-commit-config.yaml``.

As I mention in the commit message, I would like to see alternatives to ``repo`` that would point to just a URL or other non-VCS system. File checksums could be used to satisfy the version checks.

Ideally the PR for mimetypes is finished which would simplify hooks.yaml further. Simply indicating the language should be all that is needed. At that point even if the hook info is duplicated in each repo using pre-commit the amount if data is small -- the hook name, description, language, and name of the entrypoint.

shaleh:
If there is interest in this patch being accepted I will add more unit tests to check ``ExternalManifest``. The code is currently in the "works for me" state. I posted it for review and commentary. I did not want to write test code that would be removed by name/interface refactoring. Finite hacking time and all that :-)

asottile:
I was waiting for others to comment, I've already voiced my thoughts elsewhere but I'll reiterate them here.

I don't think this is a good idea for a few reasons:

- It's against the idea of sharing configurations with other individuals.  A definition of a hook should live in one place and consumers should references that single place
- It doesn't work for multiple repositories (two of these "phantom" repositories would reference the same hooks.yaml which would introduce available but broken hook ids to end users)
- Installation of these "phantom" hook repositories would be orphaned from their configuration, potentially leading to unexpected behaviour when using pre-commit across multiple repositories on the same machine
- It introduces yet-another-way to do things which would require maintenance

On the note of a central repository of hooks:
- One already exists: http://pre-commit.com/hooks.html or in consumable format http://pre-commit.com/all-hooks.json
- However, depending on this in a runtime situation would introduce a single-point-of-failure and would require organizations which want to use pre-commit to set up an internal "pre-commit repository" (this is a huge barrier to entry for many, especially enterprise).

On the note of non-VCS:
- This again goes against the direction of reusable hook definitions, but it also introduces significant logic into pre-commit to understand each and every package management system and the quirks of each of them

@chriskuehl @struys thoughts?

shaleh:
In corp land copying the yaml from tree to tree is easy and common place. An internal wiki or VCS makes this easier. Not every biz is GitHub savvy.

I see comments on the pre-commit site and elsewhere concerning getting linter and other tools to come on board. Requiring a file called 'hooks.yaml' is a barrier to entry and acceptance. Someone has to plead the case with every tool that exists and will come out. Not to mention "hooks.yaml" is not a specific enough name. Sure the phantom Git repos makes that less of an issue but still, that is a massive undertaking and it makes a bunch of tiny, mostly write once repos.

Now if there is a way to have pre-commit be told "all hooks are defined at this URL unless defined in the tool repo" that would be OK. That way if pre-commit.com goes away next week the tool can live on.

shaleh:
I guess part of it is I do not see what is hard for the user.

    -   id: github-openstack-bandit
        name: Bandit security checks
        description: "Runs bandit over python source."
        entry: bandit
        language: python
        files: \.py$

If I was running those in any other test environment I would need most of that info. It would be nice to assume that ``language: python`` was implied by file *.py. That would reduce a line. Or that ``language:python`` wants files that end in *.py but that could hurt some scripts maybe.

Putting those lines in a file allow my project to be tested by bandit. Quick, easy, understandable. What is the concern? Duplication of effort? For some bits of English text?

chriskuehl:
I think we can close this for now -- the reasons @asottile cited above are pretty strong and I don't think this is really something we need to add to pre-commit itself.

One thing that's worked really well for us internally is using one central repository with lots of Yelp-specific pre-commit hooks. Even if your org isn't big about creating lots of small repos on GitHub, you might find that model (one repo, lots of hooks) can work pretty well.

Also, you can hack a lot of this together with [local hooks](http://pre-commit.com/#repository-local-hooks) if you'd really like to do it per-project. (I know it's not a perfect solution, but I think a reasonable workaround.)

Thanks for the PR and starting the discussion!

PR #339, opened 2016-02-02
Making it possible to invoke `pre-commit run --files some.file` from a subdirectory of the repository

Lucas-C:
I included a new test (that does not pass with `pre-commit` current version) to illustrate the problem

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -8,6 +8,7 @@ import pkg_resources
 
 from pre_commit import color
 from pre_commit import five
+from pre_commit import git
 from pre_commit.commands.autoupdate import autoupdate
 from pre_commit.commands.clean import clean
 from pre_commit.commands.install_uninstall import install
@@ -110,7 +111,8 @@ def main(argv=None):
         help='Run on all the files in the repo.  Implies --no-stash.',
     )
     run_mutex_group.add_argument(
-        '--files', nargs='*', help='Specific filenames to run hooks on.',
+        '--files', nargs='*', default=[],
+        help='Specific filenames to run hooks on.',
     )
 
     help = subparsers.add_parser(
@@ -122,6 +124,11 @@ def main(argv=None):
     if len(argv) == 0:
         argv = ['run']
     args = parser.parse_args(argv)
+    if args.command == 'run':
+        args.files = [
+            os.path.relpath(os.path.abspath(filename), git.get_root())
+            for filename in args.files
+        ]
 
     if args.command == 'help':
         if args.help_cmd:
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -42,7 +42,7 @@ def repo_with_failing_hook(tempdir_factory):
         yield git_path
 
 
-def stage_a_file(filename='foo.py'):
+def stage_a_file(filename):
     cmd_output('touch', filename)
     cmd_output('git', 'add', filename)
 
@@ -83,14 +83,15 @@ def _do_run(repo, args, environ={}):
     runner = Runner(repo)
     write_mock = mock.Mock()
     write_fn = functools.partial(sys_stdout_write_wrapper, stream=write_mock)
-    ret = run(runner, args, write=write_fn, environ=environ)
+    with cwd(runner.git_root):  # replicates Runner.create behaviour
+        ret = run(runner, args, write=write_fn, environ=environ)
     printed = get_write_mock_output(write_mock)
     return ret, printed
 
 
-def _test_run(repo, options, expected_outputs, expected_ret, stage):
-    if stage:
-        stage_a_file()
+def _test_run(repo, options, expected_outputs, expected_ret, file_to_stage):
+    if file_to_stage:
+        stage_a_file(file_to_stage)
     args = _get_opts(**options)
     ret, printed = _do_run(repo, args)
 
@@ -112,14 +113,14 @@ def test_run_all_hooks_failing(
             b'Fail\nfoo.py\n',
         ),
         1,
-        True,
+        'foo.py',
     )
 
 
 def test_arbitrary_bytes_hook(tempdir_factory, mock_out_store_directory):
     git_path = make_consuming_repo(tempdir_factory, 'arbitrary_bytes_repo')
     with cwd(git_path):
-        _test_run(git_path, {}, (b'\xe2\x98\x83\xb2\n',), 1, True)
+        _test_run(git_path, {}, (b'\xe2\x98\x83\xb2\n',), 1, 'foo.py')
 
 
 def test_hook_that_modifies_but_returns_zero(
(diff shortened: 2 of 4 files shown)


asottile:
I think you might be able to simplify this to just:

```python
args.files = [
    os.path.relpath(os.path.abspath(filename), git.get_root())
    for filename in args.files
]
```

That'd eliminate the complexity added to pre_commit.run

Lucas-C:
Good idea, that's way smarter.
Done !

pre_commit/git.py
```diff
@@ -83,15 +84,16 @@ def get_files_matching(all_file_list_strategy):
...
         include_regex = re.compile(include_expr)
         exclude_regex = re.compile(exclude_expr)
-        return set(
-            filename
-            for filename in all_file_list_strategy()
-            if (
-                include_regex.search(filename) and
-                not exclude_regex.search(filename) and
-                os.path.lexists(filename)
+        with cwd(get_root()):  # because filenames can be relative
```
asottile: I don't think this is necessary, pre_commit.runner.Runner.create invariantly sets the working directory (and is the first thing invoked by pre-commit)

pre_commit/commands/run.py
```diff
@@ -61,7 +61,11 @@ def get_filenames(args, include_expr, exclude_expr):
             lambda: get_changed_files(args.origin, args.source),
         )
     elif args.files:
-        getter = git.get_files_matching(lambda: args.files)
+        relative_files = [
+            os.path.relpath(os.path.abspath(filename), git.get_root())
+            for filename in args.files
+        ]
```
asottile: I would just modify `args.files` in `pre_commit.main`, that'll make it less confusing for the rest of the program if it ever ends up using this argument in the future (and we reinvent this bug :D)

testing/resources/selective_script_hooks_repo/hooks.yaml
```diff
@@ -0,0 +1,5 @@
+-   id: bash_hook
+    name: Bash hook
+    entry: bin/hook.sh
+    language: script
+    files: '^subdir/'
```
asottile: you can probably just reuse the bash hook and modify the configuration to have this added on -- that said, I don't think this hook repo actually adds any coverage.  You really only need a test which cds and uses relative paths

tests/commands/run_test.py
```diff
@@ -88,12 +88,12 @@ def _do_run(repo, args, environ={}):
...
-def _test_run(repo, options, expected_outputs, expected_ret, stage):
-    if stage:
-        stage_a_file()
+def _test_run(repo, options, expected_outputs, expected_ret, file_to_stage):
+    if file_to_stage:
+        stage_a_file(file_to_stage)
     args = _get_opts(**options)
     ret, printed = _do_run(repo, args)
-
+    print(printed)
```
asottile: remove please :)

tests/commands/run_test.py
```diff
@@ -88,12 +88,12 @@ def _do_run(repo, args, environ={}):
     return ret, printed
 
 
-def _test_run(repo, options, expected_outputs, expected_ret, stage):
-    if stage:
-        stage_a_file()
+def _test_run(repo, options, expected_outputs, expected_ret, file_to_stage):
```
asottile: probably don't need this if you don't need the `files: ...` above.

I think you can still write a regression test with the bash hook, maybe not with this helper (which seems to have gotten too complicated :()

There's a few one-off tests at the bottom which this probably fits better into

tests/commands/run_test.py
```diff
@@ -595,3 +596,19 @@ def test_unstaged_message_suppressed(
...
+def test_invoke_precommit_in_subdir(
+        tempdir_factory, mock_out_store_directory,
+):
+    git_path = make_consuming_repo(tempdir_factory,
+                                   'selective_script_hooks_repo')
+    with cwd(git_path):
+        os.mkdir('subdir/')
+        with cwd('subdir/'):
+            cmd_output('touch', 'foo.py')
+            _test_run(git_path,
```
asottile: minor thing: https://github.com/pre-commit/pre-commit/blob/b1e6063e12242f68392eebc4c08427f2319719e7/CONTRIBUTING.md#multi-line-method-invocation

pre_commit/commands/run.py
Lucas-C: Agreed, that's what I did first, but then many of the tests would not include this logic: e.g. `tests/commands/run_test.py` uses the `Runner` class, and does not pass through `pre_commit.main`.
Hence my decision to put this code here, so it can be shared across more pieces of code.
What do you think ?

asottile: I'd rather put this in main where all the code will benefit from it -- might make the testing more difficult (though there are some similar tests you can snitch from)

Alternately, you could split this functionality into a `normalize_arguments` function and write a test for that (so you don't have to bother with integration testing)

pre_commit/git.py
Lucas-C: It is.
I tested, and at this point in the code, when running the test I added, here is what I observe:
```
os.getcwd() : '/tmp/user/1000/pytest-2/test_invoke_precommit_in_subdi0/2/subdir'
all_file_list_strategy() : [u'subdir/foo.py']
```

testing/resources/selective_script_hooks_repo/hooks.yaml
Lucas-C: No, I'd like a test with the `files: '^subdir/'` criteria if you're ok

tests/commands/run_test.py
Lucas-C: Ooops :) Done

tests/commands/run_test.py
Lucas-C: Ok, code style fixed

pre_commit/commands/run.py
Lucas-C: Ok, done

pre_commit/git.py
asottile: these are the tests being not-realistic, instead of patching the runtime code, fix the testtime code (sorry about this, it's probably my fault)

testing/resources/selective_script_hooks_repo/hooks.yaml
asottile: yeah, I'm saying you can do that without introducing a new repository by setting `files: ...` in the config file instead of the backing repo (which would allow you to reuse the script hooks repo without creating a one-off repo)

pre_commit/main.py
```diff
@@ -122,6 +123,11 @@ def main(argv=None):
     if len(argv) == 0:
         argv = ['run']
     args = parser.parse_args(argv)
+    if hasattr(args, 'files') and args.files:
```
asottile: `hasattr` will always be `True` here

Add `default=[]` to the definition of `--files` above and you won't need this branch at all :D

Lucas-C: Good idea

pre_commit/git.py
Lucas-C: Ok, I fixed `_do_run` in `tests/commands/run_test.py`

testing/resources/selective_script_hooks_repo/hooks.yaml
Lucas-C: Ok, done

pre_commit/main.py
asottile: I guess I was wrong here, it'll only have that attribute if `args.command == 'run'`, I'd prefer that over the hasasttr check :)

asottile:
I'm not sure what's up with the SSLErrors, I'll look into those on master -- I assume something with upgrading pip :/

pre_commit/main.py
Lucas-C: Ok, done

asottile:
You'll want to rebase on current master (to fix the SSL errors) and run `pre-commit run  --files pre_commit/main.py` to fix the import ordering :)

Lucas-C:
Done !

asottile:
Looks good, I'm going to merge this locally (I want to add a test and poke at the one you wrote (since I think it's only testing the test code))

Thanks again for an awesome fix!

asottile:
Via https://github.com/pre-commit/pre-commit/commit/72eeafe51667e2850eba26a082b1e1d62923eb94

Lucas-C:
Thanks for the feedback and accepting the PR !

asottile:
This has been included in v0.8.0

PR #346, opened 2016-02-21
Support terminal width on windows.  Resolves #199

### LINKED ISSUE #199: Windows: Terminal width support
We detect terminal width in unixlikes by running `tput cols`.  This works fine for those platforms but doesn't work well for windows.  Maybe find a package which does this logic for us and depend on that.

asottile:
Hit https://github.com/chrippa/backports.shutil_get_terminal_size/issues/4 while implementing.  Should be fixed entirely by my PR there (and worked around by `or 80` in this PR)

### DIFF
--- a/pre_commit/output.py
+++ b/pre_commit/output.py
@@ -1,27 +1,15 @@
 from __future__ import unicode_literals
 
-import os
-import subprocess
 import sys
+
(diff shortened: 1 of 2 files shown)


asottile:
Derp, this fixed cmd.exe but broke cygwin / mingw terminal -- fixing!

PR #348, opened 2016-02-22
Use pyterminalsize for terminal sizing

asottile:


chriskuehl:
rip windows tmux pane resizers

Issue #345, opened 2016-02-18
Maybe replace xargs with a python implementation


asottile:
There's quite a few bugs we've run into by relying on a valid implementation of xargs being installed.  Perhaps we should instead implement our own in python and call that. 

This would also allow us to do clever things with multi processing if we wanted to.

chriskuehl:
you probably just aren't using the right permutation of xargs' 40 command-line flags

asottile:
via #355

PR #355, opened 2016-03-21
Factor out bash and activate files

asottile:


asottile:
Seems I'll need to implement #345 to make this work on windows

asottile:
hmmm, one side-effect of factoring out xargs is shebangs aren't processed on windows... I guess I could try and implement that as well (ick)...

asottile:
https://github.com/polysquare/python-parse-shebang seems promising for fixing the shebang issues

asottile:
Soooooo close.  I seem to be blocked on getting full pcre support on windows: https://github.com/Alexpux/Cygwin/issues/15

chriskuehl:
(wait for it)

asottile:
Reported the pcre bug here: https://cygwin.com/ml/cygwin/2016-04/msg00157.html

PR #368, opened 2016-05-10
Set up logging handler in autoupdate

trbs:
When running ```autoupdate``` with a hook in .pre-commit.yaml that does not exist in the current version of the repository there is an error: ```No handlers could be found for logger "pre_commit"```

Before:
```
$ pre-commit autoupdate

Updating git://github.com/pre-commit/pre-commit-hooks...already up to date.
Updating git://github.com/trbs/pre-commit-hooks-trbs.git...No handlers could be found for logger "pre_commit"
```

After:
```
pre-commit autoupdate
Updating git://github.com/pre-commit/pre-commit-hooks...already up to date.
Updating git://github.com/trbs/pre-commit-hooks-trbs.git...[ERROR] `flake8` is not present in repository git://github.com/trbs/pre-commit-hooks-trbs.git.  Typo? Perhaps it is introduced in a newer version?  Often `pre-commit autoupdate` fixes this.
```

The error message is now a tiny bit misleading, since we are actually running ```autoupdate``` while this error is generated. Maybe change the wording a little bit so that it's clearer that the new hook can only be introduced in ```.pre-commit.yaml``` _after_ ```autoupdate``` has been ran.

### DIFF
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -1,18 +1,18 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import sys
 import logging
+import sys
 
 from aspy.yaml import ordered_dump
 from aspy.yaml import ordered_load
 
 import pre_commit.constants as C
-from pre_commit.logging_handler import LoggingHandler
 from pre_commit.clientlib.validate_config import CONFIG_JSON_SCHEMA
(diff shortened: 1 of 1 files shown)


asottile:
heh, looks like you didn't run the hooks when committing (awkward!) -- fixn:ship:

asottile:
Thanks!

PR #375, opened 2016-05-25
only consider forward diff in changed files

vinayinvicible:
Consider we have a branch A.
I create two branches B, C from A and make changes independently.
I merge changes from C into A.
Now if I run pre-commit between A and B, diff shows the changes from branch C also.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -51,7 +51,7 @@ def _print_user_skipped(hook, write, args):
 
 def get_changed_files(new, old):
     return cmd_output(
-        'git', 'diff', '--name-only', '{0}..{1}'.format(old, new),
+        'git', 'diff', '--name-only', '{0}...{1}'.format(old, new),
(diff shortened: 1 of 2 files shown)


asottile:
Cool! Thanks for the PR :D

Do you think you could write a regression test demonstrating this change? (either an integration test (similar to tests/commands/install_uninstall_test.py:test_pre_push_integration_accepted) or a unit test)

vinayinvicible:
Sure thing

vinayinvicible:
@asottile added the test.
I don't suppose failed checks are due to my changes

asottile:
yeah, doesn't seem like your fault -- I'll poke around with this and then merge :)

asottile:
Tracked down the issue: https://github.com/pypa/virtualenv/pull/926

I'm going to upgrade these tests to use py34 / py35 instead

asottile:
This has been released in 0.9.0

PR #376, opened 2016-05-25
Don't run on deleted files.  Resolves #374

### LINKED ISSUE #374: Newly gitignored (but file still exists) files are linted
(they should not be)

asottile:


chriskuehl:
looks solid to me!

Issue #31, opened 2014-03-19
Run hooks only on conflicting files in a merge conflict situation


asottile:
Most gripes I have with pre-commit hooks are "someone else doesn't have them enabled and I had to do their resolution during a merge conflict"

vinayinvicible:
@asottile 
I recently faced an issue with pre-commits during merge conflict.

git's STDOUT during merge conflict looks like this

> From github.com:`REPO`
>  \* branch            master     -> FETCH_HEAD
> Auto-merging `M_file1`
> Auto-merging `C_file1`
> CONFLICT (content): Merge conflict in `C_file1`
> Removing `R_file1`
> Auto-merging `M_file2`
> Auto-merging `C_file2`
> CONFLICT (content): Merge conflict in `C_file2`
> Auto-merging `M_file3`
> Auto-merging `M_file4`
> Automatic merge failed; fix conflicts and then commit the result.
> 

I haven't modified any other file than the conflict files.
I would expect pre-commits to run only on `C_file*` files.
But it was running on both `C_file*` and `M_file*` files.

I am not sure if this was intended behavior.

asottile:
Do you have a reproducible test case?

vinayinvicible:
@asottile I've raised a PR #388

vinayinvicible:
@asottile any updates on this

asottile:
Just commented on the PR, sorry it took me so long to get back to this -- been pretty busy :D

PR #388, opened 2016-06-30
modified fixture to include auto-merge files during conflicts

vinayinvicible:
this is in reference to the issue pointed out in #31

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,21 +42,35 @@ def in_tmpdir(tempdir_factory):
 
 
 def _make_conflict():
-    cmd_output('git', 'checkout', 'origin/master', '-b', 'foo')
+    cmd_output('git', 'checkout', 'origin/master', '-b', 'common_branch')
+    with io.open('common_file', 'w') as common_file:
+        common_file.write('common line\n')
+    cmd_output('git', 'add', 'common_file')
+    cmd_output('git', 'commit', '-m', 'common_file')
+    cmd_output('git', 'checkout', 'common_branch', '-b', 'foo')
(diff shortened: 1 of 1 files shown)


asottile:
So I did some research into this...

From what I can tell it is not possible to differentiate between auto-merged files and manually-merged files without actually re-running the merge (which would be expensive and fragile).  So on one point I don't think it's actually possible to exclude auto-merged files.

On the other hand, I don't think we want to exclude auto-merged files.  In the general case, both sides should have already used best practices and have passing hooks.  But also, it is often useful to identify implicit conflicts (one common one we hit _all the time_ at yelp is someone factors some import out of a file and someone else adds a new usage of that import.  Despite both passing the hooks and the merge succeeding, there's now an undefined reference to something that used to be defined.

Given that, how would you feel about me updating the test to incorporate this diff (just for documentation / spec purposes) and adding some documentation around this (since it is currently unspecified and unclear).

vinayinvicible:
that import example is a solid argument. :+1: 

you should definitely modify docs regarding this

vinayinvicible:
@asottile On second thought, shouldn't you be running pre-commit on pure merge commits also (same argument as yours')

asottile:
Nah you won't have 2 inputs to cause an implicit conflict

asottile:
ah, I see what you mean.  Yeah it should run on merge commits (that contain auto-merged files) but git doesn't give us a hook :(

asottile:
As discussed (and documented via https://github.com/pre-commit/pre-commit.github.io/pull/80)

PR #398, opened 2016-08-16
Clarify/correct error message

sanmai-NL:
The error also occurs if the `git` utility isn't available.


asottile:
:shipit:

rip gits.

Looks like we now have a strong incentive to remove python2.6 since flake8 has dropped support

PR #400, opened 2016-08-18
Fix staged-files-only with a non-utf8-trailing-whitespace diff.  Resolves #397

### LINKED ISSUE #397: Stashed changes lost if hook fails with non-UTF-8 diff containing trailing whitespace
Hi,

A colleague almost lost all the changes she was working on after launching a `git commit` (with zero file added) and `pre-commit` crashing without restoring its [patch](https://github.com/pre-commit/pre-commit/blob/master/pre_commit/staged_files_only.py#L15).

Here is the terminal message she got:

```
[WARNING] Stashed changes conflicted with hook auto-fixes... Rolling back fixes...
An unexpected error has occurred: CalledProcessError: Command: ['git', 'apply', 'C:\\Users\\toto\\.pre-commit\\patch1471341002']
```

This seems very similar to a past solved issue:
https://github.com/pre-commit/pre-commit/issues/176

I think it had to do with CRLF conversion.
I'm going to try to reproduce this.

asottile:
CC @Lucas-C

### DIFF
--- a/pre_commit/staged_files_only.py
+++ b/pre_commit/staged_files_only.py
@@ -45,7 +45,7 @@ def staged_files_only(cmd_runner):
         finally:
             # Try to apply the patch we saved
             try:
-                cmd_runner.run(['git', 'apply', patch_filename])
+                cmd_runner.run(('git', 'apply', patch_filename), encoding=None)
             except CalledProcessError:
                 logger.warning(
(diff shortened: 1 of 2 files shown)


Lucas-C:
This looks good to me
Thanks for the fix !

asottile:
This has been released in 0.9.0

PR #408, opened 2016-09-01
Use 80 or min width instead of terminal size

asottile:
This would factor out pyterminalsize and instead use 80 as a default width and expand if the hook names are long.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -30,25 +30,6 @@ def _hook_msg_start(hook, verbose):
     )
 
 
-def _print_no_files_skipped(hook, write, args):
-    write(get_hook_message(
-        _hook_msg_start(hook, args.verbose),
-        postfix='(no files to check) ',
-        end_msg='Skipped',
-        end_color=color.TURQUOISE,
-        use_color=args.color,
-    ))
-
-
-def _print_user_skipped(hook, write, args):
-    write(get_hook_message(
-        _hook_msg_start(hook, args.verbose),
-        end_msg='Skipped',
-        end_color=color.YELLOW,
-        use_color=args.color,
-    ))
-
-
 def get_changed_files(new, old):
     return cmd_output(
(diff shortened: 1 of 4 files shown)


asottile:
I'd love some input on this PR.

Here's some screenshots of before and after:

## Before

(Narrow terminal)



(80 character terminal)


(Wide terminal)


## After

(Narrow terminal)


(80 character terminal)


(Wide terminal)

chriskuehl:
I like this.

Some definitely positive things:
* Drops the only binary extension dependency
* The "narrow terminal" seems easier to parse. It's gonna look a little derpy any way it's formatted, but this seems like a clear improvement.

I was never crazy about matching the terminal width, since I resize windows all the time and it usually ends up being illegible shortly afterwards.

I think it's arguably also better UX on the "wide terminal" example, since following those dots across the entire width of the terminal is a bit hard. It might not be quite as aesthetically pleasing? But I don't really think it looks worse, just different.

Besides, if you use a 230-column-wide terminal, you should feel bad anyway :)

asottile:
@Lucas-C any thoughts on this one too?

Lucas-C:
Looks good to me. Better UX IMHO too.

pre_commit/commands/run.py
```diff
@@ -128,12 +128,32 @@ def _run_single_hook(hook, repo, args, write, skips=frozenset()):
...
+        name_len = 0
+
+    cols = name_len + 3 + len(NO_FILES) + 1 + len(SKIPPED)
+    return max(cols, 80)
+
+
 def _run_hooks(repo_hooks, args, write, environ):
     """Actually run the hooks."""
     skips = _get_skips(environ)
+    cols = _compute_cols([hook for _, hook in repo_hooks], args.verbose)
```
Lucas-C: Naming: `cols_count` ?

asottile:
I'm going to merge this, worst case people don't like it and we'll revert it later :)

PR #409, opened 2016-09-09
Fix file stashing with external diff tool

samsaggace:
When git is configured to use an external diff tool to show diffs
(eg. 'git config diff.external mytool'), the stashing unstaged files
will create an empty file that can't be recovered.

Some modifications are permanently lost...

Just disable the ext-diff of git diff to avoid any issue.

(Question to pre-commit devs: why not using git stash directly ?)

### DIFF
--- a/pre_commit/staged_files_only.py
+++ b/pre_commit/staged_files_only.py
@@ -23,7 +23,7 @@ def staged_files_only(cmd_runner):
     retcode, diff_stdout_binary, _ = cmd_runner.run(
         [
             'git', 'diff', '--ignore-submodules', '--binary', '--exit-code',
-            '--no-color',
+            '--no-color', '--no-ext-diff',
         ],
         retcode=None,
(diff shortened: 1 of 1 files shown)


asottile:
Looks good, want to write a quick regression test?  

We don't use stash because it both interferes with user stashes and can result in loss of data (especially when hooks change things which conflict with stashed changes)

samsaggace:
Hi @asottile,
Thanks for responding so quickly !
I would love to create a regression test but I'm not good at python so I'm not sure how to write it and where...

asottile:
Sure, I can handle that then!

samsaggace:
Thanks!

asottile:
I'll try and get a release out for you once I'm at a computer :)

asottile:
This is included in 0.9.1

samsaggace:
You rock ! :+1:

Issue #414, opened 2016-09-26
Configuration v2


asottile:
A bit of an RFC
## Issues

There's some "problems" with the current configuration and I'd like to address those in a backwards incompatible way (probably through a migration period where both formats are allowed).

There's two primary issues I'd like to address with the current configuration format:
- The top level is currently a list, making it difficult to add repository-level configuration: #240 #281 (as well as other things I'd like such as repository-level interpreter version)
- The configuration is not scriptably rewriteable: #210

There's also auxiliary issues:
- yaml is not maintained
- yaml has some security issues #396
- `hooks.yaml` isn't well namespaced
### A new configuration language

I shopped around for configuration languages which fit the criteria above, notably:
- Provides a non-C implementation
- Supports comments + maps + lists
- Supports round-trip rewritability
- Has a python implementaiton

And did not find anything.  Given that, I've started speccing and implementing [a configuration language](https://github.com/asottile/dumbconf) with those design goals in mind.  It also aims to be significantly simpler than yaml (the spec may not fit on a postcard like json, but it's also not 70+ pages like yaml).
### New filenames

The names of the files will need to change for two reasons:
- The configuration will no longer be yaml
- hooks.yaml is not a namespaced-enough name

Proposing:
- `.pre-commit-config` - describes the consumer configuration (previously `.pre-commit-config.yaml`)
- `.pre-commit-manifest` - describes the repository configuration (previously `hooks.yaml`)
### New configuration structure

The new configuration structure will be a top-level mapping type to allow repository-wide configuration.  This map will have a `hooks` key which will contain the current configuration.

``` yaml
hooks: [
        ...,
]
```

This will allow for top-level configuration which will apply to all hooks (such as global excludes, interpreter versions, etc.)
## Migration
- The new configuration will be introduced.
- In that release and the next two releases (or the next 3 months, whichever comes sooner), pre-commit will load both configuration formats but issue warnings for the legacy format.
- After the migration period, the old configuration format will be purged.

The "officially" maintained repositories ([pre-commit/pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) as well as the [mirrors-*](https://github.com/pre-commit) repositories) will force-push tags for the lastest $n versions and provide both configuration formats for those tags.  Going forward, [pre-commit/pre-commit-mirror-maker](pre-commit/pre-commit-mirror-maker) will only produce the new configuration format.

[...]

PR #415, opened 2016-09-28
Moving from aspy.yaml to ruamel.yaml in order to preserve comments when using autoupdate - fixes #210

### LINKED ISSUE #210: autoupdate is awesome except it gobbles comments and reformats lists :/

Lucas-C:


asottile:
So I already tried this same approach with #211 and hit a number of issues.  I have all but written off ruamel.yaml as a non solution as it was essentially trading some bugs for a bunch of other bugs.

That said,  I'd love to hear your comments on #414 where I plan to do away with yaml entirely!

Lucas-C:
Damn. I added a comment to #414
Is #211 really blocked ?

asottile:
'fraid so :(

asottile:
We'll continue to explore options in #414.

PR #423, opened 2016-11-01
Added a failure message

brewneaux:


asottile:
I'm - 1 for a couple reasons:

- logic should be in the python code 
- the messaging is already pretty clear that something failed from the hook output

If you were going to approach this, you'd need to adjust the  identifying value at the top of the script and add a test.

brewneaux:
My bad yall, i meant to do this to our companies version, but I screwed it up.  Closing.

PR #434, opened 2016-11-26
Reorganize output writing

asottile:
This eliminates calls to print() and flush() and writes some wrappers.

The key difficulty was the tangled nature of the logging handler, the run method, and passing a `write` function around.

Instead, there's now a `cap_out` test fixture and all places call into `output.write` or `output.write_line` directly.


pre_commit/commands/run.py
```diff
@@ -104,26 +104,25 @@ def _run_single_hook(hook, repo, args, write, skips, cols):
         print_color = color.GREEN
         pass_fail = 'Passed'
 
-    write(color.format_color(pass_fail, print_color, args.color) + '\n')
+    output.write_line(color.format_color(pass_fail, print_color, args.color))
 
     if (stdout or stderr or file_modifications) and (retcode or args.verbose):
-        write('hookid: {}\n'.format(hook['id']))
-        write('\n')
+        output.write_line('hookid: {}\n'.format(hook['id']))
```
chriskuehl: intentionally losing a newline here?

chriskuehl: jk

asottile: write_line adds a newline

chriskuehl:
looks okay besides the tests

asottile:
Unclear why windows isn't passing, I imagine setuptools / pip broke something.

pypy3 I'm  just going to turn off for now

asottile:
Seems this is the windows problem: https://github.com/pypa/setuptools/issues/861

asottile:
893e0a9 is a sneaky workaround for the windows setuptools issues

chriskuehl:
your dedication to windows users is admirable

PR #441, opened 2016-12-02
Fix test error "fatal: empty ident name (for <(null)>) not allowed"

alzeih:
Occurs when tests run with tox not with Travis CI or Appveyor

Changed existing tox setenv statement to use whitespace around `=` as per
http://tox.readthedocs.io/en/latest/example/basic.html#setting-environment-variables

Related to https://github.com/pre-commit/pre-commit-hooks/pull/154 and https://github.com/pre-commit/pre-commit-hooks/pull/155

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,7 @@ env: # These should match the tox env list
     - TOXENV=py27 LATEST_GIT=1
 install: pip install coveralls tox
 script: tox
-# Special snowflake.  Our tests depend on making real commits.
 before_install:
-    - git config --global user.name "Travis CI"
-    - git config --global user.email "user@example.com"
     # Our tests inspect some of *our* git history
     - git fetch --unshallow
     - git --version
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,8 +13,6 @@ install:
 build: false
 
 before_test:
-  - git config --global user.name "AppVeyor CI"
-  - git config --global user.email "user@example.com"
   # Shut up CRLF messages
   - git config --global core.safecrlf false
 
--- a/tests/commands/install_uninstall_test.py
+++ b/tests/commands/install_uninstall_test.py
@@ -223,16 +223,17 @@ def test_environment_not_sourced(tempdir_factory):
 
         # Use a specific homedir to ignore --user installs
         homedir = tempdir_factory.get()
-        # Need this so we can call git commit without sploding
-        with io.open(os.path.join(homedir, '.gitconfig'), 'w') as gitconfig:
-            gitconfig.write(
-                '[user]\n'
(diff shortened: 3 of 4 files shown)


asottile:
This should get that last test passing:

```diff
diff --git a/tests/commands/install_uninstall_test.py b/tests/commands/install_uninstall_test.py
index 4c4baad..2547d99 100644
--- a/tests/commands/install_uninstall_test.py
+++ b/tests/commands/install_uninstall_test.py
@@ -147,7 +147,8 @@ FILES_CHANGED = (
 NORMAL_PRE_COMMIT_RUN = re.compile(
     r'^\[INFO\] Initializing environment for .+\.\r?\n'
     r'Bash hook\.+Passed\r?\n'
-    r'\[master [a-f0-9]{7}\] Commit!\r?\n' +
+    r'\[master [a-f0-9]{7}\] Commit!\r?\n'
+    r' Author: test <test@example.com>\r?\n' +
     FILES_CHANGED +
     r' create mode 100644 foo\r?\n$'
 )
```

alzeih:
Thanks for the patch, it guided me in the right direction.

I discovered this extra output by git of the Author is because one (or both) GIT_*_EMAIL environment variables isn't set (which I had not due to a typo). Rebased with the correct names for the environment variables. 

Also changed to using os.environ.get('') instead of os.environ[''] which matches the convention of the rest of the code.

asottile:
If you're using get() you should provide a fallback (or not use get).  The rest use get because they're only sometimes present (windows). 

Can you give them constant values? That'll make it work better (at least for me) when I run outside of town.

alzeih:
Makes sense, os.environ.get() doesn't fail fast. Seems silly to redefine the values twice though. Would os.environ[''] be better as it will fail with a KeyError ?

Since tox also sets VIRTUALENV_NO_DOWNLOAD, I think that pytest should be responsible for ensuring the environment is set if it's expected to be there when running tests without tox. Or the tests that modify env could be written to not throw away the entire environment and only throw out or set explicitly the environment that will cause problems?

asottile:
Let's use brackets so it fails fast,  I think this plugin might help with the environment variable problem: https://github.com/MobileDynasty/pytest-env

Then it'd be set in any pytest invocation.  Add pytest-env to requirements-dev.txt and a config section to tox.ini and it should work?  I haven't used the plugin before but I imagine it's that easy?

alzeih:
Rebased with brackets. 

I haven't used pytest-env either. Suggest this be looked at in another PR as the issue already exists with VIRTUALENV_NO_DOWNLOAD.

asottile:
Sounds good :ship: :it:

asottile:
Thanks!

PR #442, opened 2016-12-02
Update the archive resources (rbenv, ruby-build, ruby-download)

alzeih:
Just testing whether Travis and Appveyor will work with these versions.

My guess is that the version needed may be dependent on platform as for which rbenv version is most stable and compiles or has a binary build available. 

In my case, 1.9.3p547 is listed as 1.9.3-p547 and can't be found, and 2.1.5 errors as it doesn't have a binary build for my system. So these numbers chosen are purely "works for me".

If these latest versions don't work for the CI, then the next step is possibly to make the ruby version used by default a working value for the appropriate system by some mechanism.


asottile:
So the current version chosen is to make it fast on precise (travis-ci) and xenial (where I develop).  You did correctly identify that the archives needed updating to get the fallback behaviour.

ruby has two paths for installation, ruby-download (which uses https://rvm.io/binaries) and ruby-build (the slow build-from-source route using rbenv-build)

From this PR I think the second commit is good but probably not the first one.  If there's a difference between 1.9.3-p547 and 1.9.3p547, perhaps we should be doing some normalization.  Originally these resolved the same but perhaps they do not now :)

asottile:
You'll notice that, though they succeeded, this PR nearly doubles the travis-ci runtime:

- before: https://travis-ci.org/pre-commit/pre-commit/builds/179137636
- after: https://travis-ci.org/pre-commit/pre-commit/builds/180671183

alzeih:
Thanks, I didn't understand why that particular ruby version was used for the tests. 

I have rebased to only include the archive tarballs, and renamed the PR to reflect what it is now.

asottile:
Thanks!

Issue #432, opened 2016-11-24
Question: support for multiple configurations?


snakescott:
I'm interested in experimenting with different sets of hooks, having different hooks run on my build system versus locally, etc. 

Straightforward ways this might be possible:
- ``` pre-commit run --yaml $PATH --all-files```
- ```PRE_COMMIT_YAML_PATH=$PATH git commit -m "example"```

A more advanced option would be something like adding a ```groups``` yaml key as a sibling of  ```id``` e.g.

```yaml
-   repo: git://github.com/pre-commit/pre-commit-hooks
    sha: v0.4.2
    hooks:
    -   id: trailing-whitespace
        groups: [basic, advanced]
    -   id: flake8
        groups: [advanced]
```

You could then have ```pre-commit run --groups ...``` and perhaps ```pre-commit set-commit-groups ...```

I don't see anything like this discussed in the docs. Are there advanced/undocumented features that might support this? Is this functionality inline with pre-commit's project goals, or too complicated/promotes a workflow the project doesn't agree with?

asottile:
So I can think of two ways to accomplish this in the current toolset, they're not the greatest but maybe they get you going in the right direction?

The `SKIP` environment variable allows you to temporarily disable hooks, so if I didn't want to run `autopep8` or `flake8` on an invocation I could run `SKIP=flake8,autopep8 pre-commit`. http://pre-commit.com/#temporarily-disabling-hooks

pre-commit can run individual hooks at a time by passing them on the commandline: `pre-commit run trailing-whitespace`.  You could do something where you maybe `echo 'trailing-whitespace flake8' | xargs -n1 pre-commit run`? (briefly mentioned here: http://pre-commit.com/#usage)

A tangentially related feature: you can already configure hooks to run at different stages (currently only pre-commit / pre-push) via http://pre-commit.com/#confining-hooks-to-run-at-certain-stages

As for implementing I'm a little concerned for a couple reasons:
- you can _kinda_ do it without adding additional tooling (yeah it's not the best, but it is possible)
- I'm not sure if the feature will get much use via the main usage of the tool: as a pre-commit hook (I'm not even sure how it would work with that, perhaps writing the configuration to  the .git/hooks/pre-commit executable, but then it's not idempotent any more and that complicates as well).
- at least for our usage we're running every hook every time on every commit

snakescott:
Makes sense!

[...]

PR #443, opened 2016-12-02
Add option to run from alternate config file

snakescott:
Basics for #432 

Was hoping to get a smell test on this before I go any further. I'm thinking I'd rather not mess with install/uninstall, but I do plan to review and improve tests.


snakescott:
Ah nuts forgot some stuff herp derp

PR #444, opened 2016-12-02
Add option to run from alternate config file

snakescott:
This PR is to add support for alternate config files as discussed in #432. I'm intentionally adding functionality  to the run and autoupdate commands and not install or uninstall as I imagine .pre-commit-config.yaml will always exist.

I think this may be all that's necessary for --config to work? If this approach looks good, I plan to add unit tests as well as run some manual smoke tests myself. Thought it would be prudent to sanity check these changes first, though.

Thanks!

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -34,6 +34,10 @@ def _add_color_option(parser):
     )
 
 
+def _add_config_option(parser):
+    parser.add_argument('-c', '--config', help='Path to alternate config file')
+
+
 def main(argv=None):
     argv = argv if argv is not None else sys.argv[1:]
     argv = [five.to_text(arg) for arg in argv]
@@ -89,6 +93,7 @@ def main(argv=None):
         help="Auto-update pre-commit config to the latest repos' versions.",
     )
     _add_color_option(autoupdate_parser)
+    _add_config_option(autoupdate_parser)
 
     run_parser = subparsers.add_parser('run', help='Run hooks.')
     _add_color_option(run_parser)
@@ -119,6 +124,7 @@ def main(argv=None):
         '--hook-stage', choices=('commit', 'push'), default='commit',
         help='The stage during which the hook is fired e.g. commit or push.',
     )
+    _add_config_option(run_parser)
     run_mutex_group = run_parser.add_mutually_exclusive_group(required=False)
     run_mutex_group.add_argument(
         '--all-files', '-a', action='store_true', default=False,
@@ -152,7 +158,10 @@ def main(argv=None):
 
     with error_handler():
         add_logging_handler(args.color)
-        runner = Runner.create()
+        runner_kwargs = {}
+        if hasattr(args, 'config_file'):
+            runner_kwargs['config_file'] = args.config_file
+        runner = Runner.create(**runner_kwargs)
         git.check_for_cygwin_mismatch()
 
         if args.command == 'install':
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -35,7 +35,10 @@ def _add_color_option(parser):
 
 
 def _add_config_option(parser):
-    parser.add_argument('-c', '--config', help='Path to alternate config file')
+    parser.add_argument(
+        '-c', '--config', default='.pre-commit-config.yaml',
+        help='Path to alternate config file'
+    )
 
 
 def main(argv=None):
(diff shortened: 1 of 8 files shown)


pre_commit/main.py
```diff
@@ -152,7 +158,10 @@ def main(argv=None):
 
     with error_handler():
         add_logging_handler(args.color)
-        runner = Runner.create()
+        runner_kwargs = {}
+        if hasattr(args, 'config_file'):
+            runner_kwargs['config_file'] = args.config_file
```
asottile: imo just add this to all the commands (even if they don't use it) (I did this for `--color` recently and it made these sorts of things simpler)

pre_commit/main.py
```diff
@@ -34,6 +34,10 @@ def _add_color_option(parser):
     )
 
 
+def _add_config_option(parser):
+    parser.add_argument('-c', '--config', help='Path to alternate config file')
```
asottile: I'd do `default='.pre-commit-config.yaml'`

pre_commit/runner.py
```diff
@@ -16,26 +16,27 @@ class Runner(object):
...
     """
 
-    def __init__(self, git_root):
+    def __init__(self, git_root, config_file=None):
         self.git_root = git_root
+        self.config_file = config_file or C.CONFIG_FILE
 
     @classmethod
-    def create(cls):
+    def create(cls, config_file=None):
```
asottile: I'd make this a required argument (in both places), I know this is going to require a bunch of test changes but it'll make the runtime much simpler and less error-prone probably?

pre_commit/main.py
snakescott: will add

pre_commit/main.py
snakescott: okay!

pre_commit/runner.py
snakescott: will do!

snakescott:
I will update the PR with requested changes as well as tests.

snakescott:
Ah, sorry -- for install and uninstall, I'm not sure if it is *safe* to pass an alternate config file?

snakescott:
Are there any "public API" considerations for Runner? I initially had the arg as optional to maintain compatibility.

asottile:
> Ah, sorry -- for install and uninstall, I'm not sure if it is safe to pass an alternate config file?

yeah it just wouldn't do anything (except for `--install-hooks`), don't bother passing the `--config` state along to the written file

> Are there any "public API" considerations for Runner? I initially had the arg as optional to maintain compatibility.

Nope! the only "public" apis are cli

snakescott:
I find it cleaner to build appropriate args for `Runner` than to have `--config` available as an option for `install` when it is actually has no effect. That said, I'm happy to defer your preference, so I'll `_add_config_option` to all the commands in my next PR.

snakescott:
@asottile would new tests in `runner_tests.py` be sufficient or do you think I should add e2e tests in other places?

asottile:
I'd probably do one test in runner, and maybe one end-to-end test in tests/commands/run_test.py just to demonstrate it definitely works

snakescott:
Great, I will add another commit with those tests later today!

pre_commit/main.py
```diff
@@ -34,11 +34,19 @@ def _add_color_option(parser):
...
+        help='Path to alternate config file'
+    )
+
+
 def main(argv=None):
     argv = argv if argv is not None else sys.argv[1:]
     argv = [five.to_text(arg) for arg in argv]
     parser = argparse.ArgumentParser()
 
+    parser.set_defaults(config='.pre-commit-config.yaml')
```
asottile: Is this line necessary? (I don't think so but I'm not sure?)

snakescott: I think without this, help gets sad? I definitely added it in response to issues I hit.

Generally, if Runner is going to be created no matter what the command is, and we are going to pass config in, it feels right to set a global default?

snakescott: I can try to reconstruct the errors I hit if that would be helpful.

asottile: I'm pretty sure Runner doesn't get constructed in the help case (otherwise my color stuff would be broken)

snakescott:
passed unit tests without that line, so removed!

snakescott:
Sweet! I'll keep my eye out for a release incorporating these changes so I can switch from pointing to an internal prerelease to mainline.

Thanks for your help on this PR and for answering all of my other questions :)

PR #448, opened 2016-12-04
Fail if grep cannot be executed. Resolves #447.

### LINKED ISSUE #447: PCRE passes even if ggrep is not installed on OS X
I was scratching my head because some hooks were passing even though they should clearly have failed.

When running `run --all-files` with the `--verbose` flag, I noticed why this was happening:

````
[do_not_commit] Block if "DO NOT COMMIT" is found..................................................................Passed
hookid: do_not_commit

Executable `ggrep` not found
````

It seems like the correct behaviour would be to fail and warn user to install `ggrep`?

For the record, I installed `ggrep` as follow: `brew install homebrew/dupes/grep`, and the hooks now fail as expected.

cmer:


asottile:
I've made #449 which fixes the root issue (`pre_commit.xargs` + `negate=True` would treat executable missing as a "success")

As for this PR:
- hooks weren't run
- `os.system` is frowned upon
- `which` isn't portable (we support windows)

Thanks for the issue and PR though!

Issue #209, opened 2015-03-07
Add support for PHP hooks


struys:
hootsuite is already using pre-commit with PHP hooks but they're using PHP from the system level (https://github.com/hootsuite/pre-commit-php). We should add support for PHP in pre-commit.

struys:
We need a system similar to pip for pear packages or whatever is most common in PHP. I don't know enough about it to say how we should be doing it.

It should look something like what we have for python/node/ruby, for example - https://github.com/pre-commit/pre-commit/blob/master/pre_commit/languages/python.py

asottile:
I looked a bit at these two projects: https://github.com/phpenv/phpenv and https://github.com/CHH/phpenv

They both seem kind of deadish unfortunately (the first has a completely dead `dev` branch, and a partially (?) working master branch).   They both seem like I could use a similar approach to our ruby support to get a virtualenv-like thing working

struys:
@shihoonhootsuite have any experience with these?

asottile:
This seems more promising: https://github.com/virtphp/virtphp

shihoonhootsuite:
Thanks for this feature suggestion! Sorry I haven't had much experience with running these local php envs - it has been still a relative new thing in PHP. But the last link from @asottile at https://github.com/virtphp/virtphp does look promising. I assume the aim of doing this is so that users can write hooks in PHP right?

asottile:
Indeed, that's the goal :)

awebdeveloper:
would  composer help

https://getcomposer.org/

asottile:
This was initially looked into in #458 but it was decided it was too fragile and difficult to support properly.

The suggested route for php-based hooks is to use the new (in 0.10.0 docker support).  http://pre-commit.com/#supported-languages talks about how to approach implementing hooks using docker.

PR #458, opened 2017-01-03
Support php for precommit

bpicolo:
This is a first pass at #209 

No tests as this sort of needs opinions first.

PHP is a tricky language to support.

Here's some assumptions I made:
1. People aren't likely to have virtphp installed. It doesn't have any sort of ubiquity in the php ecosystem as far as I know (nor does any other php virtual environment)
2. Installing different PHP versions is all over the place. There's a big TODO as I'm not sure how to resolve this. virtphp just expects a php bin dir path to determine versions.
3. `composer` should be on the system path. There are a lot of ways people use composer, e.g. just putting composer.phar somewhere and running `php composer.phar <blah>`, but we sort of have to pick one (or also add a composer.phar to resources)
4. Php hooks should declare entrypoints to the vendor-bin, e.g. "vendor/bin/myexecutable". Didn't see how to have pre-commit check here by default, but maybe that's a thing?

Solutions:
1. Dep virtphp.phar in resources. We can tar.gz it if we want, 367k as-is
2. Not solved
3. Chose "require composer on system path" as the option.

A `phar` is just basically just a bunch of php files concatted together to make up a program. It includes `#!/usr/bin/env php` at the top, so can be run raw or via php (`composer.phar <foo>` or  `php composer.phar <foo>`). For running VirtPHP I chose the latter so it doesn't need to have executable permissions.

While composer is a sane package manager in most senses, there are external factors that make PHP support among packages difficult. In particular (something I hit developing this) is that there are a number of PHP packages dependent on php extensions which need to be compiled and loaded into the Zend runtime. Composer can not resolve these. PECL exists which only sort-of resolves this (it's not great at it's job, just a standalone installer, and also doesn't support the extension I cared about for etsy/phan anyway). There's no configuration format for PECL that I'm aware of, either, and often people will install extensions via e.g. apt instead. Eat24 compiles all of it's extensions from scratch in the docker containers that the app runs in

I have a (not-yet-oss) package that managed to install an extension into the virtphp env at install-time with composer and run properly, though having composer ignore those dependencies at the install time (`--ignore-platform-reqs`) was the only way to get that working. I gave @asottile  access to that repo to see how I'm dealing with that currently.

There's currently no windows support at all. virtphp is *nix only


pre_commit/languages/php.py
```diff
@@ -0,0 +1,130 @@
...
+VIRT_PHP_ENVIRONMENTS = os.path.join(
+    os.path.expanduser('~'),
+    '.virtphp',
+    'environments.json'
+)
+
+
+def get_env_patch(venv, language_version):
+    patches = (
+        ('VIRTPHP_COMPOSER_GLOBAL', '1'),
```
bpicolo: tbh I have no idea if this is necessary. I mostly picked bits out of the virtphp bin/activate script that

pre_commit/languages/php.py
```diff
@@ -0,0 +1,130 @@
...
+    except BaseException:
+        if os.path.exists(VIRT_PHP_ENVIRONMENTS):
+            with open(VIRT_PHP_ENVIRONMENTS, 'r+') as f:
+                environments = json.loads(f.read())
+                environments = {
+                    environment_key: environment_config
+                    for environment_key, environment_config
+                    in environments.items() if environment_key != directory
+                }
+                f.seek(0)
```
bpicolo: rewriting the virtphp config is sort of bizarre but there is no option available with virtphp that magically fixes this.

pre_commit/languages/php.py
```diff
@@ -0,0 +1,131 @@
...
+
+    with cleanup_environment_on_failure(repo_cmd_runner, directory):
+        virtphp_env_command = [
+            'php', VIRT_PHP_PATH, 'create',
+            '--install-path', repo_cmd_runner.prefix_dir[:-1],
+            directory
+        ]
+
+        if version != 'default':
+            # TODO: Not sure what the best choice is here. There's doesn't seem
```
bpicolo: I haven't put version handling in other relevant parts of this file yet, in favor of figuring out something sane first

asottile:
At a first pass, this seems pretty promising!  I'm a little surprised that virtphp seems to require you to keep bookkeeping inside the homedir, I wonder if they would be open to a PR which disables that (imo) strange behaviour.

Given how strange the php ecosystem is, I'm ok with our wrapper having to do some strange things to enable it.  Sometimes ya just gotta hack hack hack

bpicolo:
Closing this in favor of docker

PR #459, opened 2017-01-03
Docker hook support for pre-commit

bpicolo:
This adds support for docker-based hooks.

Current expectations:
1. The hook should not be sad that we overwrite everything in /src by the cwd into it as a volume
2. entry in your hooks config is the cmd entrypoint inside the container

@asottile  see that same repo I sent you as an example.

Happy to add tests if we agree that this is sane

v2 might want this to clean up after itself.

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -16,7 +16,7 @@ PRE_COMMIT_LABEL = 'PRE_COMMIT'
 
 
 def md5(s):
-    return five.to_text(hashlib.md5(s).hexdigest())
+    return hashlib.md5(five.to_bytes(s)).hexdigest()
 
 
 def docker_tag(repo_cmd_runner):
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -1,0 +1,62 @@
+from __future__ import unicode_literals
+
+import hashlib
+import os
+
+from pre_commit.languages import helpers
+from pre_commit.util import clean_path_on_failure
+from pre_commit.util import mkdirp
+from pre_commit.xargs import xargs
+
+
+ENVIRONMENT_DIR = 'docker'
+
+
+def md5(s):
+    m = hashlib.md5()
+    m.update(s)
+    return m.hexdigest()
+
+
+def docker_tag(repo_cmd_runner):
+    return 'pre-commit-{}'.format(
+        md5(os.path.basename(repo_cmd_runner.path()))
+    ).lower()
+
+
+def install_environment(
+        repo_cmd_runner,
+        version='default',
+        additional_dependencies=(),
+):
+    assert repo_cmd_runner.exists('Dockerfile')
+    # I don't know of anybody trying to juggle multiple docker installations
+    # so this seems sufficient
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, 'default')
+    mkdirp(os.path.join(repo_cmd_runner.path(), directory))
+
+    cmd = (
+        'docker', 'build', '--pull',
+        '--tag', docker_tag(repo_cmd_runner),
+        '.'
+    )
+
+    # Docker doesn't really have relevant disk environment, but pre-commit
+    # still needs to cleanup it's state files on failure
+    env_dir = repo_cmd_runner.path(directory)
+    with clean_path_on_failure(env_dir):
+        helpers.run_setup_cmd(repo_cmd_runner, cmd)
+
+
+def run_hook(repo_cmd_runner, hook, file_args):
+    cmd = (
+        'docker', 'run',
+        '-t',
+        '-v', '{}:/src'.format(os.getcwd()),
+        '--workdir', '/src',
+        docker_tag(repo_cmd_runner)
+    )
+
+    return xargs(
+        cmd + (hook['entry'],) + tuple(hook['args']), file_args,
+    )
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -6,8 +6,9 @@ import os
 
 from pre_commit import five
 from pre_commit.languages import helpers
+from pre_commit.util import CalledProcessError
 from pre_commit.util import clean_path_on_failure
-from pre_commit.util import mkdirp
+from pre_commit.util import cmd_output
 from pre_commit.xargs import xargs
 
 
@@ -26,7 +27,10 @@ def docker_tag(repo_cmd_runner):
 
 
 def docker_is_running():
-    return xargs(('docker',), ['ps'])[0] == 0
+    try:
+        return cmd_output('docker', 'ps')[0] == 0
+    except CalledProcessError:
+        return False
 
 
 def assert_docker_available():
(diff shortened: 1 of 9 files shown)


pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+    # still needs to cleanup it's state files on failure
+    env_dir = repo_cmd_runner.path(directory)
+    with clean_path_on_failure(env_dir):
+        helpers.run_setup_cmd(repo_cmd_runner, cmd)
+
+
+def run_hook(repo_cmd_runner, hook, file_args):
+    cmd = (
+        'docker', 'run',
+        '-t',
```
chriskuehl: probably without -t, this will fail on e.g. jenkins

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+        '-v', '{}:/src'.format(os.getcwd()),
+        '--workdir', '/src',
```
chriskuehl: run with --rm?

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+    env_dir = repo_cmd_runner.path(directory)
+    with clean_path_on_failure(env_dir):
+        helpers.run_setup_cmd(repo_cmd_runner, cmd)
+
+
+def run_hook(repo_cmd_runner, hook, file_args):
+    cmd = (
+        'docker', 'run',
+        '-t',
+        '-v', '{}:/src'.format(os.getcwd()),
```
bpicolo: we'll probably want this as :rw but we'll need docker to not write things as root.

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+from pre_commit.xargs import xargs
+
+
+ENVIRONMENT_DIR = 'docker'
+
+
+def md5(s):
+    m = hashlib.md5()
+    m.update(s)
+    return m.hexdigest()
```
chriskuehl: you can just

```python
hashlib.md5('my string').hexdigest()
```

bpicolo: :( oh

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+from pre_commit.util import mkdirp
+from pre_commit.xargs import xargs
+
+
+ENVIRONMENT_DIR = 'docker'
+
+
+def md5(s):
+    m = hashlib.md5()
+    m.update(s)
```
asottile: in python3 I think this needs `s.encode('UTF-8')`

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+    ).lower()
+
+
+def install_environment(
+        repo_cmd_runner,
+        version='default',
+        additional_dependencies=(),
+):
+    assert repo_cmd_runner.exists('Dockerfile')
+    # I don't know of anybody trying to juggle multiple docker installations
```
asottile: I would `assert version == 'default'` with an explanation that we're not going to support multiple docker versions

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Docker doesn't really have relevant disk environment, but pre-commit
+    # still needs to cleanup it's state files on failure
+    env_dir = repo_cmd_runner.path(directory)
+    with clean_path_on_failure(env_dir):
+        helpers.run_setup_cmd(repo_cmd_runner, cmd)
+
+
+def run_hook(repo_cmd_runner, hook, file_args):
+    cmd = (
```
asottile: we'll probably want to make sure the image exists at run time, since we can't guarantee the image exists just because of the filesystem directory.  Especially given how aggressively we run docker cleanup at yelp

bpicolo: And just rebuild it if it doesn't?

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def docker_tag(repo_cmd_runner):
+    return 'pre-commit-{}'.format(
+        md5(os.path.basename(repo_cmd_runner.path()))
+    ).lower()
+
+
+def docker_is_running():
+    return xargs(('docker',), ['ps'])[0] == 0
```
asottile: heh, I think you want `cmd_output('docker', 'ps')[0] == 0`

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,88 @@
...
+    assert repo_cmd_runner.exists('Dockerfile'), (
+        'No Dockerfile was found in the hook repository'
+    )
+    assert version == 'default', (
+        'Pre-commit does not support language_version for docker '
+    )
+    assert_docker_available()
+
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, 'default')
+    mkdirp(os.path.join(repo_cmd_runner.path(), directory))
```
asottile: `os.mkdir(repo_cmd_runner.path(directory))`

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,88 @@
...
+    with clean_path_on_failure(env_dir):
+        build_docker_image(repo_cmd_runner)
+
+
+def run_hook(repo_cmd_runner, hook, file_args):
+    assert_docker_available()
+    # Rebuild the docker image in case it has gone missing, as many people do
+    # automated cleanup of docker images.
+    build_docker_image(repo_cmd_runner)
+    # the docker lib doesn't return stdout on non-zero exit codes,
```
asottile: I'd explicitly say "dockerpy" here so we know what's being talked about :)

testing/resources/docker_hooks_repo/hooks.yaml
```diff
@@ -0,0 +1,11 @@
+-   id: docker-hook
+    name: Docker test hook
+    entry: echo
+    language: docker
+    files: \.txt$
+
+-   id: docker-hook-failing
+    name: Docker test hook with nonzero exit code
+    entry: bork
```
asottile: hmmm, so this makes it more like existing things.  I wonder if we should leave out `entry` with docker hooks or at least make it optional?  Wondering what you think about this :)

bpicolo: Well, it certainly seems useful to base multiple hook entrypoints off the same container.

Happy to make `entry` optional, we'll need to document that people have to define container entrypoints though?

pre_commit/languages/docker.py
bpicolo: I can just remove that comment since there's no dockerpy in here anymore

testing/resources/docker_hooks_repo/hooks.yaml
asottile: actually nah, that's a bunch of complexity.  We can always change it later if people think otherwise.

pre_commit/languages/docker.py
```diff
@@ -0,0 +1,92 @@
...
+    ).lower()
+
+
+def docker_is_running():
+    try:
+        parse_shebang.normexe('docker')
+    except parse_shebang.ExecutableNotFoundError:
+        return False
+
+    return cmd_output('docker', 'ps')[0] == 0
```
asottile: could even move this to line 31

bpicolo: Already have a diff that does just this hah

asottile:
*AWESOME* :tada: 

I'm going to play with this a bit and see how it goes and then make a release

PR #462, opened 2017-01-05
Support docker hooks with args

bpicolo:
@asottile  here's a fix for entry args. lmk if there's some corner case that makes .split() not satisfactory

Also, seems optimal to not pass up the user. Couldn't get writes working on my localhost without. Docker will only set permissions to root if it's creating *new* files so... /shrug not sure! People could always pass it as an args? (Or maybe not if we don't expand env vars...). This has sort of always been the most annoying part of docker heh

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -89,6 +89,7 @@ def run_hook(repo_cmd_runner, hook, file_args):
     cmd = (
         'docker', 'run',
         '--rm',
+        '-u', '{}:{}'.format(os.getuid(), os.getgid()),
         '-v', '{}:/src:rw'.format(os.getcwd()),
         '--workdir', '/src',
         '--entrypoint', entry_executable,
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import hashlib
 import os
+import shlex
 
 from pre_commit import five
 from pre_commit.languages import helpers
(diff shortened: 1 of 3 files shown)


pre_commit/languages/docker.py
```diff
@@ -81,14 +81,17 @@ def run_hook(repo_cmd_runner, hook, file_args):
     # Rebuild the docker image in case it has gone missing, as many people do
     # automated cleanup of docker images.
     build_docker_image(repo_cmd_runner, pull=False)
+
+    entry_parts = hook['entry'].split(' ')
```
asottile: this should use shlex.split (like the system hooks)

testing/resources/docker_hooks_repo/hooks.yaml
```diff
@@ -4,6 +4,12 @@
     language: docker
     files: \.txt$
 
+-   id: docker-hook-arg
+    name: Docker test hook
+    entry: echo -n
```
asottile: echo -e is probably better (and then include a `r'\n'` and make sure it goes through.  echo -n is kinda difficult to determine in the tests (since it just does `in` checks)

bpicolo: I think the docker container in question only implements -n, and all that does is remove newline if present

asottile:
Hmmm I'm not sure about not passing `--user`, what sort of messages are you getting?  I'd really hope we wouldn't run as root :/

bpicolo:
Permission denied on writes =\

asottile:
Can you show some output? could it be due to writing other directories? it shouldn't have permission denied writing to the user mounted files

bpicolo:
@asottile  looks like it's a boot2docker issue. I guess we can document that boot2docker can't do writes?

Issue #451, opened 2016-12-06
Support for SwiftPM packages


benasher44:
[SwiftLint](/Realm/SwiftLint) is a great example of a Swift-based linter that has support for [SwiftPM](/apple/swift-package-manager). It'd be great if pre-commit had support to pull down a repo that has a Package.swift, run `swift build`, and install the specified targets to be used as hooks.

asottile:
This would be great, should be as easy as implementing one of these for swift: https://github.com/pre-commit/pre-commit/blob/8837cfa7ffcc419216d4e01392cee0f1ceee9c88/pre_commit/languages/python.py

The interface is here: https://github.com/pre-commit/pre-commit/blob/8837cfa7ffcc419216d4e01392cee0f1ceee9c88/pre_commit/languages/all.py

Want to take a first stab at it?

benasher44:
Maybe! I was hoping to just get it filed and labeled for now. I'm pretty swamped atm, so anyone else is welcome to pick it up.

benasher44:
@asottile I've started work on this. I have SwiftPM packages building locally, but there's more work to be done to fully test with SwiftLint before I get a PR up. I'll keep ya posted!

asottile:
Via #467 -- available in 0.11.0

PR #467, opened 2017-01-13
SwiftPM support

benasher44:
This implements #451.

All that's needed to build a SwiftPM package is to run`swift build`. When doing this, you choose either a `debug` or `release` type build, and then build artifacts/binary is stored in <build_dir>/<release|debug>.  The `-C` flag is just used to ensure `swift build` is run with the repo as the current directory, which apparently wasn't the case when I tested the first time.


benasher44:
I still have to add a `hooks.yaml` to SwiftLint, as well as finish implementing realm/SwiftLint#810, but realm/SwiftLint#810 should be enough, since you could get away with specifying the `hooks.yaml` details locally in your `.pre-commit-config.yaml`.

pre_commit/languages/swift.py
```diff
@@ -0,0 +1,40 @@
...
+            '--build-path', os.path.join(directory, BUILD_DIR),
+        ))
+
+
+def run_hook(repo_cmd_runner, hook, file_args):
+    directory = repo_cmd_runner.path(helpers.environment_dir(
+        ENVIRONMENT_DIR, 'default',
+    ))
+    cmd = helpers.to_cmd(hook)
+    full_binary_path = os.path.join(directory, BUILD_DIR, BUILD_CONFIG, cmd[0])
```
asottile: instead of doing this, let's do a `PATH` envcontext (see any of the python / ruby / js ones for an example).  Then you can just use `helpers.to_cmd` directly

pre_commit/languages/swift.py
```diff
@@ -0,0 +1,40 @@
...
+BUILD_CONFIG = 'release'
+
+
+def install_environment(
+        repo_cmd_runner,
+        version='default',
+        additional_dependencies=(),
+):
+    directory = repo_cmd_runner.path(helpers.environment_dir(
+        ENVIRONMENT_DIR, 'default',
```
asottile: I would assert that the version is default since no other version is supported, there's prior art in the docker language

asottile:
I added some commits to your branch :)

benasher44:
\o/

asottile:
Thanks and wow!

benasher44:
🎉🎉🎉 thank you!

asottile:
I'll cut a release for this tomorrow

asottile:
Damn I'm a liar, doing this now <3

asottile:
This is in 0.11.0, thanks again for an awesome contribution.  Want to write up a bit in the documentation and add your name to the contributors?  https://github.com/pre-commit/pre-commit.github.io (fork and base a branch off of `real_master`)

benasher44:
Sure! I'll do it soon. I wanna wait for my swiftlint PRs to ship, so that I can have an example for the docs.

asottile:
👍

PR #468, opened 2017-01-21
Test py36 instead of py34

asottile:


asottile:
Note that py34+ is still known to work and is supported -- it's just faster to test fewer versions in travis.

PR #473, opened 2017-01-24
Add first class support for golang hooks

asottile:


pre_commit/git.py
```diff
@@ -32,6 +32,11 @@ def get_git_dir(git_root):
     ))
 
 
+def get_remote_url(git_root):
```
chriskuehl: nice pun

pre_commit/languages/golang.py
```diff
@@ -0,0 +1,72 @@
...
+    with clean_path_on_failure(directory):
+        remote = git.get_remote_url(repo_cmd_runner.path())
+        repo_src_dir = os.path.join(directory, 'src', guess_go_dir(remote))
+
+        # Clone into the goenv we'll create
+        helpers.run_setup_cmd(
+            repo_cmd_runner, ('git', 'clone', '.', repo_src_dir),
+        )
+
+        env = dict(os.environ, GOPATH=directory)
```
Lucas-C: Sadly this does not work with Cygwin:
```
An unexpected error has occurred: CalledProcessError: Command: ('/g/devhome/tools/Go/bin/go', 'get', './...')
Return code: 2
Expected return code: 0
Output: (none)
Errors:
    go: GOPATH entry is relative; must be absolute path: "/home/lucas_cimon/.pre-commit/repohk_ul6_a/golangenv-default".
    Run 'go help gopath' for usage.
```
I double checked : `GOPATH` must be a Windows path under Cygwin with a Go Windows install :(

asottile: That's unfortunate.  supporting a mixed mode is kind of strange usually you target all cygwin or all windows and not a mix.  It's certainly possible to make this work but I imagine it'll be very fragile :/

Lucas-C: Yes.
I used the following code in another project, but I'm not really proud of it:
```
if sys.platform == 'cygwin':
    filepath = subprocess.check_output(['cygpath', '-w', filepath]).strip().decode('utf8')
```

asottile: My concern would be that if there was suddenly a go package for cygwin this would be incorrect in the opposite direction.  If we could somehow narrow down this to an exceptional case I'd take a PR which added this

asottile: that might be difficult though ;)

Lucas-C: I give up for the moment : https://github.com/mrtazz/checkmake project structure is not standard. It requires to execute a Makefile instead of simply `go build`. Moreover its builtin rules are too lightweight to be useful in practice.

Issue #386, opened 2016-06-25
do not recommend `pre-commit autoupdate` on failure of `pre-commit autoupdate` 


matkoniecz:
It would be preferable to recommend something that has chance of fixing the problem

instruction to reproduce in #385

asottile:
So `autoupdate` is  useful to get you into a working state, however the directions should be:

- Remove the hook, run autoupdate, and add the hook back.

asottile:
See also: https://github.com/pre-commit/pre-commit/pull/368#issue-153987791

asottile:
This has been released in 0.9.0

PR #481, opened 2017-01-31
Improve messaging for missing hook given #480

asottile:
The approach that #480 takes to future-proof autoupdate a little more also improves the case for "hook not present yet".  We originally did some wordy work to suggest a way to upgrade for #386 in #387 but now that's not necessary \o/

CC @matkoniecz


matkoniecz:
Thanks!

PR #482, opened 2017-02-09
Keep dependency in order

tdeo:
I was investigating https://github.com/pre-commit/pre-commit-mirror-maker/issues/20 and realized running `gem install gem1:version1 gem2:version` can give different result from `gem install gem2:version2 gem1:version1`.

This PR keeps the list of additional dependencies of an hook in the order they appeared in the `.pre-commit-config.yaml`.

### DIFF
--- a/pre_commit/repository.py
+++ b/pre_commit/repository.py
@@ -64,7 +64,7 @@ class Repository(object):
 
     @cached_property
     def additional_dependencies(self):
-        dep_dict = defaultdict(lambda: defaultdict(set))
+        dep_dict = defaultdict(lambda: defaultdict(_UniqueList))
         for _, hook in self.hooks:
             dep_dict[hook['language']][hook['language_version']].update(
                 hook.get('additional_dependencies', []),
@@ -222,3 +222,14 @@ class LocalRepository(Repository):
     @cached_property
     def manifest(self):
         raise NotImplementedError
+
+
+class _UniqueList(list):
+    def __init__(self):
+        self._set = set()
+
+    def update(self, obj):
+        for item in obj:
+            if item not in self._set:
+                self._set.add(item)
+                self.append(item)
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -477,7 +477,20 @@ def test_additional_dependencies(tempdir_factory, store):
     config = make_config_from_repo(path)
     config['hooks'][0]['additional_dependencies'] = ['pep8']
     repo = Repository.create(config, store)
-    assert repo.additional_dependencies['python']['default'] == {'pep8'}
+    assert repo.additional_dependencies['python']['default'] == ['pep8']
+
+
+@pytest.mark.integration
+def test_additional_dependencies_duplicated(
+        tempdir_factory, store, log_warning_mock,
+):
+    path = make_repo(tempdir_factory, 'ruby_hooks_repo')
+    config = make_config_from_repo(path)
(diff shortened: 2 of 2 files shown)


pre_commit/repository.py
```diff
@@ -64,11 +64,10 @@ def languages(self):
...
     @cached_property
     def additional_dependencies(self):
-        dep_dict = defaultdict(lambda: defaultdict(set))
+        dep_dict = defaultdict(lambda: defaultdict(list))
         for _, hook in self.hooks:
-            dep_dict[hook['language']][hook['language_version']].update(
-                hook.get('additional_dependencies', []),
-            )
+            dep_dict[hook['language']][hook['language_version']] += \
+                hook.get('additional_dependencies', [])
```
asottile: So we do lose set semantics here, can you add a test with multiple hooks that have the same additional dependencies and verify that there are not duplicates?

tdeo: Well, the current `set` implementation is already accepting the same dependency listed at different versions. We could keep the same behaviour by handling duplicates manually:
```python
for dep in hook.get('additional_dependencies', []):
    if dep not in dep_dict[hook['language']][hook['language_version']]:
        dep_dict[hook['language']][hook['language_version']].append(dep)
```

An other option would be to extract the name and the version for each dependency (needs to be implemented for each language though) and then either:
  - keep only the first version of each package
  - issue a warning to the user
  - raise an exception

and then implement tests accordingly, what would you prefer?

asottile: The form is fine, I don't want to add language-specific parsing of this stuff if avoidable.

Heh, if we could only target python3.6 (or modern pypy), the set would have insertion order

pre_commit/repository.py
```diff
@@ -64,11 +64,18 @@ def languages(self):
...
+            for dep in hook.get('additional_dependencies', []):
+                if dep not in \
+                        dep_dict[hook['language']][hook['language_version']]:
+                    dep_dict[hook['language']][hook['language_version']].\
+                        append(dep)
+                else:
+                    logger.warning(
+                        'Additional dependency {} is listed twice for hook '
+                        '{}'.format(dep, hook['name'])
+                    )
```
asottile: A warning should not be raised here.

This is a perfectly valid configuration:

```yaml
    hooks:
    -   id: puppet-validate
        language_version: '2.1.5'
        additional_dependencies: ['puppet:4.5.3']
    -   id: epp-validate
        language_version: '2.1.5'
        additional_dependencies: ['puppet:4.5.3']
```

asottile: Maybe we can keep the code change smaller by making something like this:

```python
class _UniqueList(list):
    def __init__(self):
        self._set = set()

    def update(self, obj):
        for item in obj:
            if item not in self._set:
                self.append(item)
```

And then the diff is just

```diff
-        dep_dict = defaultdict(lambda: defaultdict(set))
+        dep_dict = defaultdict(lambda: defaultdict(_UniqueList))
```

tdeo: 👍 

Let me know if you want me to squash commits

asottile: Up to you!

tdeo:
I squashed the commit. Everything allright?

asottile:
Yep! I was waiting for the build and didn't notice it in my 50 tabs 🤣 

Thanks for the contribution!

tdeo:
Thanks for reactivity!

asottile:

Issue #200, opened 2015-02-27
Windows: Node Support


asottile:
This involves solving this ticket: https://github.com/ekalinin/nodeenv/issues/53

I've already started some work on this

asottile:
Track https://github.com/ekalinin/nodeenv/issues/140 for progress

alex-hutton:
Hi, I would appreciate more detail about what is broken and why it relates to `nodeenv`.

I found an apparent bug with node on Windows which may not be related to this bug. I also found a solution which I will explain below.

To summarise, the cmd given to __popen is the wrong cmd, eg the wrong exe is found by pre-commit.

I am using `pre-commit 0.12.2` and the `eslint` mirror (http://github.com/pre-commit/mirrors-eslint), `v3.9.1`.

I installed `eslint` itself through `npm` and it resides in my `node_modules` directory within my project folder. `npm` is `3.10.10` and `eslint` is `v3.15.0`.

The relevant block of my `.pre-commit-config.yaml` looks like this:
```
- repo: git://github.com/pre-commit/mirrors-eslint
  sha: 'v3.9.1'
  hooks:
  - id: eslint
    language: system
```

I invoked `eslint` like this from the windows command prompt (same thing happens from gitbash):
`pre-commit run --run-all eslint`

Note that I am in my python `virtualenv` and `node_modules/.bin` is in my PATH.

The result was a `FileNotFoundError`, raised from `util.cmd_outpuit()` at the line `proc = __popen(cmd, **popen_kwargs)`.

The value of `cmd` was `('\\bin\\sh', 'C:\\<path to project>\\node_modules\\.bin\\eslint', '<js file for linting>', '<another js file for linting')`.

In my `node_modules` directory, I actually have the following files: `eslint` and `eslint.cmd`. Looking at the sourcecode of `parse_shebang.find_execitable()` I could see that there was a candidate list of filenames that were attempted to match based on the executable file extensions of the windows environment. The first item in the list was `eslint` without any extension, so it always matched. 

In order to make `eslint.cmd` match inside `parse_shebang.find_execitable()`, I altered the code.

With the altered code, the value of `cmd` was like this: `('C:\\<path to project>\\node_modules\\.bin\\eslint', '<js file for linting>', '<another js file for linting>')`.

When subsequently invoked `eslint` as before worked in both the windows cmd prompt and gitbash.

I will submit a pull request for this shortly. I assume the new logic for matching files will not cause other problems.

[...]

PR #483, opened 2017-02-10
Fix eslint on windows

alex-hutton:
- The bare exe was the first filename attempted to match,
  this change means it will be matched last, allowing other files
  to be matched if they exist. The result is that eslint now works
  on Windows.

- Possibly related to #200


asottile:
Thanks for the PR! This'll be in the next version I release which should be in the next few days if I get enough time to do it.

asottile:

alex-hutton:
Thank you! It's a very useful piece of software.

PR #484, opened 2017-02-13
Add support for python and ruby local hooks

tdeo:
Hi,

Following up on last week's discussion about ruby-based hooks using gems in https://github.com/pre-commit/pre-commit-mirror-maker/issues/20, I tried to start hacking around to add support for ruby and python local hooks.

Let me know what you think about it.

### DIFF
--- a/pre_commit/clientlib/validate_config.py
+++ b/pre_commit/clientlib/validate_config.py
@@ -70,6 +70,7 @@ def validate_config_extra(config):
                 raise InvalidConfigError(
                     '"sha" property provided for local hooks'
                 )
+            repo['sha'] = 'local'
         elif 'sha' not in repo:
             raise InvalidConfigError(
                 'Missing "sha" field for repository {}'.format(repo['repo'])
--- a/pre_commit/clientlib/validate_config.py
+++ b/pre_commit/clientlib/validate_config.py
@@ -70,7 +70,6 @@ def validate_config_extra(config):
                 raise InvalidConfigError(
                     '"sha" property provided for local hooks'
                 )
-            repo['sha'] = 'local'
         elif 'sha' not in repo:
             raise InvalidConfigError(
                 'Missing "sha" field for repository {}'.format(repo['repo'])
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -30,7 +30,7 @@ def _update_repo(repo_config, runner, tags_only):
     Args:
         repo_config - A config for a repository
     """
-    repo = Repository.create(repo_config, runner.store)
+    repo = Repository.create(repo_config, runner.store, runner.git_root)
 
     with cwd(repo.repo_path_getter.repo_path):
         cmd_output('git', 'fetch')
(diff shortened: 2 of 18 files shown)


asottile:
so I've done some thinking about this and though it's kinda against the original pre-commit mentaility that we'd share hook configuration as much as possible, I think this might be an ok idea.

That said, it should do the following things that it isn't:
- each set of `additional_dependencies` should create a "repository" the same way that real repositories do.  This at least allows reuse instead of writing a bunch of virtualenvs to the current working directory.
- Since there's already a special class that handles `local` hooks this could be done pretty easily, here's the idea that I had for this:
- They would get a "special": `'local://{}/{}'.format(language, ','.join(sorted(deps)))` repository path (and probably a dummy value for `sha`?).  The framework would set up the per-language trivial repository (trivial setup.py, trivial gem, etc. maybe even all at the same time to make it simpler).
- The "clone" step of creating the dummy repo would be to just populate the repo inside ~/.pre-commit/repo...

tdeo:
I was altho thinking they should belong to `~/.pre-commit` somehow. 

However, this only raises the question of what to do with system / script local hooks. I think it's not great if we have to treat them differently them from let's say ruby local hooks. But I don't really see any alternative. But we cannot copy them to some `~/.pre-commit/repoXXXX` anyway.

So maybe, we should use the `LocalRepository` class only for those 2?

asottile:
We already treat system / script hooks specially (they don't have an environment directory)

testing/util.py
```diff
@@ -86,7 +87,7 @@ def cmd_output_mocked_pre_commit_home(*args, **kwargs):
 
 
 def platform_supports_pcre():
-    output = cmd_output('grep', '-P', "name='pre", 'setup.py', retcode=None)
+    output = cmd_output(GREP, '-P', "name='pre", 'setup.py', retcode=None)
```
tdeo: This is not related to this PR, but the tests were failing on my mac and I just quickly changed it. Let me know if you want me to take it out.

tdeo:
So I started working on that, and realized it was quite complex to use different folders in `~/.pre-commit/` for different hooks in the same repo.

Actually, if a remote repo contains several hooks which are listed with different (or even incompatible) additional dependencies, it's not handled today. Even worse, we could imagine the same repo being used for two hooks using different values on `language_version`.

Maybe it should be another project to be able to have local repos handle this?

asottile:
It does work today, different `language_version`s get their own virtualenv, `additional_dependencies` are installed across all configured hooks.  Do you have an example of it being broken?

tdeo:
So I tried it out using different `language_version` or `additional_dependencies` for both remote / local repos and it works the same way in both cases. 

And I couldn't find any broken example, at least for ruby, I guess it has something to do with `gem install` actually resolving the gems dependencies one by one and not altogether like `bundle` does (which was the motivation for https://github.com/pre-commit/pre-commit/pull/482). So worst case is several versions of the same gem will be installed and rbenv will pick one by default (the greater version actually).

Let me know if there's anything more to do before merging this (I'll rebase the branch on master as it's conflicting right now)

asottile:
I'm actually prototyping an additional approach to this, once I'm done I'd like to compare the two :)

asottile:
superseeded by #489

Thanks for putting great work into this, it definitely helped me with my implementation <3

PR #489, opened 2017-02-15
Local language hooks

asottile:
An alternate implementation of #484


asottile:
CC @tdeo here's my stab at this. let me know what you think

tdeo:
Using empty_templates is a good idea, so not to have to differentiate installation for each language.

However, why not also use the same logic of different venvs for different additional_dependencies for a non-local repository ?

asottile:
For "normal" repos they already have data to fill the two keys (repo, ref).  Also from our usage at least, `additional_dependencies` is super rare and even so already works in the current regime

tdeo:
It's true it's quite rare. And I agree the objective is not to create mirrors for everything (which will be solved with local hooks). I just had in mind to keep everything as close as possible.

But anyway, this will definitely address my needs, so let's go for it.

asottile:
Awesome, thanks for all the work and ideas and brainstorming on this @tdeo!

And while I think this has the potential to replace a lot of the tedium of the "mirror" repositories, I think that official support is still going to be better -- since it's much easier to configure a "traditional" hook repo than the language-local repos.  Either way, I think this definitely makes it easier for people to integrate tools and gives us the potential to deprecate our mirror repositories.

tdeo:
Awesome, you already released that!

Issue #457, opened 2016-12-31
InvalidConfigError: .pre-commit-config.yaml does not exist


livc:
Afer `pre-commit install` in my repo, I commit it and get:

```
➜  recommender_system git:(recommender_sys) ✗ git commit -m "test pre-commit"
An error has occurred: InvalidConfigError: File /Users/zhao/repo/book/.pre-commit-config.yaml does not exist
Check the log at ~/.pre-commit/pre-commit.log
```
pre-commit.log
```
An error has occurred: InvalidConfigError: File /Users/zhao/repo/book/.pre-commit-config.yaml does not exist
Traceback (most recent call last):
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/pre_commit/error_handler.py", line 47, in error_handler
    yield
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/pre_commit/main.py", line 181, in main
    return run(runner, args)
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/pre_commit/commands/run.py", line 210, in run
    repo_hooks = list(get_repo_hooks(runner))
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/pre_commit/commands/run.py", line 159, in get_repo_hooks
    for repo in runner.repositories:
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/cached_property.py", line 26, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/pre_commit/runner.py", line 43, in repositories
    config = load_config(self.config_file_path)
  File "/usr/local/Cellar/pre-commit/0.9.4/libexec/lib/python2.7/site-packages/pre_commit/clientlib/validate_base.py", line 41, in validate
    raise exception_type('File {} does not exist'.format(filename))
InvalidConfigError: File /Users/zhao/repo/book/.pre-commit-config.yaml does not exist


```

asottile:
Do you have a .pre-commit-config.yaml? You need one of those to tell pre-commit what hooks you want to run.  To see how to configure one, see http://pre-commit.com

livc:
[...]

PR #496, opened 2017-02-21
Do not run pre-commit if there is no config file

philipgian:
When no '.pre-commit-config.yaml' file exists while `pre-commit` hooks
are enabled, `pre-commit` returns an error and the action is aborted.
This is a very common scenario when pre-commit is added later on a
project and the user wants to work on a previous branch where the
configuration file does not exist.

This commit checks whether a config file exists and runs only the legacy
pre-commit hooks (if any) when it is missing.

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -37,7 +37,10 @@ def is_previous_pre_commit(filename):
     return any(hash in contents for hash in PREVIOUS_IDENTIFYING_HASHES)
 
 
-def install(runner, overwrite=False, hooks=False, hook_type='pre-commit'):
+def install(
+        runner, overwrite=False, hooks=False, hook_type='pre-commit',
+        skip_on_missing_conf=False
+):
     """Install the pre-commit hooks."""
     hook_path = runner.get_hook_path(hook_type)
     legacy_path = hook_path + '.legacy'
@@ -70,10 +73,12 @@ def install(runner, overwrite=False, hooks=False, hook_type='pre-commit'):
         else:
             pre_push_contents = ''
 
+        skip_on_missing_conf = 'true' if skip_on_missing_conf else 'false'
         contents = io.open(resource_filename('hook-tmpl')).read().format(
             sys_executable=sys.executable,
             hook_type=hook_type,
             pre_push=pre_push_contents,
+            skip_on_missing_conf=skip_on_missing_conf
         )
         pre_commit_file_obj.write(contents)
     make_executable(hook_path)
--- a/pre_commit/resources/hook-tmpl
+++ b/pre_commit/resources/hook-tmpl
@@ -10,6 +10,7 @@ retv=0
 args=""
 
 ENV_PYTHON='{sys_executable}'
+SKIP_ON_MISSING_CONF={skip_on_missing_conf}
 
 which pre-commit >& /dev/null
 WHICH_RETV=$?
@@ -35,6 +36,20 @@ if [ -x "$HERE"/{hook_type}.legacy ]; then
     if [ $? -ne 0 ]; then
         retv=1
     fi
+fi
+
+CONF_FILE=$(git rev-parse --show-toplevel)"/.pre-commit-config.yaml"
+if [ ! -f $CONF_FILE  ]; then
+    if [ $SKIP_ON_MISSING_CONF = true ] || [ ! -z $PRE_COMMIT_ALLOW_NO_CONFIG ]; then
+            echo '`.pre-commit-config.yaml` config file not found. Skipping `pre-commit`.'
+            exit $retv
+    else
+        echo 'No .pre-commit-config.yaml file was found\n'\
+             '- To temporarily silence this, run `PRE_COMMIT_ALLOW_NO_CONFIG=1 git ...`\n'\
+             '- To permanently silence this, install pre-commit with the `--allow-missing-config` option\n'\
+             '- To uninstall pre-commit run `pre-commit uninstall`'
+        exit 1
+	fi
 fi
 
 {pre_push}
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -75,6 +75,13 @@ def main(argv=None):
         '-t', '--hook-type', choices=('pre-commit', 'pre-push'),
         default='pre-commit',
     )
+    install_parser.add_argument(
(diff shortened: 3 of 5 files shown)


asottile:
There's a discussion about this somewhere in the issues but I'm struggling to find it.

This error *is* intentional and points out a user error.  I don't really want to change this.   If a user encounters this error they can simply `pre-commit uninstall` and move along with committing if they _really_ want to where otherwise silently doing nothing may hide an actual mistake (forgetting to commit the config file, misnaming it, etc.).

asottile:
For comparison, consider similar tools:

```
$ tox; echo $?
ERROR: toxini file 'tox.ini' not found
1
$ make; echo $?
make: *** No targets specified and no makefile found.  Stop.
2
$ pip install .; echo $?
Directory '.' is not installable. File 'setup.py' not found.
1
$ git status; echo $?
fatal: Not a git repository (or any of the parent directories): .git
128
```

philipgian:
Indeed. What you mentions is true, apart from the silently part. A warning gets printed.

But consider this. The equivalent to tox, make, pip, git is the `pre-commit` executable. If someone was to use for example `pip` with a script on a directory that may or may not contain the file (I think I've demonstrated a legitimate case for this with `pre-commit`), then I guess he/she would wrap the actual call and print a warning instead of failing.

If you manage to find the issue, please post it. I'd really like to read the arguments.

Being in the situation I describe, I too resulted in uninstalling the hooks. But I think it is rather cumbersome to do this every time. Plus it has the danger of forgetting to re-install them when working on an appropriate branch.

asottile:
I can't find it, perhaps I am not remembering correctly (or it was an internal issue).  This is the closet to what I found: https://github.com/pre-commit/pre-commit/issues/457

The situation you're in, is it that one branch will _never_ have a config file or is it waiting for merge?

philipgian:
Both actually. I've a couple of branches that should be rebased on top of the branch that first includes the config file. So in this situation, the problem should be resolved eventually and relatively quickly. But the problem also exists when branching from a stable tag (e.g. for a hotfix), where chances are that is gonna take longer for the config file to reach the latest stable tags. The situation gets worse if you try to include pre-commit on several repositories where the problem is multiplied.

asottile:
With this and the following accident this allows an unvetted commit:

```bash
# oops!
rm .pre-commit-config.yaml
git commit -m "foo"
```

I also really want to minimize the amount of logic in bash if possible.  I really think it's better to error and improve the error message to suggest uninstall than complicate the bash scripts

philipgian:
The above is true also in the case you uninstall and forget to re-install. It is also worse, as there is no warning message printed. As I said before, I think that the error message should be only in the `pre-commit` executable.

As far as I can tell, perhaps we could reach a middle ground by allowing the hooks template to be installed with an option. This could be a `--disallow-missing-config` option, which will set a variable on the template accordingly when installing. But this would also add some extra (minimal) logic to the bash script. What do you think ?

asottile:
hmm without complicating the bash we could add a `PRE_COMMIT_ALLOW_NO_CONFIG=1` or some sort of environment variable that could accomplish the same thing.  The error message could suggest accordingly:

```
No .pre-commit-config.yaml file was found
- To temporarily silence this, run `PRE_COMMIT_ALLOW_NO_CONFIG=1 git ...`
- To uninstall pre-commit run `pre-commit uninstall`
```

Though doesn't `--no-verify` solve the same thing (or do you also have a legacy script?)

philipgian:
No personally I don't. Others might. But I think that the solution to the problem should not be tailored to one's specific need, if there is a better solution.

What I was essentially proposing, was to let the option to statically set the variable you proposed to the chosen action. Perhaps we could also support overriding it from the environment.

asottile:
To be honest, I don't really want to support a strange edge case if nobody is using it due to the potential maintenance burden :)

asottile:
You know what, I changed my mind.  Let's add a flag to install which templates this in.  Write a simple integration test which demonstrates the behavior and I'll happily merge it :)

pre_commit/commands/install_uninstall.py
```diff
@@ -37,7 +37,8 @@ def is_previous_pre_commit(filename):
     return any(hash in contents for hash in PREVIOUS_IDENTIFYING_HASHES)
 
 
-def install(runner, overwrite=False, hooks=False, hook_type='pre-commit'):
+def install(runner, overwrite=False, hooks=False, hook_type='pre-commit',
+            skip_on_missing_conf=False):
```
asottile: p3: our [style guidelines](https://github.com/pre-commit/pre-commit/blob/41dcaff3fb53fb7819a1d783d67a9ccb42464c1d/CONTRIBUTING.md#multi-line-method-invocation) would suggest one of these instead:

```python
def install(
        runner, overwrite=False, hooks=False, hook_type='pre-commit',
        skip_on_missing_conf=False,
):
```

or 

```python
def install(
        runner,
        overwrite=False,
        ...
):
```

The reasoning here is:
- Prevents diff churn for adding / removing arguments
- Prevents merge conflicts (sometimes)
- Prevents arbitrary indentation that's awkward to do in editors

pre_commit/resources/hook-tmpl
```diff
@@ -37,6 +38,21 @@ if [ -x "$HERE"/{hook_type}.legacy ]; then
...
 fi
 
+CONF_FILE=$(git rev-parse --show-toplevel)"/.pre-commit-config.yaml"
+if [ ! -f $CONF_FILE  ]; then
+    if [ $SKIP_ON_MISSING_CONF = true ] || [ ! -z $PRE_COMMIT_ALLOW_NO_CONFIG ]; then
+            echo '`.pre-commit-config.yaml` config file not found. Skipping `pre-commit`.'
+            exit $retv
+    else
+        echo 'No .pre-commit-config.yaml file was found
+        - To temporarily silence this, run `PRE_COMMIT_ALLOW_NO_CONFIG=1 git ...`
```
asottile: does this cause all of these to be indented?

pre_commit/commands/install_uninstall.py
philipgian: OK!

pre_commit/resources/hook-tmpl
philipgian: You are right. Fixed!

asottile:
This is available in 0.13.3!

Thanks again :D

PR #497, opened 2017-02-22
Make hook-tmpl resilient to future changes

asottile:
I realized this wasn't necessarily rollback safe.  This makes it slightly more future-proof.

This feature really isn't used too much anyway (we usually run install with `--force`)

Perhaps the default behaviour of install should be changed as well?


asottile:
Actually, I think this is mostly unnecessary.  We're idempotently writing the file so we don't suffer from the install problem.  The template can change freely and will still be properly reinstalled.

I'm going to scale this back just to have the simplifications and not need the new generated hashes

PR #498, opened 2017-02-23
Python pip requirements file

jzafran:


asottile:
This should already work without new code:

```yaml
additional_dependencies: [-rrequirements.txt]
```

That said, this will suffer from a reproducibility issue as the cache key will not properly represent the installed state and I don't think it's a good idea to encode knowledge of parsing requirements files.  Especially those outside the hook repository

jzafran:
ahh, awesome! I don't know why I didn't think of trying that...  thanks!

PR #501, opened 2017-02-25
Add a --show-diff-on-failure option

### LINKED ISSUE #253: Add option to print diff if hooks failed and files were changed.
I'm currently debugging a pre-commit failure in only one Python version that happens on Travis.

Not sure if it is out of scope of the project, but it would be handy to have some kind of `--print-diff` option that simply prints a diff of the files on failure. It would generally only be useful in CI when you can't just go back and inspect what changed.

Currently all I see in Travis logs is

```
Reordering imports in pre_commit/main.py
Reordering imports in pre_commit/clientlib/validate_base.py
Reordering imports in tests/main_test.py
```

which doesn't really tell me what changed or give me useful hints on tracking down the problem.

As an example, here's the actual diff. It's clear now what actually happened and why this only happened on Python 2.6:

```
diff --git a/pre_commit/clientlib/validate_base.py b/pre_commit/clientlib/validate_base.py
index 707bdde..af18725 100644
--- a/pre_commit/clientlib/validate_base.py
+++ b/pre_commit/clientlib/validate_base.py
@@ -1,11 +1,11 @@

-import argparse
 import os.path
 import re
 import sys

+import argparse
 import jsonschema
 import jsonschema.exceptions
 import yaml
```

asottile:
Resolves #253

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import logging
 import os
+import subprocess
 import sys
 
 from pre_commit import color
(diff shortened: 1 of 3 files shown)


asottile:
This was released with 0.13.4!

asottile:
errr... 0.13.5, somehow 0.13.4 contained incorrect files when pushed to pypi O_o

Issue #511, opened 2017-03-26
Known issue: appveyor is currently failing


asottile:
See upstream bugs:
- https://github.com/git-for-windows/git/issues/1106
- https://github.com/appveyor/ci/issues/1428

asottile:
https://github.com/git-for-windows/git/issues/1106 is resolved
https://github.com/appveyor/ci/issues/1483 tracks the updates to appveyor

elaichi:
I was demoing `pre-commit` to my manager and he immediately noticed the "failing" status on the homepage. I then had to explain that "the issue is actually with their CI", etc, etc, "trust me, it's a really cool project", etc, etc.
Glad to see that this is almost fixed :)

![pre-commit_by_yelp](https://cloud.githubusercontent.com/assets/1084792/24913776/31a42fc6-1ea0-11e7-8a68-819b7abd154a.png)

asottile:
This is the new issue to watch: https://github.com/appveyor/ci/issues/1482

asottile:
I think this is the new issue to watch: https://github.com/appveyor/ci/issues/1542

asottile:
Yay it looks fixed!  Going to retrigger a job so our badge goes green again: https://ci.appveyor.com/project/asottile/pre-commit/build/job/8ije5h0y9nklpytr

asottile:
All green!

PR #513, opened 2017-03-27
Fix regression in 0.13.5 with always_run + files

asottile:


asottile:
Appveyor is a known failure: #511

PR #515, opened 2017-03-30
Add a `pre-commit sample-config` command

asottile:
CC @elaichi

### DIFF
--- a/pre_commit/commands/sample_config.py
+++ b/pre_commit/commands/sample_config.py
@@ -1,0 +1,25 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+# TODO: maybe `git ls-remote git://github.com/pre-commit/pre-commit-hooks` to
+# determine the latest revision?  This adds ~200ms from my tests (and is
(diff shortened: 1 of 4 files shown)


pre_commit/commands/sample_config.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+# TODO: maybe `git ls-remote git://github.com/pre-commit/pre-commit-hooks` to
+# determine the latest revision?  This adds ~200ms from my tests (and is
+# significantly faster than https:// or http://).  For now, periodically
+# manually updating the revision is fine.
```
chriskuehl: Another option is to read https://api.github.com/repos/pre-commit/pre-commit-hooks/tags which may (or may not) be easier than invoking git.

revolter:
Shouldn't this also save the `.pre-commit-config.yaml` file?

asottile:
@revolter by product of least surprise it just shows you a sample configuration -- if you'd like to save it, that can be done simply with `pre-commit sample-config > .pre-commit-config.yaml`

PR #517, opened 2017-04-01
Windows: enable ANSI escape support in console

### LINKED ISSUE #198: Windows: Color support
On unixlike platforms we support coloring via ANSI escapes.  Windows supports coloring somehow, figure out how that works and implement it.

hackedd:
As of Windows 10, the Windows console supports (some) ANSI escape sequences, but it nees to be enabled using [`SetConsoleMode`](https://msdn.microsoft.com/en-us/library/windows/desktop/mt638032(v=vs.85).aspx) first.

If more extensive support is needed, or you want this to work on earlier versions of Windows, it would be probably be better to use a package like [colorama](https://pypi.python.org/pypi/colorama) instead, which automatically translates ANSI escape sequences to the equivalent Windows API calls. It also contains functions to format colored text.

This fixes #198.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,6 +8,7 @@ omit =
     setup.py
     # Don't complain if non-runnable code isn't run
     */__main__.py
+    pre_commit/color_windows.py
 
 [report]
 show_missing = True
--- a/pre_commit/color.py
+++ b/pre_commit/color.py
@@ -1,6 +1,14 @@
 from __future__ import unicode_literals
 
+import os
 import sys
+
+if os.name == 'nt':  # pragma: no cover (windows)
+    from pre_commit.color_windows import enable_virtual_terminal_processing
+    try:
+        enable_virtual_terminal_processing()
+    except WindowsError:
+        pass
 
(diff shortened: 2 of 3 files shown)


asottile:
Oh hmm! I was under the impression that windows had its own set of color escapes -- would those be more portable to more versions of windows?

Sorry I've been so slow to get to this, I've been traveling :)

hackedd:
I'm not aware of any color escapes for the windows terminal. As far as I know, for earlier versions of Windows, text attributes can only be set by calling the [`SetConsoleTextAttribute`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686047(v=vs.85).aspx) function (which is what [colorama does](https://github.com/tartley/colorama/blob/f0f3220/colorama/winterm.py#L82)).

asottile:
Ok! In that case this seems fine -- can you add the new file to the excluded files in `.coveragerc`

asottile:
Thanks!

asottile:
This feature is in 0.14.0

Issue #519, opened 2017-04-07
 Unknown Encoding error


awebdeveloper:
I get the following error on commit

```
An unexpected error has occurred: LookupError: unknown encoding:
Traceback (most recent call last):
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/error_handler.py", line 47, in error_handler
    yield
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/main.py", line 195, in main
    return run(runner, args)
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/commands/run.py", line 210, in run
    repo_hooks = list(get_repo_hooks(runner))
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/commands/run.py", line 159, in get_repo_hooks
    for repo in runner.repositories:
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/cached_property.py", line 26, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/runner.py", line 46, in repositories
    repository.require_installed()
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/repository.py", line 174, in require_installed
    _install_all(self._venvs, self.repo_config['repo'])
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/repository.py", line 68, in _install_all
    for cmd_runner, language_name, version, deps in venvs
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/repository.py", line 69, in <genexpr>
    if not _installed(cmd_runner, language_name, version, deps)
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/repository.py", line 60, in _installed
    _read_installed_state(cmd_runner, venv) == _state(additional_deps)
  File "/usr/local/Cellar/pre-commit/0.13.2/libexec/lib/python2.7/site-packages/pre_commit/repository.py", line 43, in _read_installed_state
[...]

PR #524, opened 2017-04-19
Default to UTF-8 encoding in open() calls

asottile:
I really don't like this, but I don't have access to vscode to try and reproduce the bug.

I'm also tempted to try and add `LANG` to vscode if I can find the part in their codebase that calls out to git.

I'm also confused why python doesn't handle this on osx (it seems to work on linux and windows).

For now this is a workaround branch, I don't really want to merge this until I get to the bottom of it


asottile:
Closed via #519

PR #530, opened 2017-05-04
Adds support for 'log_file' in hook config

### LINKED ISSUE #499: A config option to store hook output into a file
Hi, this is a feature request and if people would like it then I will submit a pull request.

What I would like is to be able to specify in the config an output file on  a per hook basis. When the hook is run, it will output to that file (as well as to STDOUT as currently).

The reason I want this is because I want to run pre-commit on Jenkins and I want the reports of certain tools to be stored in files so that they can be easily consumed by Jenkins.

alex-hutton:
Specify a filename on a per hook basis and
pre-commit will write the STDOUT and STDERR
of that hook into the file. Useful for CI.

Resolves #499.

### DIFF
--- a/pre_commit/output.py
+++ b/pre_commit/output.py
@@ -71,8 +71,17 @@ def write(s, stream=stdout_byte_stream):
     stream.flush()
 
 
-def write_line(s=None, stream=stdout_byte_stream):
-    if s is not None:
-        stream.write(five.to_bytes(s))
-    stream.write(b'\n')
-    stream.flush()
+def write_line(s=None, stream=stdout_byte_stream, logfile_name=None):
+    def output_streams():
+        yield stream
+        try:
+            with open(logfile_name, 'ab') as logfile:
+                yield logfile
+        except (TypeError, IOError):
+            pass
+
+    for output_stream in output_streams():
+        if s is not None:
+            output_stream.write(five.to_bytes(s))
+        output_stream.write(b'\n')
+        output_stream.flush()
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -211,6 +211,35 @@ def test_run(
     )
 
 
+def test_run_output_logfile(
+        cap_out,
+        tempdir_factory,
+        mock_out_store_directory,
+):
+
+    expected_output = (
+        b'This is STDOUT output\n',
+        b'This is STDERR output\n',
+    )
+
+    git_path = make_consuming_repo(tempdir_factory, 'logfile_repo')
+    with cwd(git_path):
+        _test_run(
+            cap_out,
+            git_path, {},
+            expected_output,
+            expected_ret=1,
+            stage=True
+        )
+    logfile_path = os.path.join(git_path, 'test.log')
+    assert os.path.exists(logfile_path)
+    with open(logfile_path, 'rb') as logfile:
+        logfile_content = logfile.readlines()
+
+    for expected_output_part in expected_output:
+        assert expected_output_part in logfile_content
+
+
 def test_always_run(
         cap_out, repo_with_passing_hook, mock_out_store_directory,
 ):
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -53,6 +53,7 @@ MANIFEST_HOOK_DICT = schema.Map(
         '^$',
     ),
     schema.Optional('language_version', schema.check_string, 'default'),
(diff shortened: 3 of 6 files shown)


setup.py
```diff
@@ -41,6 +41,7 @@
     install_requires=[
         'aspy.yaml',
         'cached-property',
+        'contextlib2',
```
asottile: I actually run a youtube channel and my next episode in "porting to python 3" is about proper use of backports.  Haven't written up the slides yet or I'd link them but for backported modules you only want to conditionally install them on the versions that need them (otherwise we're installing contextlib2 in python3 where it's entirely unnecessary).

Here's the basic gist though: https://github.com/pre-commit/pre-commit/blob/v0.8.2/setup.py#L50-L52

pre_commit/output.py
```diff
@@ -1,6 +1,10 @@
 from __future__ import unicode_literals
 
 import sys
+try:
+    from contextlib import ExitStack
+except ImportError:
+    from contextlib2 import ExitStack
```
asottile: when possible, prefer putting conditional imports after all the other imports (the import reorderer stops when it sees non-import code)

tests/commands/run_test.py
```diff
@@ -211,6 +211,37 @@ def test_run(
...
+
+    expected_output = (b'This is STDOUT output',
+                       b'This is STDERR output')
+
+    def raise_error_on_none(*args):
+        if len(args) and args[0] is None:
+            raise TypeError()
+        return mock.DEFAULT
+
+    with mock.patch('pre_commit.output.open', mock.mock_open()) as mocked_open:
```
asottile: I'd rather see a real integration test which the file is actually written -- especially since the code as written doesn't properly write the file -- There's fixtures in testing/fixtures for setting up a git repository (and fixtures in testing/conftest.py as well)

alex-hutton:
Thanks for the helpful feedback. Please take a look at the changes.

pre_commit/output.py
```diff
@@ -71,8 +76,18 @@ def write(s, stream=stdout_byte_stream):
     stream.flush()
 
 
-def write_line(s=None, stream=stdout_byte_stream):
-    if s is not None:
-        stream.write(five.to_bytes(s))
-    stream.write(b'\n')
-    stream.flush()
+def write_line(s=None, stream=stdout_byte_stream, logfile_name=None):
+    output_streams = [stream, ]
```
asottile: just write `[stream]`

setup.py
```diff
@@ -46,6 +46,9 @@
         'six',
         'virtualenv',
     ],
+    extras_require={
+        ':python_version<"3.0"': ['contextlib2', ],
```
asottile: just write `['contextlib2']`

pre_commit/output.py
```diff
@@ -71,8 +76,18 @@ def write(s, stream=stdout_byte_stream):
...
+
+    with ExitStack() as stack:
+        try:
+            logfile = stack.enter_context(open(logfile_name, 'ab'))
+            output_streams.append(logfile)
+        except (TypeError, IOError):
+            pass
```
asottile: did you try the noop ctx idea from the first review? what did you think?

asottile: Oh wut, my original comment never posted! lol

Here's an idea that avoids needing `contextlib2` (we use this pattern elsewhere in pre-commit too!)

```python
if log_file:
    ctx = open(...)
else:
    ctx = noop_ctx()
with ctx:
    ...
```

alex-hutton:
Here's a way of doing it without contextlib2.

asottile:
Thanks!  I'll touch this up and put it into the next release :)

asottile:
One thought: I think it only makes sense to set this in the consuming repository (and not the remote hook definition) -- would you be ok with me moving the configuration setting into the pre-commit config instead?

alex-hutton:
Thanks Anthony. And thanks for the awesome software!

I received a github comment as an email but I don't see it in github itself:

> One thought: I think it only makes sense to set this in the consuming repository (and not the remote hook definition) -- would you be ok with me moving the configuration setting into the pre-commit config instead?

My answer to that is, sure, go ahead 👍 . I'm not sure what you are referring to, though. Did I put the schema definition in the wrong place?

asottile:
Ah yeah I deleted the comment as I'll handle it better at a later date. Sorry for the email spam!

PR #538, opened 2017-05-13
Add pass_filenames hook option

absassi:
This option controls whether filenames are passed along as arguments to
the hook program.

### DIFF
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -731,3 +731,29 @@ def test_files_running_subdir(
                 tempdir_factory=tempdir_factory,
             )
         assert 'subdir/foo.py'.replace('/', os.sep) in stdout
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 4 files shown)


tests/commands/run_test.py
```diff
@@ -731,3 +731,29 @@ def test_files_running_subdir(
...
+        (False, [], b''),
+        (True, ['some', 'args'], b'some args foo.py'),
+        (False, ['some', 'args'], b'some args'),
+    )
+)
+def test_pass_filenames(
+        cap_out, repo_with_passing_hook, mock_out_store_directory,
+        pass_filenames,
+        hook_args,
+        expected_out,
```
asottile: this seems unused

absassi: It is used in line 758, to check hook output, since it prints its args.

asottile: I am blind!  Thanks :D

asottile:
Thanks for the PR :D

Issue #257, opened 2015-07-31
Can't match files based on shebang language


chriskuehl:
Often we have scripts written in Python with no extension. Hooks usually miss these since they match on extension.

Would be nice if pre-commit could somehow read shebangs. Maybe something to think about when trying to solve #220 as well. Not sure how to best approach this yet.

chriskuehl:
I spent a little time trying to think of a good way to handle this, as well as #254 (ability for hooks to target types of objects, e.g. file/symlink/submodule) and #220 (ability for hooks to target "text" files and avoid list of "text" extensions).

Here's my suggestion:

#### Files get automatically tagged

Each file gets tagged with a list of tags describing it. For example, an executable file `something.py` might be classified `['file', 'text', 'python', 'executable']`.

#### Hooks target file types

Hooks can then specify `types` (which defaults to `file`):

```
# hooks.yaml
-   id: trailing-whitespace
    types: [text]
    entry: trailing-whitespace-fixer

-   id: symlinks-point-inside-repo
    types: [symlink]

-   id: executables-have-shebang
    types: [executable]

-   id: double-quote-string-fixer
    types: [python, ruby, javascript]  # if this hook actually worked on all of these
```

There's an added benefit in that we no longer need to duplicate the hook extensions in repos if we want to change the path.

```
# .pre-commit-config.yaml
-   id: puppet-lint
    path: ^modules/ocf
```

Currently we have to do `exclude: ^modules/(?!ocf).*$` instead which will get really complex as soon as we need to start excluding anything else. Now instead of hooks targeting files by extension, the suggested way is targeting them by type. (You can still do it by extension, of course.)

`types` is an OR operation; a file with any of the tags will match.

I would suggest not including an `all` types alias, since I think it risks people copy-pasting it without realizing it includes symlinks and submodules. `[file, symlink, submodule]` is not very cumbersome for the very few hooks that will actually want it (the only existing one I can think of is `check-case-conflict`).


#### Implementation

Making this happen is fairly easy (and I'd volunteer to do the work, since I really want to check files with shebangs):

* Write a file classifier. A very functional classifier is easy to write (check if text, then check extension, fallback to reading shebang) and covers almost all cases.

* A little plumbing

[...]

PR #551, opened 2017-07-02
Types

asottile:
This is the big one!

Should resolve the following issues:
- [pre-commit should ignore symlinks by default](https://github.com/pre-commit/pre-commit/issues/191)
- [pre-commit should ignore submodules by default](https://github.com/pre-commit/pre-commit/issues/191#issuecomment-113249998)
- [A special "text" type for text files](https://github.com/pre-commit/pre-commit/issues/220)
- [Can't match files based on shebangs](https://github.com/pre-commit/pre-commit/issues/257)
- [Run pre-commit-hooks#end-of-file-fixer on all text files](https://github.com/pre-commit/pre-commit-hooks/issues/61)

Some notable things:

- `files` and `types` are now both optional -- I'm not sure if this is ok? I'm considering making these inclusive-or-required (with `always_run`).  `files` defaults to `''` (matching all) and `types` defaults to `['file']` matching only files (this excludes submodules and symlinks)
- `files` and `types` are evaluated with *and*, that is, for a file to be linted it must match *both* the `files` regex and the `types` listed.
- Each tag inside `types` is evaluated with *and* -- that is, for a file to be chosen for linting it must be classified as all of the tags.
- A separate library ([identify](https://github.com/chriskuehl/identify)) was chosen to allow the inevitable rapid updates to classification without encumbering the pre-commit release schedule.
- The implementation allows for extension of "tags" by pre-commit itself, I think something cool that pre-commit could do is a `newly-added` tag (would cut out a lot of boilerplate in several hooks).

Known breaking changes:
- [pre-commit-hooks#check-symlinks](https://github.com/pre-commit/pre-commit-hooks/blob/e9446ff0cc8ffa65309e0a37c247518127a3bf18/.pre-commit-hooks.yaml#L59-L65) - this hook intentionally lints `symlink` "files" and will now need `types: ['symlink']` to work properly
- `always_run` without `files` now defaults to `files: ''` instead of `files: '^$'` -- I think this actually makes more sense given the choices in the rest of the framework and is a "least surprise" outcome here (it also makes the code simple!)

@chriskuehl and I have put a ton of work into this (my branch is roughly based on his prototype branch) :tophat: *hattip*

@Lucas-C is probably interested in this PR so... hi!

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -49,7 +49,13 @@ MANIFEST_HOOK_DICT = schema.Map(
         'files', schema.check_and(schema.check_string, schema.check_regex),
         '',
     ),
+    schema.Optional(
+        'exclude',
+        schema.check_and(schema.check_string, schema.check_regex),
+        '^$',
+    ),
     schema.Optional('types', schema.check_array(check_type_tag), ['file']),
+    schema.Optional('exclude_types', schema.check_array(check_type_tag), []),
 
     schema.Optional(
         'additional_dependencies', schema.check_array(schema.check_string), [],
@@ -58,11 +64,6 @@ MANIFEST_HOOK_DICT = schema.Map(
     schema.Optional('always_run', schema.check_bool, False),
     schema.Optional('pass_filenames', schema.check_bool, True),
     schema.Optional('description', schema.check_string, ''),
-    schema.Optional(
-        'exclude',
-        schema.check_and(schema.check_string, schema.check_regex),
-        '^$',
-    ),
     schema.Optional('language_version', schema.check_string, 'default'),
     schema.Optional('log_file', schema.check_string, ''),
     schema.Optional('minimum_pre_commit_version', schema.check_string, '0'),
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -5,6 +5,7 @@ import argparse
 import functools
 
 from aspy.yaml import ordered_load
+from identify.identify import ALL_TAGS
 
 import pre_commit.constants as C
 from pre_commit import schema
@@ -16,6 +17,14 @@ def check_language(v):
     if v not in all_languages:
         raise schema.ValidationError(
             'Expected {} to be in {!r}'.format(v, all_languages),
+        )
+
+
+def check_type_tag(tag):
+    if tag not in ALL_TAGS:
+        raise schema.ValidationError(
+            'Type tag {!r} is not recognized.  '
+            'Try upgrading identify and pre-commit?'.format(tag),
         )
 
 
@@ -36,10 +45,11 @@ MANIFEST_HOOK_DICT = schema.Map(
         'language', schema.check_and(schema.check_string, check_language),
     ),
 
-    schema.Conditional(
+    schema.Optional(
         'files', schema.check_and(schema.check_string, schema.check_regex),
-        condition_key='always_run', condition_value=False,
+        '',
     ),
+    schema.Optional('types', schema.check_array(check_type_tag), ['file']),
 
     schema.Optional(
         'additional_dependencies', schema.check_array(schema.check_string), [],
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -43,12 +43,14 @@ def get_changed_files(new, old):
     )[1].splitlines()
 
 
-def filter_filenames_by_types(filenames, types):
-    types = frozenset(types)
-    return tuple(
-        filename for filename in filenames
-        if tags_from_path(filename) >= types
-    )
+def filter_filenames_by_types(filenames, types, exclude_types):
+    types, exclude_types = frozenset(types), frozenset(exclude_types)
+    ret = []
+    for filename in filenames:
+        tags = tags_from_path(filename)
+        if tags >= types and not tags & exclude_types:
+            ret.append(filename)
+    return tuple(ret)
 
 
 def get_filenames(args, include_expr, exclude_expr):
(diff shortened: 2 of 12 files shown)


pre_commit/commands/run.py
```diff
@@ -37,7 +43,15 @@ def get_changed_files(new, old):
     )[1].splitlines()
 
 
-def get_filenames(args, include_expr, exclude_expr):
+def filter_filenames_by_types(filenames, types):
+    types = frozenset(types)
+    return tuple(
+        filename for filename in filenames
+        if tags_from_path(filename) >= types
```
Lucas-C: Nice ! :+1:

tests/commands/run_test.py
```diff
@@ -153,6 +153,20 @@ def test_hook_that_modifies_but_returns_zero(
...
+):
+    git_path = make_consuming_repo(tempdir_factory, 'types_repo')
+    with cwd(git_path):
+        stage_a_file('bar.py')
+        stage_a_file('bar.notpy')
+        args = _get_opts()
+        ret, printed = _do_run(cap_out, git_path, args)
+        assert ret == 1
+        assert b'bar.py' in printed
+        assert b'bar.notpy' not in printed
```
Lucas-C: The new `types_repo/bin/hook.sh` file is not tested ?

Lucas-C:
Great ! Thanks a lot for your efforts !

tests/commands/run_test.py
asottile: All `hook.sh` does is print the filenames and invariantly fail -- I think this tests that?

pre_commit/commands/run.py
chriskuehl: Do you think it makes sense for the list of tags to be an OR rather than an AND?

I'm wondering if there's a case where someone will do something like `types: [python, ruby]` and be surprised that it matches nothing rather than all python and ruby files.

I'm having trouble coming up with a scenario where I'd want to target files which have multiple tags... but I also can't really think of too many cases where I'd want to target both python and ruby but not just `text`, so maybe both cases are pretty rare?

My original (much more complicated) branch just checked for any intersection of the tags, but I didn't really spend much time considering the alternative.

tests/commands/run_test.py
Lucas-C: Yes sure. Actually I thought `hook.sh` role was to test shebang detection.
My mistake

pre_commit/clientlib.py
```diff
@@ -19,6 +20,14 @@ def check_language(v):
         )
 
 
+def check_type_tag(tag):
+    if tag not in ALL_TAGS:
+        raise schema.ValidationError(
+            'Type tag {!r} is not recognized.  '
+            'Try upgrading identify and pre-commit?'.format(tag),
+        )
```
asottile: Yeah... the more common errors they're going to get is "no file hooks.yaml" and "missing key 'files'" though -- and unfortunately no way to retroactively improve error messages in old versions :S

pre_commit/commands/run.py
asottile: Yeah I do think it should be `AND` - I also think I should implement `exclude_types` as we have (`files`, `exclude`) but no way to do that with `types`.

asottile: Actually, maybe it should be `OR` - though I can't think of a compelling reason why someone would ever use _more than one tag_.

chriskuehl: Yeah, I can't really come up with a real use-case for either. It's still possible to get the OR behavior by specifying the hook twice if you really need it. I'm fine with either way.

fwiw all of the standard hooks were easy to write with just one type: https://github.com/chriskuehl/pre-commit-hooks/blob/new-typers/hooks.yaml

asottile: shrugs, easier to leave it how it is so I'll do that.  If someone has a legitimate usecase for `OR` that neither `exclude_types` nor separate hooks satisfies we can revisit this.

asottile:
Depending on how migrated repositories are, I'd suggest the following (at least temporary) if you plan to move repositories to use `types`:

## if you still are maintaining `hooks.yaml`

Convert your hooks.yaml to a skeleton providing the minimum possible to satisfy `<0.15.0` and add a `minimum_pre_commit_version: 0.15.0` key to each hook.

For example ([original here](https://github.com/pre-commit/pre-commit-hooks/blob/e9446ff0cc8ffa65309e0a37c247518127a3bf18/hooks.yaml#L1-L7)):

```yaml
-   id: autopep8-wrapper
    language: system
    name: upgrade-your-pre-commit-version
    entry: upgrade-your-pre-commit-version
    files: ''
    minimum_pre_commit_version: 0.15.0
```

- For `pre-commit<0.6.7`, an error about not being able to execute `upgrade-your-pre-commit-version` will be produced
- For `pre-commit>=0.6.7,<0.12.0` an error suggesting an upgrade
- For `pre-commit>=0.12.0`, `.pre-commit-hooks.yaml` will be used instead (see below)

## forward-backward compatible `.pre-commit-hooks.yaml`

- Add `minimum_pre_commit_version: 0.15.0` to each hook using `types`
- Change `files: ...` to `files: ''` (this is the default value when `files` is missing)
- The entire reason for the noop `files` is old versions of pre-commit won't even load the configuration far enough to produce a useful error message on `minimum_pre_commit_version`

For example ([original here](https://github.com/pre-commit/pre-commit-hooks/blob/e9446ff0cc8ffa65309e0a37c247518127a3bf18/.pre-commit-hooks.yaml#L15-L20)):

```yaml
-   id: check-ast
    name: Check python ast
    description: Simply check whether the files parse as valid python.
    entry: check-ast
    language: python
    files: ''
    types: [python]
    minimum_pre_commit_version: 0.15.0
```

- `pre-commit<0.12.0` won't read this file
- `pre-commit>0.12.0,<=0.15.0` will suggest an upgrade
- `pre-commit>=0.15.0` will use `types` and `files` will be a noop

When you don't care about old versions, see below

## once you no longer care about old versions of pre-commit

- Rremove the noop `files: ''` and the `minimum_pre_commit_version`.

PR #553, opened 2017-07-03
Remove legacy 'hooks.yaml' file support

asottile:
I'll wait to merge this until July 24th (6 months since deprecation in 0.12.0).


chriskuehl:
After merging, it might be a good idea to stop listing repos which don't have `.pre-commit-hooks.yaml` on the website to avoid confusion.

asottile:
Yeah I plan to change the script to not fall back (and comment them out temporarily)

PR #556, opened 2017-07-09
Use a more intelligent default language version

### LINKED ISSUE #365: Detect the python version when creating `default` py_envs
Given two separate virtualenvs, one running with python3 and the other running with python2, and each with the appropriate pre-commit installed (`pip install pre-commit` in the first, `pip3 install pre-commit` in the second), the check-ast plugin fails to parse Python 2 (or 3) syntax correctly, depending on which virtualenv pre-commit was installed in first.

Suggestions: 
- the ~/.pre-commit cache should instead be a per virtualenv cache, rather than per user.
- instead of maintaining the cache, consider just using pip dependencies instead.

Example failure in Python 2 venv if Python 3 venv installs pre-commit first:

```
p2/main.py: failed parsing with python3.4:

      File "/home/ubuntu/.pre-commit/repo_3mr61_f/py_env-default/lib/python3.4/site-packages/pre_commit_hooks/check_ast.py", line 23, in check_ast
        ast.parse(open(filename, 'rb').read(), filename=filename)
      File "/usr/lib/python3.4/ast.py", line 35, in parse
        return compile(source, filename, mode, PyCF_ONLY_AST)
      File "p2/main.py", line 178
        except socket.error, e:
                           ^
    SyntaxError: invalid syntax
```

Example failure in Python 3 venv if Python 2 venv installs pre-commit first:

```
p3/__main__.py: failed parsing with python:

      File "/home/ubuntu/.pre-commit/repoCzM4lg/py_env-default/local/lib/python2.7/site-packages/pre_commit_hooks/check_ast.py", line 23, in check_ast
        ast.parse(open(filename, 'rb').read(), filename=filename)
      File "/usr/lib/python2.7/ast.py", line 37, in parse
        return compile(source, filename, mode, PyCF_ONLY_AST)
      File "p3/__main__.py", line 30
        "c=%s" % args.c, file=sys.stderr)
                                ^
    SyntaxError: invalid syntax
```

`rm -fr ~/.pre-commit` was run between each test.

asottile:
Resolves #365

Thoughts on this?

It has the potential to increase the number of virtualenvs when `language_version` is not specified in `.pre-commit-config.yaml` but at least they won't be defaulting to \~\*\~\*\~mysterious\~\*\~\*\~ undefined versions of python.

This should also remove the need to have `language_version` for python in most cases (as it'll now be properly inferred from the executable used to run pre-commit -- which is already [*kind of* the status quo](https://github.com/pre-commit/pre-commit/blob/853cbecd4e15aeb59e4730320dc90fe16afa219e/pre_commit/languages/python.py#L71-L74))

### DIFF
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -10,16 +10,18 @@ from pre_commit.languages import script
 from pre_commit.languages import swift
 from pre_commit.languages import system
 
-# A language implements the following constant and two functions in its module:
+# A language implements the following constant and functions in its module:
 #
 # # Use None for no environment
 # ENVIRONMENT_DIR = 'foo_env'
 #
-# def install_environment(
-#         repo_cmd_runner,
-#         version='default',
-#         additional_dependencies=(),
-# ):
+# def get_default_version():
+#     """Return a value to replace the 'default' value for language_version.
+#
+#     return 'default' if there is no better option.
+#    """
+#
+# def install_environment(repo_cmd_runner, version, additional_dependencies):
 #     """Installs a repository in the given repository.  Note that the current
 #     working directory will already be inside the repository.
 #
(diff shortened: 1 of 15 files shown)


asottile:
test failures are expected, I'll fix those up

PR #558, opened 2017-07-09
Fix windows virtualenv issue

evhub:
With `v0.15`, I started getting the following bug when running `pre-commit install -f --install-hooks` on my Windows 10 machine:
```
An unexpected error has occurred: CalledProcessError: Command: ('C:\\Users\\Evan\\.pre-commit\\repo8d84nqt_\\py_env-python3.6\\Scripts\\pip.exe', 'install', '.')
Return code: 1
Expected return code: 0
Output: (none)
Errors: (none)
```

I managed to track the bug down to `virtualenv` not actually creating the `py_env-python3.6` directory, and from there to `virtualenv` not having been passed an actual executable on the `-p` option. The reason this was not happening was because `norm_version` was returning `version` instead of `find_executable(version)` when an executable was found. Thus, `find_executable` discovered that `version` had an executable, but then refused to actually pass that executable along.

This PR fixes the `norm_version` logic to return `find_executable(version)` instead of `version`, which completely resolved the issue on my machine.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -86,8 +86,9 @@ def get_default_version():
 def norm_version(version):
     if os.name == 'nt':  # pragma: no cover (windows)
         # Try looking up by name
-        if find_executable(version) and find_executable(version) != version:
-            return version
+        version_exec = find_executable(version)
+        if version_exec and version_exec != version:
+            return version_exec
 
         # If it is in the form pythonx.x search in the default
         # place on windows


asottile:
woops, I broke this in 0.15.1 :(

asottile:
actually hmm... I didn't really change the logic there: https://github.com/pre-commit/pre-commit/pull/556/files#diff-235f2434ddb56aaeef64d44558e82943R89

evhub:
@asottile I just tested on `v0.15.0`, and it worked, but when I reinstalled `v0.15.1`, it failed again, so whatever broke this for me happened between `v0.15.0` and `v0.15.1`. Re-installing my local copy of `v0.15.1` with the change in this PR fixed it back, though.

asottile:
Before I merge, I'd like to understand what's actually different -- can you produce the output of these commands for me?

```python
import distutils.spawn; print(distutils.spawn.find_executable("python3.6"))
import pre_commit.parse_shebang; print(pre_commit.parse_shebang.find_executable("python3.6"))
```

I expect both to return `None`, but now I'm not so sure :)

asottile:
Also your .pre-commit-config.yaml will probably be useful as well :)

evhub:
Here is my [`.pre-commit-config.yaml`](https://gist.github.com/evhub/3ea1d5fa3a0284dea53e5bef92b54894), and both commands above produce `C:\Anaconda3\python3.6.exe` as output.

asottile:
Ah I see now:

```
Anthony@AnthonysDesktop MINGW64 ~/Desktop/git/pre-commit (master)
$ cp venv/Scripts/python.exe venv/Scripts/python2.7.exe
(venv)
Anthony@AnthonysDesktop MINGW64 ~/Desktop/git/pre-commit (master)
$ virtualenv venv27 -ppython2.7
The path python2.7 (from --python=python2.7) does not exist
(venv)
Anthony@AnthonysDesktop MINGW64 ~/Desktop/git/pre-commit (master)
$ virtualenv venv27 -p python2.7
The path python2.7 (from --python=python2.7) does not exist
(venv)
Anthony@AnthonysDesktop MINGW64 ~/Desktop/git/pre-commit (master)
$ virtualenv venv27 -p python2.7.exe
Using real prefix 'c:\\python27'
New python executable in C:\Users\Anthony\Desktop\git\pre-commit\venv27\Scripts\python2.7.exe
Also creating executable in C:\Users\Anthony\Desktop\git\pre-commit\venv27\Scripts\python.exe
Installing setuptools, pip, wheel...done.
Running virtualenv with interpreter C:\Users\Anthony\Desktop\git\pre-commit\venv\Scripts\python2.7.exe
```

Basically, `virtualenv` has a bug here: https://github.com/pypa/virtualenv/blob/6d8a4ed81f40223a3f9ea1ba5d11be500d9f667a/virtualenv.py#L1542-L1565

evhub:
Yeah, I was rather surprised that `virtualenv` couldn't handle `--python python2.7` and needed `--python /path/to/python2.7.exe` instead.

asottile:
I'm going to try and send them a patch: https://github.com/pypa/virtualenv/issues/1062

asottile:
This is now available in v0.15.2

evhub:
@asottile Tested and working; thanks!

PR #565, opened 2017-07-19
Recover from invalid python virtualenvs

asottile:
This will help with travis-ci caches

It's efficient to cache ~/.pre-commit in travis-ci as it makes hook execution fast the next run.  When travis-ci upgrades their system pythons this often causes failures due to libpython going missing

Here's a blurb of what this breakage looks like:

```
Trim Trailing Whitespace.................................................Failed

hookid: trailing-whitespace

/home/travis/.pre-commit/repoxffdwnip/py_env-python3.6/bin/python3.6: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
```

I've only implemented `healthy` for python, the others can be written in the future.

### DIFF
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -21,6 +21,9 @@ from pre_commit.languages import system
 #     return 'default' if there is no better option.
 #    """
 #
+# def healthy(repo_cmd_runner, language_version):
+#     """Return whether or not the environment is considered functional."""
+#
 # def install_environment(repo_cmd_runner, version, additional_dependencies):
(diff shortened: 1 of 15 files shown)


Alain1405:
I'm currently getting, at commit:

```
```

asottile:
@Alain1405 you'll want to check on https://github.com/pre-commit/pre-commit/issues/577, it's likely `pre-commit autoupdate` will solve your problem (though it's not possible to know without seeing your `.pre-commit-config.yaml`)

PR #569, opened 2017-07-28
Making golang-based hooks compatible with Cygwin

Lucas-C:


asottile:
This'll probably need some `# pragma: no cover`s to satisfy coverage.

Quick question: this assumes if you're using cygwin python you're also using cygwin go -- what happens if they're mixed?

Lucas-C:
To my understanding, there is no Cygwin Go, only Windows Go, that only handles Windows paths.
So no versions mixing.

I added the missing `# pragma: no cover`

asottile:
Thanks! I'll get this in a release once I'm at a computer

PR #571, opened 2017-07-31
Use more git plumbing commands in staged-files-only

asottile:


asottile:
well this patch seems to not work well on windows!  I'll need to look into why though :)

asottile:
Here's the [mailing list entry](http://marc.info/?l=git&m=150152051316004&w=2) I sent to the git mailing list

PR #581, opened 2017-08-07
Update swift urls (the old ones 404d)

asottile:


asottile:
I've made an issue on the [swift project](https://bugs.swift.org/browse/SR-5637) as well in case this wasn't intentional

PR #583, opened 2017-08-11
When possible, preserve config format on autoupdate

asottile:


pre_commit/commands/autoupdate.py
```diff
@@ -65,6 +64,43 @@ def _update_repo(repo_config, runner, tags_only):
     return new_config
 
 
+SHA_LINE_RE = re.compile(r'^(\s+)sha:(\s*)([^\s#]+)(.*)$', re.DOTALL)
```
chriskuehl: why using DOTALL here?

asottile:
```
>>> line
'        sha: v1.0.0\n'
```
so it matches the `\n` line ending

PR #592, opened 2017-08-23
Remove --no-stash and --allow-unstaged-config

### LINKED ISSUE #587: Deprecate and remove some (useless?) options
I find the following don't really have any good use cases (and don't come up in normal day-to-day) and are undocumented beyond `--help`.  **I'm proposing removing these options**:

## `pre-commit run --no-stash`

This disables the auto-stashing of files when running -- though this is already the case for `pre-commit run --all-files` and `pre-commit run --files ...`.

The behaviour of `--no-stash` (without using `--no-stash`) can be achieved via `git diff --name-only | xargs pre-commit run --files`

It was added [along with the avoiding behaviour](https://github.com/pre-commit/pre-commit/pull/80) in the same pull request.  I want to say this was my first idea for "fixing" the [original problem](https://github.com/pre-commit/pre-commit/issues/68) and then I forgot to undo it.

## `pre-commit run --allow-unstaged-config`

This (unfortunately) collides with `pre-commit run --all` (prefix match of `pre-commit run --all-files`) so I've wanted to get rid of it anyway.

This allows one to run with an unstaged configuration, which then (in most cases) gets the changes swiped out from under you and causes confusing situations where the hooks that are run aren't what was on disk at the time of running.  The warning that's printed when doing this also explains this.

This was [originally my idea](https://github.com/pre-commit/pre-commit/issues/157#issuecomment-99080756) but now I think we can just do without the option at all -- requiring the pre-commit configuration to be staged when running pre-commit.

asottile:
Resolves #587

### DIFF
--- a/tests/git_test.py
+++ b/tests/git_test.py
@@ -137,8 +137,7 @@ def test_get_conflicted_files_in_submodule(in_conflicting_submodule):
 
 
 def test_get_conflicted_files_unstaged_files(in_merge_conflict):
-    # If they for whatever reason did pre-commit run --no-stash during a
-    # conflict
+    """This case no longer occurs, but it is a useful test nonetheless"""
     resolve_conflict()
 
     # Make unstaged file.
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -217,7 +217,7 @@ def _has_unstaged_config(runner):
 
 
 def run(runner, args, environ=os.environ):
-    no_stash = args.no_stash or args.all_files or bool(args.files)
+    no_stash = args.all_files or bool(args.files)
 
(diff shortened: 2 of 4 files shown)


tests/git_test.py
```diff
@@ -137,7 +137,7 @@ def test_get_conflicted_files_in_submodule(in_conflicting_submodule):
 
 
 def test_get_conflicted_files_unstaged_files(in_merge_conflict):
-    # If they for whatever reason did pre-commit run --no-stash during a
+    # If they for whatever reason did pre-commit run --files ... during a
```
asottile: actually, this case never occurs now but I think this test is still useful

Issue #363, opened 2016-04-13
Creating repositories fails in parallel


asottile:
Each process races to set up its virtualenv in the same place.

Probably needs more flock -- but how to make it work cross platform?

@chriskuehl  hits this a lot.
@ymilki just hit this as well

Workaround is to run once not-in-parallel and then it'll _just work_

Stacktraces vary, some complain of directories already existing `An unexpected error has occurred: OSError: [Errno 39] Directory not empty: '.../.pre-commit/repop35TZY/py_env-default'`

asottile:
Collecting some of the stacktraces I'm seeing while writing a fix for this:

```
Traceback (most recent call last):
  File "/home/asottile/workspace/pre-commit/pre_commit/error_handler.py", line 48, in error_handler
    yield
  File "/home/asottile/workspace/pre-commit/pre_commit/main.py", line 221, in main
    return run(runner, args)
  File "/home/asottile/workspace/pre-commit/pre_commit/commands/run.py", line 247, in run
    repo_hooks = list(get_repo_hooks(runner))
  File "/home/asottile/workspace/pre-commit/pre_commit/commands/run.py", line 197, in get_repo_hooks
    for repo in runner.repositories:
  File "/home/asottile/workspace/pre-commit/venv/lib/python3.6/site-packages/cached_property.py", line 26, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/asottile/workspace/pre-commit/pre_commit/runner.py", line 46, in repositories
    repository.require_installed()
  File "/home/asottile/workspace/pre-commit/pre_commit/repository.py", line 177, in require_installed
    _install_all(self._venvs, self.repo_config['repo'])
  File "/home/asottile/workspace/pre-commit/venv/lib/python3.6/site-packages/cached_property.py", line 26, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/asottile/workspace/pre-commit/pre_commit/repository.py", line 166, in _venvs
    for _, hook in self.hooks:
  File "/home/asottile/workspace/pre-commit/venv/lib/python3.6/site-packages/cached_property.py", line 26, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/asottile/workspace/pre-commit/pre_commit/repository.py", line 146, in hooks
    if hook['id'] not in self.manifest.hooks:
[...]

PR #594, opened 2017-08-24
Limit repository creation to one process

asottile:
This attempts to solve #363

I've tested this on windows and linux and it appears to work as intended.  This adds some additional printing of `[INFO] Locking pre-commit directory` when cloning / installing repositories.

I used this script to test this somewhat-thoroughly:

```bash
#!/usr/bin/env bash
set -euxo pipefail

tmpdir="$(mktemp -d)"
cleanup() {
    cat "$tmpdir/pre-commit.log" 2> /dev/null || true
    rm -rf "$tmpdir"
}
trap cleanup EXIT

export PRE_COMMIT_HOME="$tmpdir"
seq 10 |
    xargs -P10 --replace \
        pre-commit run trailing-whitespace --files README.md

echo 'done!'
```

Would definitely love a second set of eyes on this and someone to verify that this approach also works on macos

### DIFF
--- a/pre_commit/file_lock.py
+++ b/pre_commit/file_lock.py
@@ -1,0 +1,53 @@
+import contextlib
+import errno
+
+
+try:  # pragma: no cover (windows)
+    import msvcrt
+
+    # https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/locking
+
+    # on windows we lock "regions" of files, we don't care about the actual
+    # byte region so we'll just pick *some* number here.
+    _region = 0xffff
+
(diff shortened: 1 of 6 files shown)


asottile:
Verified this also works on macos

artcyb:
Give me tutorial

asottile:
@artcyb the documentation for pre-commit can be found at [pre-commit.com](http://pre-commit.com)

If you have a particular question not covered by the documentation, feel free to [open an issue](https://github.com/pre-commit/pre-commit/issues) on the project

PR #598, opened 2017-09-03
Add new docker_image language type.

### LINKED ISSUE #597: Support using existing Docker images as hooks
Currently a `language: docker` hook only supports running a Docker image which it built itself (via a `Dockerfile`).

However this seems a little unnecessary and expensive since we can and should be building and publishing our Docker images to a registry.

Would there be interest in supporting using existing images as hooks? It should be simple and I'd be happy to make the PR.

## Proposed `.pre-commit-hooks.yaml` change

```yaml
-   id: flake8-docker
    name: flake8 (via docker)
    description: flake8, but provided by docker
    files: \.py$

    language: docker
    # Image to `docker pull`
    docker_image: my.registry.example.com/pre-commit-hooks/flake8:latest
    # Optional: use `ENTRYPOINT` of the image if not specified
    entry: flake8
```

asottile:
`docker_image` is intended to be a lightweight hook type similar to system /
script which allows one to use an existing docker image which provides a
hook.

Resolves #597

### DIFF
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 from pre_commit.languages import docker
+from pre_commit.languages import docker_image
 from pre_commit.languages import golang
 from pre_commit.languages import node
 from pre_commit.languages import pcre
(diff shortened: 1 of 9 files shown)


asottile:
I'll follow up with documentation after merging!

PR #600, opened 2017-09-04
Only print that the lock is being acquired when waiting

### LINKED ISSUE #599: "Locking pre-commit directory" should only print if waiting for a lock
Otherwise this is just useless console noise

asottile:
Resolves #599


asottile:
This also fixes locking under python2 on windows

PR #604, opened 2017-09-05
Fix specify config file not work while installing

coldnight:
Via `pre-commit install -c .other-config.yaml`

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -71,6 +71,7 @@ def install(
             sys_executable=sys.executable,
             hook_type=hook_type,
             hook_specific=hook_specific_contents,
+            config_file=runner.config_file,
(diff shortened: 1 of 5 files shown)


coldnight:
@asottile Please review and I don't how to handle the hash things [here](https://github.com/pre-commit/pre-commit/blob/master/pre_commit/commands/install_uninstall.py#L14-L22).

asottile:
Looks great!  I haven't changed the hashes in forever and don't think they need to change -- iirc they're only if the calling construct of pre-commit changes.

Could you add a quick test that demonstrates that this works?  There should be some prior art in `tests/commands/install_uninstall_test.py`

asottile:
Oh neat, I was able to add a test to your branch directly.

asottile:
Thanks for the PR!  This will make it into the next release I have (I have some plans to add a few more things into this one as well)

PR #610, opened 2017-09-05
Enable map configurations (config v2).

### LINKED ISSUE #414: Configuration v2

asottile:
This resolves #414

I've also added the following to aid in migration:
- `pre-commit migrate-config`
    - attempts to intelligently upgrade the configuration to the new map format
- `pre-commit autoupdate` - by default calls `migrate_config(...)`

For now I have opted to _not_ issue a deprecation warning, I'll follow up with that in a later release.

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import argparse
+import collections
 import functools
 
 from aspy.yaml import ordered_load
@@ -125,7 +126,11 @@ CONFIG_REPO_DICT = schema.Map(
         ensure_absent=True,
     ),
 )
-CONFIG_SCHEMA = schema.Array(CONFIG_REPO_DICT)
+CONFIG_SCHEMA = schema.Map(
+    'Config', None,
+
+    schema.RequiredRecurse('repos', schema.Array(CONFIG_REPO_DICT)),
+)
 
 
 def is_local_repo(repo_entry):
@@ -136,10 +141,19 @@ class InvalidConfigError(FatalError):
     pass
 
 
+def ordered_load_normalize_legacy_config(contents):
+    data = ordered_load(contents)
+    if isinstance(data, list):
+        # TODO: Once happy, issue a deprecation warning and instructions
+        return collections.OrderedDict([('repos', data)])
+    else:
+        return data
+
+
 load_config = functools.partial(
     schema.load_from_filename,
     schema=CONFIG_SCHEMA,
-    load_strategy=ordered_load,
+    load_strategy=ordered_load_normalize_legacy_config,
     exc_tp=InvalidConfigError,
 )
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks.git
     sha: v0.9.1
     hooks:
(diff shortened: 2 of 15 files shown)


asottile:
This is mostly ready for review, I need to make `migrate-config` handle yaml that looks like this:

```yaml
---
-   repo: ...
    ...
```

And should probably handle this case better:

*current*
```diff
+repos:
 # this is a pre-commit config
 -   repo: ...
     ...
```
*better*
```diff
 # this is a pre-commit config
+repos:
 -   repo: ...
     ...
```

asottile:
Now handles both of those cases!

pre_commit/clientlib.py
```diff
@@ -125,7 +126,11 @@ def validate_manifest_main(argv=None):
         ensure_absent=True,
     ),
 )
-CONFIG_SCHEMA = schema.Array(CONFIG_REPO_DICT)
+CONFIG_SCHEMA = schema.Map(
+    'Config', None,
+
+    schema.RequiredRecurse('repos', schema.Array(CONFIG_REPO_DICT)),
```
chriskuehl: weird blank line?

asottile: this matches the other ones in the file, I plan to add more keys here so it'll look more natural _eventually_

PR #612, opened 2017-09-06
NodeJS hooks compatibilty fix for Cygwin - take 2

Lucas-C:


pre_commit/languages/node.py
```diff
@@ -17,11 +18,14 @@
 
 
 def get_env_patch(venv):  # pragma: windows no cover
-    config = os.path.join(venv, 'bin') if sys.platform == 'cygwin' else venv
+    install_prefix = venv
+    if sys.platform == 'cygwin':
+        _, win_venv, _ = cmd_output('cygpath', '-w', venv)
+        install_prefix = '{}\\bin'.format(win_venv.strip())
```
asottile: Cool!

I'd personally write this as if + else and use a raw string.  This works as is

asottile: Ah you'll need the no cover comment here, want to make the other changes while you're at it too?

Lucas-C: Sure, but what do you mean by "raw string" ? `r'{}\bin'.format(...)` ?

asottile: Yep!

asottile: Ah, I can see how it's confusing here. `windows no cover` is "this won't be covered under test on appveyor" while `no cover` is for Travis and appveyor. The existing comment should stay and you should add a plain `no cover` to your new code. It didn't need it previously because ternaries aren't considered branches.

Lucas-C: Oh ok ! Thanks for the explanation
This should now be fixed

Lucas-C:
Thanks ;)

asottile:
@Lucas-C no, thank you! :D

PR #616, opened 2017-09-08
Implement `fail_fast`.

### LINKED ISSUE #240: Feature request : add a "fail fast" configuration entry
Here is my use case:
- I have integration tests (in Java + with protractor) that are costly to execute (in term of execution time)
- when earlier hooks detect issues with the code (e.g. `eslint` checks), I don't want to execute my integration tests until I've fixed those errors

Btw, I don't ask to change the current behaviour, but could you develop on the reasons behind the design decision to run all tests, indifferently of their outcome ? What are the benefits ?

asottile:
Resolves #240 CC @Lucas-C

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -130,6 +130,7 @@ CONFIG_SCHEMA = schema.Map(
     'Config', None,
 
     schema.RequiredRecurse('repos', schema.Array(CONFIG_REPO_DICT)),
+    schema.Optional('fail_fast', schema.check_bool, False),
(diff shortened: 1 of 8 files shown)


asottile:
Ah now that there's default things I need to adjust autoupdate to strip them again

asottile:
I'll revert this and use it : https://github.com/pre-commit/pre-commit/commit/ee392275f308032dc47ec0dea9d19c92b89d5996#diff-83205468a6e0cd4a6f7722858822325f

Issue #625, opened 2017-09-21
Apostrophe in syspath causes error


seanon414:
Runtime
```
$ pre-commit
An unexpected error has occurred: ValueError: No closing quotation
Check the log at /Users/sean.o'neill/.cache/pre-commit/pre-commit.log
```

Stacktrace
```
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shlex.py", line 279, in split
    return list(lex)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shlex.py", line 269, in next
    token = self.get_token()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shlex.py", line 96, in get_token
    raw = self.read_token()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shlex.py", line 172, in read_token
    raise ValueError, "No closing quotation"
ValueError: No closing quotation
```

Username in path has an apostrophe in it (not by choice) and causes pre-commit to bomb

Computer info
```
OS: Mac El Caption 10.11.6
Version: pre-commit 1.1.2
```

asottile:
Through some guesswork, I've been able to reproduce this.  Here's a full trace:

```
An unexpected error has occurred: ValueError: No closing quotation
Traceback (most recent call last):
  File "/tmp/foo'bar/venv/local/lib/python2.7/site-packages/pre_commit/error_handler.py", line 44, in error_handler
    yield
  File "/tmp/foo'bar/venv/local/lib/python2.7/site-packages/pre_commit/main.py", line 231, in main
    return run(runner, args)
  File "/tmp/foo'bar/venv/local/lib/python2.7/site-packages/pre_commit/commands/run.py", line 267, in run
    return _run_hooks(runner.config, repo_hooks, args, environ)
  File "/tmp/foo'bar/venv/local/lib/python2.7/site-packages/pre_commit/commands/run.py", line 189, in _run_hooks
    retval |= _run_single_hook(filenames, hook, repo, args, skips, cols)
  File "/tmp/foo'bar/venv/local/lib/python2.7/site-packages/pre_commit/commands/run.py", line 103, in _run_single_hook
    tuple(filenames) if hook['pass_filenames'] else (),
  File "/tmp/foo'bar/venv/local/lib/python2.7/site-packages/pre_commit/repository.py", line 198, in run_hook
[...]

PR #626, opened 2017-09-21
Use pipes.quote for executable path

asottile:


asottile:
Actually noticed this while poking around at https://github.com/pre-commit/pre-commit/issues/625

PR #630, opened 2017-09-23
Implement pygrep language as a replacement for pcre

### LINKED ISSUE #404: Expose `negate` and deprecate pcre hooks
See https://github.com/pre-commit/pre-commit/blob/b05cc4077e621cec637b22d6ad2b80ee84749f97/pre_commit/xargs.py#L48

I imagine something like this:

``` yaml
- id: test
  language: system
  negate_returncode: true
  entry: grep -P herp
```

Instead of the equivalent pcre hook

asottile:
This is more cross-platform compatible than pcre.

Resolves #404


asottile:
CC @chriskuehl @Lucas-C

Issue #405, opened 2016-08-31
Warn on useless `exclude:`s during `--all-files`


asottile:
Some of our repositories have some copy-paste pre-commit configs which end up with problems like https://github.com/Yelp/undebt/pull/12#discussion-diff-76108499

It'd be nice to surface these issues to the end-user in a nice way

chriskuehl:
+1 this would be great, though I imagine we can only alert if the regex doesn't match anything at all.

Meaning a common thing like:

```yaml
exclude: ^(file_that_exists|file_that_does_not_exist)\.py$
```

wouldn't really benefit from this. Unless you have some magic way to interpret regexes.

asottile:
I've always wanted to implement a "regex coverage" routine, but I don't think this is the time :) (especially since python regexes are complicated).  I was mainly aiming for the terribly-wrong-doesn't-cover-anything case

chriskuehl:
We'd likely also want to differentiate between `exclude` being defined in the hook repo and in the project repo (as an override). For example, the `replace-tabs` hook does something like `exclude: (Makefile|\.mk)$` which is useful, but we wouldn't want that to error in every project.

asottile:
via https://github.com/pre-commit/pre-commit/pull/643

Issue #582, opened 2017-08-08
Sign-off git commit if all pre-commit checks pass


ylogx:
It would be great if there was an option to add a sign at the end of commit message, that this commit was verified by pre-commit.

A dummy git commit message maybe:

```
Do some amazing thing to database connection

This commit will do amazing magical things.
More such things

NOTE:
* It will also make the moon brighter
* It will make the sun shine better

Signed-off-by: Shubham Chaudhary <shubham@chaudhary.xyz>
Signed-off-by: pre-commit verified with 8 hooks
```

asottile:
This sounds like a really cool idea -- and I think the latest updates which add `commit-msg` hooks could support this idea.

The concerns I have about this:
- What about `--no-verify`?
    - Neither `.git/hooks/pre-commit` nor `.git/hooks/commit-msg` run when `--no-verify` is passed, so nothing happens
- What about `SKIP`?
    - This could be detected by a cli and adjust the number of hooks
- What about hooks that don't affect any files?
    - I'd still consider these hooks to have "run", but trivially -- it'd be quite difficult to figure out which ones were applicable without replicating the entire logic of pre-commit in a hook
- How would a client hook know if all the hooks passed?
    - Simply put, it wouldn't but I don't think that's actually a problem -- in the case of `pre-commit` hooks failing, it would never get to the `commit-msg` hook which does the modification.
    - When another `commit-msg` hook fails, even if the hook has done the modification to the commit message the commit would still get properly cancelled

If you don't mind, I'm going to create another repository and start hacking away on this idea -- I think this is a good one :)

asottile:
For a first pass, I'm just going to hardcode a message -- we can work on making it fancier in the future if that's desirable :)

shubhamchaudhary:
+1 Makes sense.

asottile:
Just an update on this, I _am_ still working on it though I've put this more on the back burner.  Due to the tight coupling between pre-commit and this hook I think it makes more sense to keep this internal to the framework.  So now I'm mostly stuck on how to enable such a thing -- I have a hunch I'll want to do something with #414

asottile:
A framework for adding something like this was contributed as part of https://github.com/pre-commit/pre-commit/pull/643

MaxymVlasov:
JFYI, it is already done as pre-commit hook
https://github.com/mattlqx/pre-commit-sign

adam-moss:
[...]

PR #638, opened 2017-10-20
Warn when exclude patterns are not matched

hackedd:
This is a first pass on implementing #405. It works, but because I could not find a way to check where the exclude pattern was defined, the false-positives described by [@chriskuehl](https://github.com/pre-commit/pre-commit/issues/405#issuecomment-244613712) are not handled. The code also specifically checks for `'^$'` (the default exclude pattern) which is not great.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -38,14 +38,15 @@ def _hook_msg_start(hook, verbose):
 
 def _filter_by_include_exclude(filenames, include, exclude):
     include_re, exclude_re = re.compile(include), re.compile(exclude)
-    return {
-        filename for filename in filenames
-        if (
-            include_re.search(filename) and
-            not exclude_re.search(filename) and
-            os.path.lexists(filename)
-        )
-    }
+    filtered = set()
+    exclude_matched = False
+    for filename in filenames:
+        if include_re.search(filename):
+            if exclude_re.search(filename):
+                exclude_matched = True
+            elif os.path.lexists(filename):
+                filtered.add(filename)
+    return filtered, exclude_matched
 
 
 def _filter_by_types(filenames, types, exclude_types):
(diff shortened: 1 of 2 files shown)


asottile:
First: thanks for the PR!  This is awesome!

Second: I dropped the ball here a bit on this issue, I actually have a plan for this (that perhaps you could champion?) that helps solve this in a more generic way.  I probably should have communicated my plan in the three issues I hope to tackle in similar ways such that it was more obvious 😆 -- thought I'd get to it before anyone else though (guess I'll do something else this weekend!).

Anyway, the rough sketch for how I think this should look is as follows (and applies equally to https://github.com/pre-commit/pre-commit/issues/405, https://github.com/pre-commit/pre-commit/issues/582, and https://github.com/pre-commit/pre-commit-hooks/issues/63):

Add a new `Repository` type (similar to `repo: local`) called `meta` (name pending, if you've got a better name I'm all ears :D).  Similar to `repo: local` this would be included in the configuration by a special name and would not require a `sha` key:

```yaml
repos:
-    repo: meta
     hooks: ...
```

This `meta` repo would contain hooks which primarily function on pre-commit itself, I think in the 99% case they would be "linting" `.pre-commit-config.yaml`.  The rationale behind making these a "first class" repository type is while they _could_ be implemented as third party hooks, they're really skittering on the edge of private implementation detail (tight coupling) and would probably do better inside pre-commit itself.

pre-commit would grow a package which has some executables for these "meta" hooks, probably something like `pre_commit.meta_hooks.*`.  Each meta hook would define an executable and be able to use tightly coupled implementation detail.

The [`Repository`](https://github.com/pre-commit/pre-commit/blob/2a984c37463da9f712cd89c5b4c1c9889d3b200f/pre_commit/repository.py#L121) would probably be implemented to have a static manifest defined inline.  Perhaps something like this:

```python
manifest = [
    {
        'id': 'check-useless-excludes',
        'name': 'Check for useless excludes',
        'files': '.pre-commit-config.yaml',
        'language': 'system',
        'entry': '{} -m pre_commit.meta_hooks.check_useless_excludes'.format(sys.executable),
    },
    # ...
]
```

Does this make sense?  Then consumers could opt into this as a hard failure.

hackedd:
I think the meta hook idea makes sense, I'll see if I can get something like that working.

asottile:
superseded by #643

PR #641, opened 2017-10-20
Add PoC for on-demand hooks repo download

### LINKED ISSUE #637: Too much data pre-downloaded when not needed
Look here:
https://travis-ci.org/cherrypy/cherrypy/jobs/290358917#L698-L713

I run only `pep257` hook, but pre-commit downloads all other repos, while it is possible to detect that just one repo is used in the command I run.

I suggest improving logic so that it would only touch hook repos if really needed.
i.e. clone only `git://github.com/FalconSocial/pre-commit-mirrors-pep257` in my case.

webknjaz:
This makes pre-commit skip all config entries not listing certain hook

Resolves #637

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -57,6 +57,8 @@ def _add_hook_type_option(parser):
 def _add_run_options(parser):
     parser.add_argument('hook', nargs='?', help='A single hook-id to run')
     parser.add_argument('--verbose', '-v', action='store_true', default=False)
+    parser.add_argument('--only', '-i', action='store_true', default=False,
+                        help='Save some bandwidth')
     parser.add_argument(
         '--origin', '-o',
         help="The origin branch's commit_id when using `git push`.",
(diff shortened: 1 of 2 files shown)


asottile:
probably won't merge something like this but I'll keep it in mind if I do some big refactors -- I think this adds some coupling that I'm not terribly interested in maintaining for a feature that I won't use and most won't use.  Good idea though and if there's some clean way to avoid the coupling that accomplishes this I'll definitely incorporate it!  Thanks again for the patch!

webknjaz:
> this adds some coupling

Yeah, that's why I put `PoC` in title.

I'd also like to have smth cleaner and won't accept such dirty code. I'll think if I can make this better, though.

asottile:
since we're not going to merge this as is, I'm going to close this for now -- we may revisit making pre-commit more lazy in this regard in the future :)

Thanks again for the proposed patch!

PR #642, opened 2017-10-23
Add selinux labelling option to docker_image hook type

jimmidyson:
Without this, the hook fails on envs running with selinux enabled. This option automatically labels mounted volumes correctly.

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -82,7 +82,7 @@ def docker_cmd():
         'docker', 'run',
         '--rm',
         '-u', '{}:{}'.format(os.getuid(), os.getgid()),
-        '-v', '{}:/src:rw'.format(os.getcwd()),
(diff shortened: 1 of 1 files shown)


asottile:
Thanks for the PR!

Could you paste a failure mode (just as a comment is fine) such that others can hopefully find this issue as well? Perhaps also a code comment which explains what the `Z` flag does would be useful as well.

From [the docs](https://docs.docker.com/engine/reference/commandline/run/#mount-volumes-from-container-volumes-from):

> The `Z` option tells Docker to label the content with a private unshared label. Only the current container can use a private volume.

asottile:
Thanks again for the PR!

jimmidyson:
Oh thank you! Sorry I'd missed your comment on this PR in the deluge of other notifications... Thank you for adding the comment.

asottile:

PR #643, opened 2017-10-25
Implement support for meta hooks

hackedd:
As discussed in [#638](https://github.com/pre-commit/pre-commit/pull/638#issuecomment-338256399), I've implemented support for meta hooks.

Two hooks are currently implementing, one to check if every hook with a `files` pattern actually matches any files (`check-files-matches-any`, pre-commit/pre-commit-hooks#63), the other to check if every `exclude` pattern actually matches any files (`check-useless-excludes`, #405).

They can be activated by adding this to your `.pre-commit-config.yaml`:

```yaml
-   repo: meta
    hooks:
    -   id: check-useless-excludes
    -   id: check-files-matches-any
```


pre_commit/schema.py
```diff
@@ -206,6 +209,14 @@ def __eq__(self, other):
         return other is not MISSING and other != self.val
 
 
+class NotIn(object):
+    def __init__(self, values):
```
asottile: I'd probably do `__init__(self, *values)` just to save some parens (other classes in this module do similarly).

p4: bonus points if it's a namedtuple (so we get free `repr`):

```python
class NotIn(collections.namedtuple('NotIn', ('values',))):
    def __new__(cls, *values):
        return super(NotIn, cls).__new__(cls, values=values)
    def __eq__(...):
        ...
```

pre_commit/clientlib.py
```diff
@@ -121,7 +123,8 @@ def validate_manifest_main(argv=None):
 
     schema.Conditional(
         'sha', schema.check_string,
-        condition_key='repo', condition_value=schema.Not(_LOCAL_SENTINEL),
+        condition_key='repo',
+        condition_value=schema.NotIn((_LOCAL_SENTINEL, _META_SENTINEL)),
```
asottile: ++, this is exactly how I imagined this would work

pre_commit/meta_hooks/check_files_matches_any.py
```diff
@@ -0,0 +1,36 @@
+import re
+import sys
+
+import pre_commit.constants as C
+from pre_commit.clientlib import load_config
+from pre_commit.git import get_all_files
+
+
+def files_matches_any(filenames, include):
```
asottile: let's try and reuse the `include_exclude` from `pre_commit.commands.run` (such that this more reflects reality and doesn't drift)

pre_commit/meta_hooks/check_files_matches_any.py
```diff
@@ -0,0 +1,36 @@
...
+    include_re = re.compile(include)
+    for filename in filenames:
+        if include_re.search(filename):
+            return True
+    return False
+
+
+def check_files_matches_any(config_file=None):
+    config = load_config(config_file or C.CONFIG_FILE)
```
asottile: to be a valid hook, this needs to take `1..*` arguments and process them.  Let's add an argument parser similar to [this one](https://github.com/pre-commit/pre-commit-hooks/blob/e718847ccbc65cd72a89fefb4db61c55ec4d430d/pre_commit_hooks/check_vcs_permalinks.py#L27-L40)

pre_commit/meta_hooks/check_files_matches_any.py
```diff
@@ -0,0 +1,36 @@
...
+    files = get_all_files()
+    files_not_matched = False
+
+    for repo in config['repos']:
+        for hook in repo['hooks']:
+            include = hook.get('files', '')
```
asottile: `files` will always be present when loaded through `load_config` due to `pre_commit.schema` -- this can be `hook['files']`

pre_commit/meta_hooks/check_files_matches_any.py
```diff
@@ -0,0 +1,36 @@
...
+                    'The files pattern for {} does not match any files'
+                    .format(hook['id'])
+                )
+                files_not_matched = True
+
+    return files_not_matched
+
+
+if __name__ == '__main__':
+    sys.exit(check_files_matches_any())
```
asottile: just a consistency thing: the rest of the modules use `exit` (introduced as a convenience in python2.5 as part of the `site` module)  this is equivalent and fine, just a slight inconsistency

pre_commit/meta_hooks/check_files_matches_any.py
asottile: I've also been moving more to conventionally naming this function `main` -- the function itself makes sense given the context that the module name gives us

pre_commit/meta_hooks/check_files_matches_any.py
```diff
@@ -0,0 +1,36 @@
+import re
```
asottile: My thoughts with this hook (again, I didn't communicate this, sorry!) is to make `(no files to check)Skipped` an *error* when running with `--all-files`.  So what I'd like to see in this hook is a reuse of the `include` / `exclude` for both files and types from `pre_commit.commands.run` -- the messaging can be `"{hook_id} does not apply to this repository"` or something of the sort

pre_commit/meta_hooks/check_useless_excludes.py
```diff
@@ -0,0 +1,41 @@
...
+def exclude_matches_any(filenames, include, exclude):
+    include_re, exclude_re = re.compile(include), re.compile(exclude)
+    for filename in filenames:
+        if include_re.search(filename) and exclude_re.search(filename):
+            return True
+    return False
+
+
+def check_useless_excludes(config_file=None):
+    config = load_config(config_file or C.CONFIG_FILE)
```
asottile: instead of loading the configuration, I think you can just use `ordered_load` -- then you don't have the issue of picking up the values from the manifest -- you'll of course then need to do tests for keys and can't assume they're always there -- but I think you'll be able to make stronger assertions about correctness and eliminate false positives.

pre_commit/meta_hooks/check_useless_excludes.py
```diff
@@ -0,0 +1,41 @@
...
+    files = get_all_files()
+    useless_excludes = False
+
+    exclude = config.get('exclude')
+    if exclude != '^$' and not exclude_matches_any(files, '', exclude):
+        print('The global exclude pattern does not match any files')
```
asottile: Let's include `'{}'` in the error messages so we know *what* pattern failed to match

pre_commit/meta_hooks/check_useless_excludes.py
```diff
@@ -0,0 +1,41 @@
...
+        useless_excludes = True
+
+    for repo in config['repos']:
+        for hook in repo['hooks']:
+            include, exclude = hook.get('files', ''), hook.get('exclude')
```
asottile: hmmm, unless I'm thinking about this wrong I don't think `files` is relevant to this hook?

pre_commit/repository.py
```diff
@@ -242,6 +247,55 @@ def _venvs(self):
...
+    # Note: the hook `entry` is passed through `shlex.split()` by the command
+    # runner, so to prevent issues with spaces and backslashes (on Windows) it
+    # must be quoted here.
+    meta_hooks = {
+        'check-useless-excludes': {
+            'name': 'Check for useless excludes',
+            'files': '.pre-commit-config.yaml',
+            'language': 'system',
+            'entry': pipes.quote(sys.executable),
+            'args': ['-m', 'pre_commit.meta_hooks.check_useless_excludes'],
```
asottile: one benefit of `entry` is it gets `shlex`'d for us.  so we can do this:

```python
'entry': '{} -m {}'.format(pipes.quote(sys.executable), check_useless_excludes.__name__),
```

instead of (error prone) hardcoding the module names, we can import them and use `__name__` as well

pre_commit/repository.py
```diff
@@ -242,6 +247,55 @@ def _venvs(self):
...
+    }
+
+    @cached_property
+    def hooks(self):
+        for hook in self.repo_config['hooks']:
+            if hook['id'] not in self.meta_hooks:
+                logger.error(
+                    '`{}` is not a valid meta hook.  '
+                    'Typo? Perhaps it is introduced in a newer version?  '
+                    'Often `pre-commit autoupdate` fixes this.'.format(
```
asottile: mmm `autoupdate` won't actually help out here -- but `pip install pre-commit --upgrade` might!

asottile:
I made a helper `_hook` in #646 that should help the dict-merging + validation in `Repository`s easier

pre_commit/meta_hooks/check_files_matches_any.py
hackedd: `load_config` does not merge the data from `.pre-commit-config.yaml` with the configuration in the hook repo, that happens in `Repository.hooks`.

pre_commit/meta_hooks/check_useless_excludes.py
hackedd: This way there will be an error when you're excluding a file that exists, but is not included by `files`.

For example, you could have a hook that only runs on Python files, but has an `exclude` pattern for the README. Without looking at `files`, `check-useless-excludes` would conclude that everything is fine, although in reality the `exclude` is never matched.

pre_commit/repository.py
hackedd: +1 for using the module's `__name__`, I didn't think of that.

hackedd: Annoyingly, this lead to a cyclic import if, for example, the hook tries to import `Repository` or `Runner`.

pre_commit/meta_hooks/check_files_matches_any.py
asottile: oh true!  totally forgot about this 👍

pre_commit/meta_hooks/check_useless_excludes.py
asottile: oh nice, great idea!

pre_commit/repository.py
asottile: hmmmm womp.

Perhaps make a

```python
@cached_property
def manifest_hooks(self):
    from pre_commit.meta_hooks import check_useless_excludes
    ...
```

to avoid the circular imports? _(circular imports are such a bummer)_

asottile:
I also merged master into your branch to resolve the tiny conflict (and get you the new functionality) -- hope that's ok!

asottile:

PR #646, opened 2017-10-26
Apply interpreter version defaulting to local hooks too

asottile:


tests/manifest_test.py
```diff
@@ -58,13 +58,3 @@ def test_hooks(manifest):
...
-
-
-def test_default_python_language_version(store, tempdir_factory):
-    path = make_repo(tempdir_factory, 'python_hooks_repo')
-    repo_path = store.clone(path, git.head_sha(path))
-    manifest = Manifest(repo_path)
-
-    # This assertion is difficult as it is version dependent, just assert
-    # that it is *something*
-    assert manifest.hooks['foo']['language_version'] != 'default'
```
asottile: note that I removed this test, but it's already asserted [here](https://github.com/pre-commit/pre-commit/blob/39d5205e3112cc4afd4c0c1e861f54785b132bd9/tests/repository_test.py#L437)

PR #649, opened 2017-10-27
change the HOME for git commands

gaborbernat:
this avoids using the users configuration, which can contain e.g. gpg signing with password which would break this tests

### DIFF
--- a/testing/util.py
+++ b/testing/util.py
@@ -42,6 +42,16 @@ skipif_cant_run_swift = pytest.mark.skipif(
     reason='swift isn\'t installed or can\'t be found',
 )
 
+skipif_cant_run_go = pytest.mark.skipif(
+    parse_shebang.find_executable('go') is None,
+    reason='go isn\'t installed or can\'t be found',
+)
+
+skipif_cant_run_bash = pytest.mark.skipif(
+    parse_shebang.find_executable('bash') is None,
+    reason='bash isn\'t installed or can\'t be found',
+)
+
 xfailif_windows_no_ruby = pytest.mark.xfail(
     os.name == 'nt',
     reason='Ruby support not yet implemented on windows.',
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -39,6 +39,11 @@ def tempdir_factory(tmpdir):
     yield TmpdirFactory()
 
 
+@pytest.fixture()
+def setup_git(monkeypatch, tmpdir):
+    # we should not use user configuration
+    monkeypatch.setenv('HOME', tmpdir)
+
 @pytest.yield_fixture
 def in_tmpdir(tempdir_factory):
     path = tempdir_factory.get()
@@ -67,7 +72,7 @@ def _make_conflict():
 
 
 @pytest.yield_fixture
-def in_merge_conflict(tempdir_factory):
+def in_merge_conflict(tempdir_factory, setup_git):
     path = make_consuming_repo(tempdir_factory, 'script_hooks_repo')
     with cwd(path):
         open('dummy', 'a').close()
@@ -82,7 +87,7 @@ def in_merge_conflict(tempdir_factory):
 
 
 @pytest.yield_fixture
-def in_conflicting_submodule(tempdir_factory):
+def in_conflicting_submodule(tempdir_factory, setup_git):
     git_dir_1 = git_dir(tempdir_factory)
     git_dir_2 = git_dir(tempdir_factory)
     with cwd(git_dir_2):
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -41,50 +41,49 @@ def _norm_out(b):
     return b.replace(b'\r\n', b'\n')
 
 
-def _test_hook_repo(
-        tempdir_factory,
-        store,
-        repo_path,
-        hook_id,
-        args,
-        expected,
-        expected_return_code=0,
-        config_kwargs=None,
-):
-    path = make_repo(tempdir_factory, repo_path)
-    config = make_config_from_repo(path, **(config_kwargs or {}))
-    repo = Repository.create(config, store)
-    hook_dict, = [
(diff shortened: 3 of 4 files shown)


tests/conftest.py
```diff
@@ -39,6 +39,11 @@ def get(self):
     yield TmpdirFactory()
 
 
+@pytest.fixture()
+def setup_git(monkeypatch, tmpdir):
+    # we should not use user configuration
+    monkeypatch.setenv('HOME', tmpdir)
```
asottile: we had the same situation in pre-commit-hooks and I think this is the wrong solution as it has the potential to *hide* problems with different configurations.  We instead opted to pass `--no-gpg-sign` to any `git commit`s under test: https://github.com/pre-commit/pre-commit-hooks/pull/121

asottile: this fixture also doesn't scale (adding it to *literally every test*) -- perhaps instead an `autouse` fixture would have been more appropriate

tests/repository_test.py
```diff
@@ -41,50 +43,49 @@ def _norm_out(b):
...
-    repo = Repository.create(config, store)
-    hook_dict, = [
-        hook for repo_hook_id, hook in repo.hooks if repo_hook_id == hook_id
-    ]
-    ret = repo.run_hook(hook_dict, args)
-    assert ret[0] == expected_return_code
-    assert _norm_out(ret[1]) == expected
+@pytest.fixture(name='test_hook_repo')
+def test_hook_repo_fixture(tempdir_factory, store, setup_git):
+    def do(repo_path,
```
asottile: why this change?

testing/util.py
```diff
@@ -42,6 +42,16 @@ def cmd_output_mocked_pre_commit_home(*args, **kwargs):
...
 
+skipif_cant_run_go = pytest.mark.skipif(
+    parse_shebang.find_executable('go') is None,
+    reason='go isn\'t installed or can\'t be found',
+)
+
+skipif_cant_run_bash = pytest.mark.skipif(
+    parse_shebang.find_executable('bash') is None,
+    reason='bash isn\'t installed or can\'t be found',
+)
```
asottile: without bash, most of the testsuite is useless -- pre-commit tests lots of the framework functionality with script hooks as they are _fast_.  I don't think I can accept a PR which adds skips for `bash`.  As `bash` is necessarily installed on all the platforms we support (it's necessary for `git` itself to function) this also seems inappropriate.

tests/repository_test.py
```diff
@@ -616,7 +616,10 @@ def test_invalidated_virtualenv(tempdir_factory, store):
     paths = [
         os.path.join(libdir, p) for p in ('site.py', 'site.pyc', '__pycache__')
     ]
-    cmd_output('rm', '-rf', *paths)
+
+    for path in paths:
+        if os.path.exists(path):
+            os.rmdir(path)
```
asottile: 👍

gaborbernat:
@asottile I did not knew bash is required. I'm personally not comfortable knowing  this to add this project to tox; as you make it at best cumbersome for people on Windows to develop; now they need to install some kind of bash beside just Python.

asottile:
It's not required for runtime, just very important for the test suite. Also if you're running git on windows, you're running bash ;)

gaborbernat:
@asottile I'm using cmder which includes bash, however the test fails as it tries to find ``/usr/bin/bash`` which is not here; any ideas?

asottile:
Can you produce the output of the test as well as `which bash` and `echo $PATH`?

gaborbernat:
```
C:\Users\berna\git\tox (speed)
λ env | grep PATH
GIT_LFS_PATH=C:\Program Files\Git LFS
HOMEPATH=\Users\berna
PATH=/c/Program Files/cmder/bin:/c/Program Files/cmder/vendor/conemu-maximus5/ConEmu/Scripts:/c/Program Files/cmder/vendor/conemu-maximus5:/c/Program Files/cmder/vendor/conemu-maximus5/ConEmu:/c/ProgramData/Oracle/Java/javapath:/c/Program Files (x86)/Intel/iCLS Client:/c/Program Files/Intel/iCLS Client:/c/Windows/system32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/c/Program Files (x86)/Intel/Intel(R) Management Engine Components/DAL:/c/Program Files/Intel/Intel(R) Management Engine Components/DAL:/c/Program Files (x86)/Intel/Intel(R) Management Engine Components/IPT:/c/Program Files/Intel/Intel(R) Management Engine Components/IPT:/c/Program Files (x86)/NVIDIA Corporation/PhysX/Common:/cmd:/c/WINDOWS/system32:/c/WINDOWS:/c/WINDOWS/System32/Wbem:/c/WINDOWS/System32/WindowsPowerShell/v1.0:/c/Program Files/Git LFS:/c/Program Files/PuTTY:/c/Program Files/nodejs:/c/HashiCorp/Vagrant/bin:/:/c/ProgramData/Anaconda3:/c/ProgramData/Anaconda3/Scripts:/c/Users/berna/AppData/Local/conda/conda/envs:/c/Users/berna/AppData/Local/Microsoft/WindowsApps:/c/Program Files (x86)/Microsoft VS Code/bin:/c/Users/berna/AppData/Roaming/npm:/c/Users/berna/AppData/Local/Google/Cloud SDK/google-cloud-sdk/bin:/usr/bin:/usr/share/vim/vim74:/c/Program Files/cmder
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
VBOX_MSI_INSTALL_PATH=C:\Program Files\Oracle\VirtualBox\

C:\Users\berna\git\tox (speed)
λ which bash
/c/Windows/system32/bash
```

tests/conftest.py
gaborbernat: I see

PR #653, opened 2017-11-03
Rename check-files-matches-any to check-hooks-apply

asottile:
CC @hackedd -- I decided to change the name of this meta hook before I release it :)


hackedd:
Makes sense, check-files-matches-any was not a very good name.

Issue #656, opened 2017-11-06
autoupdate just one hook-id


KevinHock:
Hi Anthony et al.,

I was going to make a PR such that you can run `pre-commit autoupdate just-one-hook-id` and only update that one hook.

Let me know if you think this is a bad idea for some unforeseen reason on my part.

Cheers,
Kevin

asottile:
since revisions are pinned on a repository level, perhaps a repository would be a better level to work against?

You *could* do based on a hook-id, but this would have the side-effect of upgrading all hooks in the same repository

asottile:
But yeah, sounds like a good idea to me, perhaps something like this:

```
pre-commit autoupdate --repo https://github.com/pre-commit/pre-commit
```

KevinHock:
Good point, I'll do it that way, thanks @asottile 👍

asottile:
via #657

PR #657, opened 2017-11-07
Add repo option to autoupdate

KevinHock:
Aims to add the feature brought up in https://github.com/pre-commit/pre-commit/issues/656

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -168,10 +168,7 @@ def main(argv=None):
         ),
     )
     autoupdate_parser.add_argument(
-        '--repo', nargs=1, default=None,
-        help=(
-            'Repository to update the hooks of.'
-        ),
+        '--repo', help='Only update this repository.',
     )
 
     migrate_config_parser = subparsers.add_parser(
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -167,6 +167,12 @@ def main(argv=None):
             'tagged version (the default behavior).'
         ),
     )
+    autoupdate_parser.add_argument(
+        '--repo', nargs=1, default=None,
+        help=(
+            'Repository to update the hooks of.'
+        ),
+    )
 
     migrate_config_parser = subparsers.add_parser(
         'migrate-config',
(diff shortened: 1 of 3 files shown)


tests/commands/autoupdate_test.py
```diff
@@ -114,8 +114,11 @@ def test_autoupdate_out_of_date_repo(
     )
     write_config('.', config)
 
+    runner = Runner('.', C.CONFIG_FILE)
     before = open(C.CONFIG_FILE).read()
-    ret = autoupdate(Runner('.', C.CONFIG_FILE), tags_only=False)
+    repo_name = 'file://{}'.format(out_of_date_repo.path)
+    # It will update the repo, because the name matches
+    ret = autoupdate(runner, tags_only=False, repo=repo_name)
```
asottile: let's leave this test with the default `repo=None` and just add new tests

pre_commit/main.py
```diff
@@ -167,6 +167,12 @@ def main(argv=None):
             'tagged version (the default behavior).'
         ),
     )
+    autoupdate_parser.add_argument(
+        '--repo', nargs=1, default=None,
+        help=(
+            'Repository to update the hooks of.'
+        ),
+    )
```
asottile: the default for `nargs` is `1` and the default for `default` is None (so no need to specify them

but I guess that brings up the question: should I be able to do:

```
pre-commit autoupdate --repo repo1 --repo repo2 --repo repo3
```

to update multiple repositories?

That can be implemented pretty easily with `action='append'` -- thoughts?

KevinHock: Since it doesn't match my use-case I'm not very motivated to, to be honest. We can make an issue for future improvement though, `good-first-issue` or something.

tests/commands/autoupdate_test.py
KevinHock: 👍

KevinHock:
Sure thing, happy to contribute :)

pre_commit/main.py
asottile: good enough for me!

KevinHock:
Great working with you, you should checkout https://github.com/python-security/pyt if you like contributing to OSS ;)

asottile:

Issue #666, opened 2017-11-30
Question: listing pre-push files


orenmazor:
I'm running a few hooks on the pre-push hook, namely flake8. unfortunately, if I'm pushing to a remote that doesnt exist yet, flake8 decides to run across _all_files_, which is unfortunate and time consuming.

I don't see any option in the code of run.py that would let me only run only on the files about to be pushed. namely, the output of `git diff --stat --cached origin/master --name-only` (or something like that that just lists the files adjusted in this new branch that is not yet a remote branch).

more than happy to PR this, but I feel like this is something that's gotta already exist?

orenmazor:
I'm thinking something like this (just tested it locally)

https://github.com/orenmazor/pre-commit/commit/ac07e957d78bf4960620a0cf997408bf19e3fb16?diff=unified

basically switching the default from `all-files` to `what changed between master`. all files just takes too long for us

perobertson:
The other time it runs all files is after a rebase and force push.

asottile:
This does actually seem like something that should be improved, I'll keep my thoughts here so it's not entirely linked to a specific PR.

pre-commit at the core is a developer tool, it's aim is to help developers eliminate quick mistakes and help them, not get in the way.  When used as a pre-commit/pre-push hook its goal _isn't_ to be the last line of defense (there are easy ways to bypass its checks, both with `SKIP=...` and with (dreaded) `--no-verify`).  If you want pre-commit to be a last line of defense, it makes more sense to set it up either as a `pre-receive` hook or in [CI](http://pre-commit.com/#usage-in-continuous-integration).

Because of this, we can probably cut some corners on `pre-push` to bring the filenames down to a more reasonable count for the cases listed above.  Here's some of my thoughts on this (again, only talking about the fallback case where it's currently defaulting to the entire history):

- We can't assume origin names or upstream branch names (this means `origin/master` is out)
- We could use upstream branch
    - upstream can be missing: fall back to slow
    - upstream could be the same branch name: probably ok? it would compare the changed against the same branch itself
    - upstream could be a non-master branch: this is probably fine too, it's a pretty common workflow to branch off a non-master branch and a good comparison for what is being pushed is just the difference between the current commits and that
    - upstream is based on master: I think there's no qualms from me about comparing against origin/master

In conclusion, yeah let's do _something_ about this!  I see this as being very similar to the current optimizations done for [pre-commit during merge conflicts](http://pre-commit.com/#pre-commit-during-merges)

[...]

PR #667, opened 2017-11-30
compare to master as a faster baseline

orenmazor:
relevant to #666, I found that defaulting to master is doing my whole project which ends up being a bit slow.

would comparing to master make more sense?

### DIFF
--- a/pre_commit/resources/pre-push-tmpl
+++ b/pre_commit/resources/pre-push-tmpl
@@ -9,11 +9,12 @@ do
                 # Check that the ancestor has at least one parent
                 git rev-list --max-parents=0 "$local_sha" | grep "$first_ancestor" > /dev/null
                 if [ $? -ne 0 ]; then
-                    args="--all-files"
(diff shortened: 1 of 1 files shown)


pre_commit/resources/pre-push-tmpl
```diff
@@ -9,11 +9,12 @@ do
                 # Check that the ancestor has at least one parent
                 git rev-list --max-parents=0 "$local_sha" | grep "$first_ancestor" > /dev/null
```
perobertson: Would it make more sense to change this to `git rev-parse --verify "$first_ancestor~1" >/dev/null` since `first_ancestor` is set to the sha of the first commit that is not in a remote branch. That way we can leave the `--all-files` case for when its a new repo being pushed to a remote for the first time.

asottile:
This ended up getting resolved in another way -- see this comment: https://github.com/pre-commit/pre-commit/issues/666#issuecomment-362871479

PR #674, opened 2018-01-02
Simplify cross version tests

asottile:


asottile:
oh right, windows :/

I'll have to rethink version discovery to make that work.  Perhaps looking into the clever `py` launcher stuff that `tox` does.

asottile:
This should also improve interpreter detection for @gaborbernat

PR #677, opened 2018-01-09
Expand tilde in repo path

revolter:
This allows setting a repo URL like `~/Development/MyHooks`.

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import argparse
 import collections
 import functools
+import os
 
 from aspy.yaml import ordered_load
 from identify.identify import ALL_TAGS
@@ -154,6 +155,9 @@ def ordered_load_normalize_legacy_config(contents):
     if isinstance(data, list):
         # TODO: Once happy, issue a deprecation warning and instructions
         data = collections.OrderedDict([('repos', data)])
+
+    for repo_config in data['repos']:
+        repo_config['repo'] = os.path.expanduser(repo_config['repo'])
 
     return data
 
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -153,9 +153,9 @@ def ordered_load_normalize_legacy_config(contents):
     data = ordered_load(contents)
     if isinstance(data, list):
         # TODO: Once happy, issue a deprecation warning and instructions
-        return collections.OrderedDict([('repos', data)])
-    else:
-        return data
+        data = collections.OrderedDict([('repos', data)])
+
+    return data
 
 
 load_config = functools.partial(


pre_commit/commands/run.py
```diff
@@ -222,10 +223,10 @@ def run(runner, args, environ=os.environ):
...
         return 1
     if _has_unstaged_config(runner) and not no_stash:
-        logger.error(
-            'Your .pre-commit-config.yaml is unstaged.\n'
-            '`git add .pre-commit-config.yaml` to fix this.',
-        )
+        logger.error((
+            'Your {0} is unstaged.\n'
+            '`git add {0}` to fix this.'
+        ).format(C.CONFIG_FILE),)
```
asottile: unrelated (please rebase out)

pre_commit/main.py
```diff
@@ -42,7 +42,7 @@ def _add_color_option(parser):
 
 def _add_config_option(parser):
     parser.add_argument(
-        '-c', '--config', default='.pre-commit-config.yaml',
+        '-c', '--config', default=C.CONFIG_FILE,
```
asottile: unrelated (please rebase out)

pre_commit/repository.py
```diff
@@ -269,14 +269,14 @@ def _make_entry(mod):
...
-                'files': '.pre-commit-config.yaml',
+                'files': C.CONFIG_FILE,
                 'language': 'system',
                 'entry': _make_entry(check_hooks_apply),
             },
             {
                 'id': 'check-useless-excludes',
                 'name': 'Check for useless excludes',
-                'files': '.pre-commit-config.yaml',
+                'files': C.CONFIG_FILE,
```
asottile: unrelated (please rebase out)

testing/fixtures.py
```diff
@@ -47,7 +47,7 @@ def modify_manifest(path):
     with io.open(manifest_path, 'w') as manifest_file:
         manifest_file.write(ordered_dump(manifest, **C.YAML_DUMP_KWARGS))
     cmd_output(
-        'git', 'commit', '-am', 'update .pre-commit-hooks.yaml', cwd=path,
+        'git', 'commit', '-am', 'update {}'.format(C.MANIFEST_FILE), cwd=path,
```
asottile: unrelated (please rebase out)

tests/commands/install_uninstall_test.py
```diff
@@ -617,9 +617,9 @@ def test_install_allow_mising_config(tempdir_factory):
         ret, output = _get_commit_output(tempdir_factory)
         assert ret == 0
         expected = (
-            '`.pre-commit-config.yaml` config file not found. '
+            '`{}` config file not found. '
```
asottile: this actually makes the test less readable:

https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

also this is unrelated, please rebase this out.

pre_commit/clientlib.py
```diff
@@ -155,6 +156,9 @@ def ordered_load_normalize_legacy_config(contents):
         # TODO: Once happy, issue a deprecation warning and instructions
         data = collections.OrderedDict([('repos', data)])
 
+    for repo_config in data['repos']:
+        repo_config['repo'] = os.path.expanduser(repo_config['repo'])
+
```
asottile: my other hesitation is I plan to *delete* this function after the deprecation period.

revolter:
I see this as a mere enhancement, instead of a new responsibility.

Even if `try-repo` exists, you can still set `repo` to `/Users/revolt/Development/MyHooks`, so setting it to `~/Development/MyHooks` instead doesn't have any reason not to work.

I would want to create a development config that specifies a local path and the default config that specifies a GitLab URL.

revolter:
Expanding the path myself isn't possible as the user on the home computer is different than the user at the work computer.

asottile:
The reason for it to not work is pre-commit is _not a shell_.  Should `entry: foo {1,2,3}` work?  Should `repo: ${MY_WORKSPACE}/hooks` work?  No, and no.  Just trying to keep it simple, especially when it works just fine to specify the whole path or use `try-repo`.

I don't think you'd distribute a repository with a repo pointing at a local checkout either (and I'd like to discourage it if possible).  This would require extra set up by the end user which would potentially reflect poorly on the tool itself (it is after all supposed to be the thing that bootstraps all of this stuff for you!).

The fact that pre-commit supports local paths at all is a bit of an accident of implementation -- it's incredibly useful for testing repositories but I really don't think it has a strong use case for being checked in that way and used as a long term workflow.

pre_commit/clientlib.py
```diff
@@ -153,9 +153,9 @@ def ordered_load_normalize_legacy_config(contents):
     data = ordered_load(contents)
     if isinstance(data, list):
         # TODO: Once happy, issue a deprecation warning and instructions
-        return collections.OrderedDict([('repos', data)])
-    else:
-        return data
+        data = collections.OrderedDict([('repos', data)])
+
+    return data
```
asottile: fwiw this style was intentional -- it's very easy for a human to see at a glance that there are two disparate paths that the code could take.  By decomposing the `if` / `else`, there's overhead to parsing the code.

revolter: Interesting view, thanks.

revolter:
Makes sense, thank you! At least I got to understand the codebase a little 😄

asottile:
Cheers!  Thanks again for the PR (and thanks for the other PR as well!), even though it doesn't make sense for now.

revolter:
My pleasure, thanks for an awesome tool!

PR #678, opened 2018-01-09
Replace string literals with constants

revolter:


pre_commit/commands/run.py
```diff
@@ -222,10 +223,10 @@ def run(runner, args, environ=os.environ):
...
         return 1
     if _has_unstaged_config(runner) and not no_stash:
-        logger.error(
-            'Your .pre-commit-config.yaml is unstaged.\n'
-            '`git add .pre-commit-config.yaml` to fix this.',
-        )
+        logger.error((
+            'Your {0} is unstaged.\n'
+            '`git add {0}` to fix this.'
+        ).format(C.CONFIG_FILE),)
```
asottile: shouldn't have a comma here.

Also now that I notice it, this error message could actually be incorrect in the case of a custom `--config` -- it should probably format in the `runner`'s configuration path

revolter: The comma was there previously too. I found it odd too but didn't know the reason.

asottile: and probably with the updated substitution this message should instead read something like:

```python
logger.error(
    'Your pre-commit configuration is unstaged.\n'
    '`git add {}` to fix this.'.format(runner.config_file),
)
```

asottile: The comma was there, but the parens were not as you've put them.

For rationale on the comma [see this](https://github.com/asottile/add-trailing-comma#multi-line-method-invocation-style----why)

revolter:
And thank you for the edits on this!

PR #685, opened 2018-01-12
Support node on windows with long path hack 

### LINKED ISSUE #200: Windows: Node Support

asottile:
Resolves #200

Huge huge huge thanks to @wenzowski :tada: :raised_hands:


asottile:
ok, so there's a bug in cpython 2.7: https://bugs.python.org/issue32539

I have a patch for it though: https://github.com/python/cpython/pull/5169

Issue #689, opened 2018-01-17
Allow certain hook to be verbose


bagerard:
Hi,
While integrating flake8 as a hook, I'm trying to have certain checks (like complexity warnings) run on every commit but I don't necessarily want them to be blocking. flake8 has this very nice --exit-zero flag that allows this but the only way to get the output printed through pre-commit is to run pre-commit with --verbose but this prints a ton of useless information from other hooks.

My suggestion is to allow a 'verbose' flag to be set per hook as opposed to existing --verbose which is set globally. IMO it would allow to have extremely valuable information out of flake8 and pylint for instance. I could even run part of flake8 as being blocking, and other part as being non-blocking.
Let me know your thoughts, I'd be happy to work on this

asottile:
First off, thanks for the issue :)

So I thought I had addressed this somewhere else before but I couldn't find it -- it's possible I'm remembering a private discussion about the same thing.  I'll try and reiterate my thoughts on that here and we can go from there!

One design goal of a lot of unix utilities is "silence is golden" and for pre-commit the meaning takes a bit of a different form.  As pre-commit is intended to be ~somewhat of a gateway for vetting changes it should generally be as quiet as possible unless something goes wrong.  In other words, attempt to keep as high of a signal-to-noise ratio as possible.  This is also suggested in the design of hooks that plug into the framework ([an example](https://github.com/tox-dev/tox/pull/708#issuecomment-350567303)) -- by default they provide just as much information to answer why they're stopping the commit at all (preferring "filename.py: import reordered" over a full-on diff of a file).

The previous ~similar systems that `pre-commit` replaced had a huge verbosity problem in that _most_ of their output was "warnings".  Yes the messages were probably useful if you were going to take the time to read them, but for 99% of developers if it wasn't stopping them from committing they did not care and ignored the messages.  Not only did this decrease the likelihood that these developers would fix the problems presented in warnings, but it also hurt the attentiveness to messages in other (related **_and_** unrelated) commandline tools.  I'm probably stealing the name from somewhere but I usually refer to this problem as _warning fatigue_.  More succinctly, if users are presented with lots of non-critical information, they are trained to ignore all of the (important) output.

I think you'll find that while you may read the warning messages about code complexity and such, other contributors will filter this out -- and worse may filter out _all_ of pre-commit's messaging.

All that said, implementing such a thing is trivial (though I personally wouldn't use it and I would generally discourage anyone from using it ;) ).

[...]

PR #695, opened 2018-01-31
Add a verbose hook option

bagerard:
Relates to #689
It adds a `verbose` flag that can be set per hook and that allows stdout to be printed even when hook returned 0

### DIFF
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -292,6 +292,23 @@ def test_always_run_alt_config(
     )
 
 
+def test_hook_verbose_enabled(
+        cap_out, repo_with_passing_hook, mock_out_store_directory,
+):
+    with modify_config() as config:
+        config['repos'][0]['hooks'][0]['always_run'] = True
+        config['repos'][0]['hooks'][0]['verbose'] = True
+
+    _test_run(
+        cap_out,
+        repo_with_passing_hook,
+        {},
+        (b'Hello World',),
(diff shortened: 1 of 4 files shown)


tests/commands/run_test.py
```diff
@@ -292,6 +292,53 @@ def test_always_run_alt_config(
...
+        config['repos'][0]['hooks'][0]['verbose'] = True
+
+    _test_run(
+        cap_out,
+        repo_with_passing_hook,
+        {},
+        (b'Hello World',),
+        0,
+        stage=False,
+    )
```
asottile: this test is probably sufficient, the others I don't think are super necessary (we already test all three of those scenarios)

asottile:
Also looks like there's a test that needs `'verbose': False` here: 

https://github.com/pre-commit/pre-commit/blob/7f0b427b7475796d9eb861744aeb68601238a5d5/tests/repository_test.py#L791-L792

bagerard:
Thanks @asottile for the immediate review. Should I do a PR to document the flag in pre-commit.github.io ?

asottile:
@bagerard that would be awesome :) source is https://github.com/pre-commit/pre-commit.github.io (branch from `real_master`)

bagerard:
Indeed I should have thought to run pre-commit on pre-commit :) Here you go

asottile:
Thanks again!

PR #715, opened 2018-02-25
Migrate sha -> rev

### LINKED ISSUE #106: Support tags as well as shas
They kind-of work as-is unless tags collide across different hooks.

The simplest solution is to make the repository symlinks `repo + ref` instead of just `ref` as they are right now.

Then anything that git understands as `git checkout {0}` will work.

asottile:
Resolves #106 

The ticket mentions "ref", but I think "rev" is perhaps a better name?  Would love to get input here

This will be a slow migration, and during that time both `sha` and `rev` will be supported (but not both!)

I was able to leverage a [custom `cfgv` validator](https://github.com/asottile/cfgv#writing-your-own-validator) to accomplish this cleanly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,16 +13,16 @@ repos:
     -   id: requirements-txt-fixer
     -   id: flake8
 -   repo: https://github.com/pre-commit/pre-commit.git
-    sha: v0.16.3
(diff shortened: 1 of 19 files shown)


asottile:
People may typo "rev" <=> "ref" -- if this is a common issue I'll add an alias.

Issue #634, opened 2017-10-10
Deprecate `pcre` language


asottile:
Now that pygrep (#630) is a much more portable alternative, pcre is unnecessary and should be deprecated.

A deprecation warning should be issued when loading a configuration containing `language: pcre` and should point the consumer in the right direction to correcting it (either by suggesting a pull request, or by indicating they should modify their `local` configuration).

The `pcre` language will likely be removed in `pre-commit==2.0.0`

PR #716, opened 2018-02-26
Add multiline mode to pygrep

tdeo:
As a follow-up on https://github.com/pre-commit/pre-commit/issues/634, I'm using today pcre hooks with `-z` option, which is the multiline mode, I'd like to add support for this to be able to move to pygrep hooks. I used the same argument name as `grep` does

### DIFF
--- a/pre_commit/languages/pygrep.py
+++ b/pre_commit/languages/pygrep.py
@@ -30,13 +30,21 @@ def _process_filename_by_line(pattern, filename):
 def _process_filename_at_once(pattern, filename):
     retv = 0
     with open(filename, 'rb') as f:
-        match = pattern.search(f.read())
+        contents = f.read()
+        match = pattern.search(contents)
         if match:
             retv = 1
-            output.write(
-                '{}:{}-{}:'.format(filename, match.start(), match.end()),
+            line_no = len(
+                re.compile('\n'.encode()).findall(contents, 0, match.start()),
             )
-            output.write_line(match.group())
+            output.write(
+                '{}:{}:'.format(filename, line_no + 1),
+            )
+
+            matched_lines = match.group().split('\n')
+            matched_lines[0] = contents.split('\n')[line_no]
+
+            output.write_line('\n'.join(matched_lines))
     return retv
 
 
@@ -55,20 +63,20 @@ def main(argv=None):
         ),
     )
     parser.add_argument('-i', '--ignore-case', action='store_true')
-    parser.add_argument('-z', '--null-data', action='store_true')
+    parser.add_argument('--multiline', action='store_true')
     parser.add_argument('pattern', help='python regex pattern.')
     parser.add_argument('filenames', nargs='*')
     args = parser.parse_args(argv)
 
     flags = re.IGNORECASE if args.ignore_case else 0
-    if args.null_data:
+    if args.multiline:
         flags = flags | re.MULTILINE | re.DOTALL
 
     pattern = re.compile(args.pattern.encode(), flags)
 
     retv = 0
     for filename in args.filenames:
-        if args.null_data:
+        if args.multiline:
             retv |= _process_filename_at_once(pattern, filename)
         else:
             retv |= _process_filename_by_line(pattern, filename)
--- a/pre_commit/languages/pygrep.py
+++ b/pre_commit/languages/pygrep.py
@@ -29,7 +29,7 @@ def _process_filename_by_line(pattern, filename):
 def _process_filename_at_once(pattern, filename):
     retv = 0
     with open(filename, 'rb') as f:
-        match = pattern.search(f.read())
+        match = pattern.search(f.read().decode('utf-8').replace('\n',''))
         if match:
             retv = 1
             output.write('{}:'.format(filename))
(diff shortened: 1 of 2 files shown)


pre_commit/languages/pygrep.py
```diff
@@ -27,6 +27,17 @@ def _process_filename_by_line(pattern, filename):
     return retv
 
 
+def _process_filename_at_once(pattern, filename):
+    retv = 0
+    with open(filename, 'rb') as f:
+        match = pattern.search(f.read().decode('utf-8').replace('\n', ''))
```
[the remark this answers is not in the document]
tdeo: Indeed, I was running tests with Python2 and had encoding issues there.

pre_commit/languages/pygrep.py
```diff
@@ -42,6 +53,7 @@ def main(argv=None):
         ),
     )
     parser.add_argument('-i', '--ignore-case', action='store_true')
+    parser.add_argument('-z', '--null-data', action='store_true')
```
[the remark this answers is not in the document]
tdeo: Not sure what you mean here. What is zsplit ?

tdeo:
I made a few changes, now using both `DOTALL` and `MULTILINE` flags. I think it's simple enough this way, but maybe worth renaming the option to `--multiline` or something ?

pre_commit/languages/pygrep.py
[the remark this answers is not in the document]
asottile:
```
$ git grep 'def zsplit'
pre_commit/git.py:def zsplit(s):
```

`zsplit` is essentially the splitlines for `-z` (split lines delimited by `\0`)

tests/languages/pygrep_test.py
```diff
@@ -38,3 +38,24 @@ def test_ignore_case(some_files, cap_out):
...
+    out = cap_out.get()
+    assert ret == 1
+    assert out == 'f1:0-7:foo\nbar\n'
+
+
+def test_null_data_dotall_flag_is_enabled(some_files, cap_out):
+    ret = pygrep.main(('--null-data', r'o.*bar', 'f1', 'f2', 'f3'))
+    out = cap_out.get()
+    assert ret == 1
+    assert out == 'f1:1-7:oo\nbar\n'
```
asottile: I think my ideal output for this would show the full line numbers, the character offsets are a bit misleading (and less useful) imo

asottile: that said, as a v1, I'd be ok with the same output that `grep -z` gives

pre_commit/languages/pygrep.py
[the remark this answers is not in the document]
tdeo: Oh, so that's what the `-z` option in grep means. I ever thought it was replacing newlines with null bytes in the input.

tests/languages/pygrep_test.py
tdeo: I'm not a big fan of the `grep -z` output in general, I always use it in conjuction with `-bo` to get only the matching part and the byte offset instead of the line number (which is always `1` when you don't have null bytes in the file).

tdeo:
So I actually figured a way around to display the multiline line match with the line number and renamed the option, let me know what you think

pre_commit/languages/pygrep.py
```diff
@@ -27,6 +27,27 @@ def _process_filename_by_line(pattern, filename):
...
 
 
+def _process_filename_at_once(pattern, filename):
+    retv = 0
+    with open(filename, 'rb') as f:
+        contents = f.read()
+        match = pattern.search(contents)
+        if match:
+            retv = 1
+            line_no = len(
```
asottile: more simply: `contents[:match.start()].count(b'\n')`

pre_commit/languages/pygrep.py
```diff
@@ -27,6 +27,27 @@ def _process_filename_by_line(pattern, filename):
...
+                re.compile('\n'.encode()).findall(contents, 0, match.start()),
+            )
+            output.write(
+                '{}:{}:'.format(filename, line_no + 1),
+            )
```
asottile: minor: probably fits on one line

pre_commit/languages/pygrep.py
```diff
@@ -27,6 +27,27 @@ def _process_filename_by_line(pattern, filename):
...
+
+            matched_lines = match.group().split('\n')
+            matched_lines[0] = contents.split('\n')[line_no]
+
+            output.write_line('\n'.join(matched_lines))
```
asottile: the three `'\n'` literals here I believe need to be `b'\n'`

pre_commit/languages/pygrep.py
```diff
@@ -42,16 +63,23 @@ def main(argv=None):
...
     )
     parser.add_argument('-i', '--ignore-case', action='store_true')
+    parser.add_argument('--multiline', action='store_true')
     parser.add_argument('pattern', help='python regex pattern.')
     parser.add_argument('filenames', nargs='*')
     args = parser.parse_args(argv)
 
     flags = re.IGNORECASE if args.ignore_case else 0
+    if args.multiline:
+        flags = flags | re.MULTILINE | re.DOTALL
```
asottile: could even write `flags |= ...`

asottile:
👍 this is great -- thanks for hacking this out :)

tdeo:
Thanks for your help ! Hope you'll release soon

asottile:
currently aiming for some time this weekend :)

asottile:
This has been released as part of [v1.8.0](https://github.com/pre-commit/pre-commit/releases/tag/v1.8.0)!

PR #721, opened 2018-03-08
Normalize git urls in pre-commit config

asottile:
Improves cache performance

Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks.git
+-   repo: https://github.com/pre-commit/pre-commit-hooks
     sha: v0.9.1
     hooks:
     -   id: trailing-whitespace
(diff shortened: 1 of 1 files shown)


chriskuehl:
@asottile this gives me a crazy idea... what if the cache key used for a `(repo, ref)` was just the sha, and didn't involve the repo path at all?

In practice, a git SHA is unique across all git repositories (yeah okay, I know Google finally found the first sha1 collision, but it took them 110 GPU years of work), so a git SHA could be a decent key. It would mean that the same repository and revision could use the same cache regardless of how you get to it.

Not sure if this is a good idea, especially given that SHA1 has recently been (somewhat) broken. An interesting idea, though :)

asottile:
This is how it [used to work](https://github.com/pre-commit/pre-commit/pull/127/files#diff-15f18d316fa8b61d817bff726b3a9711L107), but then tags :)

chriskuehl:
Ah, makes sense. I guess it's probably not worth adding another layer of indirection (i.e. caching the (repo, revision) => full SHA) just to improve the hit rate a bit.

asottile:
yeah plus it would make branches even worse if I'm thinking about it correctly :)

PR #724, opened 2018-03-12
Restore git 1.8 support

### LINKED ISSUE #723: Older versions of git no longer supported
Since #714 I am no longer able to run `pre-commit` on RHEL7 hosts.  The version of `git` provided by the OS vendor does not support the `-C` option.

Was this intentional?  If not, would you consider re-adding support for older versions?

My environment:
```
$ cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.3 (Maipo)

$ git --version
git version 1.8.3.1
```
I understand that version of `git` is 5+ years old, but it is the latest version provided by RedHat.  If nothing else, the documentation should state which version of `git` is required.

Thanks

asottile:
Resolves #723

I manually verified the testsuite against git 1.8 (built from source) -- I don't think it's worth adding to the CI suite but that could be implemented as a followup.

When running the testsuite against 1.8 I also fixed ~3 tests in `tests/commands/install_uninstall_test.py` which are unrelated to the changes here (though in adjacent lines)

### DIFF
--- a/latest-git.sh
+++ b/latest-git.sh
@@ -3,6 +3,5 @@
 set -ex
 git clone git://github.com/git/git --depth 1 /tmp/git
 pushd /tmp/git
-make prefix=/tmp/git -j 8 all
-make prefix=/tmp/git install
+make prefix=/tmp/git -j8 install
 popd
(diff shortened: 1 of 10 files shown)


asottile:
More specifically, to restore `git<=1.8.4` support.  `git -C ...` was added in git 1.8.5 😞 

I've retested against 1.8.3.1 manually (the skipped test is the `swift` test):

```
=================== 482 passed, 1 skipped in 171.99 seconds ====================
```

asottile:
I'll need to rebase after https://github.com/pre-commit/pre-commit/pull/725 merges

PR #738, opened 2018-04-25
Allow opt-in for using venv over virtualenv for python hooks

### LINKED ISSUE #631: macOS 10.12.6 -- ERROR: virtualenv is not compatible with this system or executable
I'm using python3.6 in venv on macOS 10.12.6:

```bash
(venv) Dongyus-MBP:resume-backend gary$ pre-commit
[INFO] Installing environment for https://github.com/asottile/reorder_python_imports.
An unexpected error has occurred: CalledProcessError: Command: ('/Users/gary/backend/venv/bin/python3.6', '-m', 'virtualenv', '/Users/gary/.cache/pre-commit/reporzq6xjqs/py_env-python3.6', '-p', 'python3.6')
Return code: 100
Expected return code: 0
Output:
    Already using interpreter /Users/gary/backend/venv/bin/python3.6
    Using base prefix '/usr/local/bin/../Cellar/python3/3.6.2/bin/../Frameworks/Python.framework/Versions/3.6'
    New python executable in /Users/gary/.cache/pre-commit/reporzq6xjqs/py_env-python3.6/bin/python3.6
    Also creating executable in /Users/gary/.cache/pre-commit/reporzq6xjqs/py_env-python3.6/bin/python
    ERROR: The executable /Users/gary/.cache/pre-commit/reporzq6xjqs/py_env-python3.6/bin/python3.6 is not functioning
    ERROR: It thinks sys.prefix is '/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6' (should be '/Users/gary/.cache/pre-commit/reporzq6xjqs/py_env-python3.6')
    ERROR: virtualenv is not compatible with this system or executable

Errors: (none)

Check the log at /Users/gary/.cache/pre-commit/pre-commit.log
```

ojii:
This fixes #631.

Using `+venv` as a suffix to `language_version` allows you to instruct pre-commit to use the `venv` module instead of `virtualenv` when creating environments for the hooks. This helps compatibility (especially on OSX).

(Note: edited from original proposal)

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import contextlib
 import os
+import subprocess
 import sys
 
 from pre_commit.envcontext import envcontext
@@ -13,6 +14,11 @@ from pre_commit.util import CalledProcessError
 from pre_commit.util import clean_path_on_failure
 from pre_commit.util import cmd_output
 from pre_commit.xargs import xargs
+
+try:
+    devnull = subprocess.DEVNULL
+except AttributeError:
+    devnull = open(os.devnull, 'w')
 
 
 ENVIRONMENT_DIR = 'py_env'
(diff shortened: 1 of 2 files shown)


asottile:
The failures on linux are exactly why I'm hesitant to use the `venv` module.

ojii:
> The failures on linux are exactly why I'm hesitant to use the venv module.

I see. Would you instead accept a PR that allows using venv by using `language_version: <version>+venv`? Eg `language_version: 3.6+venv`.

asottile:
Yeah! That would be cool, an explicit opt in sounds good to me!

asottile:
I'll take a look at this in the morning when I'm thinking more clearly ;) -- currently considering this approach vs making a `python_venv` "language" instead which would be opted into with:

```yaml
    language: python_venv
```

or

```yaml
    language: python_venv
    language_version: python3.6
```

It would share most of the code with the `python` language except the venv setup (and wouldn't require so much branch modification).  Need to sleep now though :)

jfroche:
Hello, I was wondering if it would be ok to use the python api for venv (https://docs.python.org/3/library/venv.html#api) instead of calling subprocess ?

asottile:
@jfroche I don't think it's possible to support `language_version` via the api only

jfroche:
@asottile not sure I understand. Do you mean that all functionalities in the venv cli are not present in the api ? 
What do you think of an implementation where `def install_environment` installs an environment with venv api only if language_version ends with `+venv` ?

ojii:
Will open a fresh PR with a new language.

PR #739, opened 2018-04-27
added python venv language

ojii:
Next attempt at #738 (and #631).

Adds a new language `python_venv`. It works like `python` but uses `python -m venv` instead of `python -m virtualenv`. 

I couldn't re-use as much code from `python` as I would have liked because of `in_env`, but I don't think it's too bad.

Manually tested on my machine in a project using `venv` where the default `pre-commit` didn't work.


asottile:
cool, I'll help write up some tests and simplify the code a bit -- are you ok with me pushing to your branch? (figured I'd ask before I just do it 😆)

ojii:
go right ahead

asottile:
(just a note: I haven't forgotten about this! just busy with travel and work 😆)

artcyb:
hahaha im dizzy

ojii:
Could you make a release with this please?

asottile:
yep! planning to do so, thought I'd wait just a bit to see if I could get @chriskuehl's contribution which adds rust support into this release but I guess I can just make another release after that's ready :)

asottile:
This has been released as part of [v1.9.0](https://github.com/pre-commit/pre-commit/releases/tag/v1.9.0)!

PR #744, opened 2018-05-03
Fixes file `types` from .pre-commit-config.yaml issue

zph:
This PR fixes a bug where files are not being recognized based on their `types` using configuration file setting `types`.

# Issue / Reproduction case
When using a pre-commit-config.yaml in a repository with many bash scripts (that lack a `.sh` suffix to filename):

```
- repo: local
 - id: shellcheck
    name: 'Shellcheck Linter'
    entry: shellcheck
    language: system
    types:
      - shell
      - sh
      - bash
```

And running:
```
pre-commit run --all-files
```

## Actual output:
```
Shellcheck Linter....................................(no files to check)Skipped
```

## Expected behavior
That pre-commit finds the files with a shebang of `#!/usr/bin/env bash` and lints them.

# Solution

I added two test cases, one for Python (which did behave properly) and one for a bash script like: 

Filename: `bash_script`
```
#!/usr/bin/env bash
echo "Hello"
```

Then I tracked down the issue to the comparison between a List and a FrozenSet here:
```
if tags >= types and not tags & exclude_types:
  ret.append(filename)
```

I rewrote that comparison to compare two FrozenSets using the `intersection` function. This produces predictable results on Python 2.7 and 3.5.

Once this is approved, I'll happily squash it down to a single commit. Right now it's 2 commits to show the spec failing before the patch.

Note: I stubbed in a `get_tags` function via dependency injection to simplify testing. If you prefer a different mechanism, feel free to adjust the PR to suit your testing preferences.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -48,11 +48,14 @@ def _filter_by_include_exclude(filenames, include, exclude):
     }
 
 
-def _filter_by_types(filenames, types, exclude_types):
+def _filter_by_types(filenames,
+                     types,
+                     exclude_types,
+                     get_tags=tags_from_path):
     types, exclude_types = frozenset(types), frozenset(exclude_types)
     ret = []
     for filename in filenames:
-        tags = tags_from_path(filename)
+        tags = get_tags(filename)
         if tags >= types and not tags & exclude_types:
             ret.append(filename)
     return tuple(ret)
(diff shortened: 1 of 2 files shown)


zph:
@asottile Please let me know if you think this is my PR breaking the tests. I was unable to get them passing locally even before making changes.

artcyb:
thanks

PR #751, opened 2018-05-16
Add Rust support

chriskuehl:
Here's an example hook that works: https://github.com/chriskuehl/example-rust-pre-commit-hook

Filing the PR a bit early so I can take advantage of Travis to run CI.

Still TODO:
* [x] Tests
* [x] CI support (at least Travis, hopefully AppVeyor)
* [x] Test it with some real Rust linters. The example one I wrote works, but I want to make sure this is actually applicable to real linters.
* [x] Make sure local hooks work (might need to add a blank `Cargo.toml` to the template)
* [ ] (maybe one day) consider using rustup to grab a specific version of rust, rather than relying on the system installation (similar to ruby)


chriskuehl:
Travis and AppVeyor are both passing :D

It adds about 20 seconds to the Travis pre-install step which isn't great but probably okay? (swift adds like 23 seconds /shrug)

I tested this out with shellharden, a real Rust linter (and a pretty cool one), and it seems to work. Going to try this with a couple other linters (maybe clippy) before I check off that box.

chriskuehl:
Tried this out with some real Rust linters: [shellharden](https://github.com/anordal/shellharden), [systemd-linter](https://github.com/mackwic/systemd-linter), [rustfmt-nightly](https://github.com/rust-lang-nursery/rustfmt), and [clippy](https://github.com/rust-lang-nursery/rust-clippy) via the mirror maker, they all worked :D.

Gonna check off that box, just need to make local hooks work and then I think this will be good to merge.

chriskuehl:
partycat

PR #762, opened 2018-06-04
Simplify setup.py in arbitrary_bytes_repo

asottile:


asottile:
I think 37f49d8 is only necessary due to a regression in git 2.17 -- going to follow up with that one on the mailing list

asottile:
Here's my mailing list entry: https://marc.info/?l=git&m=152809164601078&w=2

PR #781, opened 2018-07-04
Travis cache

expobrain:
Added cache for Travis and Swift assets

### DIFF
--- a/testing/get-swift.sh
+++ b/testing/get-swift.sh
@@ -4,13 +4,18 @@ set -ex
 
 . /etc/lsb-release
 if [ "$DISTRIB_CODENAME" = "trusty" ]; then
-    SWIFT_URL='https://swift.org/builds/swift-4.0.3-release/ubuntu1404/swift-4.0.3-RELEASE/swift-4.0.3-RELEASE-ubuntu14.04.tar.gz'
+    SWIFT_TARBALL="swift-4.0.3-RELEASE-ubuntu14.04.tar.gz"
+    SWIFT_URL="https://swift.org/builds/swift-4.0.3-release/ubuntu1404/swift-4.0.3-RELEASE/$SWIFT_TARBALL"
 else
-    SWIFT_URL='https://swift.org/builds/swift-4.0.3-release/ubuntu1604/swift-4.0.3-RELEASE/swift-4.0.3-RELEASE-ubuntu16.04.tar.gz'
+    SWIFT_TARBALL="swift-4.0.3-RELEASE-ubuntu16.04.tar.gz"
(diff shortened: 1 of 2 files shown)


asottile:
🤔 
```
WARNING: timestamping does nothing in combination with -O. See the manual
for details.
```

expobrain:
Argh, missed that! Let me check for alternatives

testing/get-swift.sh
```diff
@@ -4,13 +4,18 @@ set -ex
...
-    SWIFT_URL='https://swift.org/builds/swift-4.0.3-release/ubuntu1404/swift-4.0.3-RELEASE/swift-4.0.3-RELEASE-ubuntu14.04.tar.gz'
+    SWIFT_TARBALL="swift-4.0.3-RELEASE-ubuntu14.04.tar.gz"
+    SWIFT_URL="https://swift.org/builds/swift-4.0.3-release/ubuntu1404/swift-4.0.3-RELEASE/$SWIFT_TARBALL"
 else
-    SWIFT_URL='https://swift.org/builds/swift-4.0.3-release/ubuntu1604/swift-4.0.3-RELEASE/swift-4.0.3-RELEASE-ubuntu16.04.tar.gz'
+    SWIFT_TARBALL="swift-4.0.3-RELEASE-ubuntu16.04.tar.gz"
+    SWIFT_URL="https://swift.org/builds/swift-4.0.3-release/ubuntu1604/swift-4.0.3-RELEASE/$SWIFT_TARBALL"
 fi
 
+pushd "$HOME"/.swift
```
asottile: I'm surprised (?) that this directory doesn't need to be created first?  should there be a `mkdir -p "$HOME/.swift"` above this? (note that I also use this script to test locally :D)

asottile: another idea if wget doesn't pan out (though it seems from travis-ci that this works!) would be to checksum test the file and retrieve if incorrect (missing)

expobrain: `pushd`: 

now that you mention that I'm surprise as well. My untested theory is that Travis creates the directory in `cache->directories` ahead of time so `pushd` is not failing.

expobrain: `wget`:

you mean storing the hardcoded checksum for the Swift tarballs in the script and check them at runtime? Can do it, probably a little paranoid-ish but still a good point

asottile: I meant more if we couldn't find a solution that worked with wget :)

let's add the `mkdir -p` so I can keep using this script locally and then I'll merge this!

asottile:
Thanks again!

PR #789, opened 2018-07-05
Added Python 3.7 to Travis CI

expobrain:


.travis.yml
```diff
@@ -13,6 +13,8 @@ matrix:
             python: 3.6
         -   env: TOXENV=pypy
             python: pypy2.7-5.10.0
+        -   env: TOXENV=py37
+            python: 3.7-dev
```
asottile: 3.7-dev is 3.7a4 which is not much like 3.7.0 final

You can read more about my writeup about this: https://github.com/deadsnakes/travis-ci-python3.7-example#travis-ci-python37-example

setup.py
```diff
@@ -41,7 +41,7 @@
         'cfgv>=1.0.0',
         'identify>=1.0.0',
         'nodeenv>=0.11.1',
-        'pyyaml',
+        'pyyaml>=4.2b4',
```
asottile: pyyaml 4.x is very much in flux, and pre-commit works fine with the 3.x version

Fortunately, there's supposed to be a `3.13` release that comes out today -- but isn't out yet https://github.com/yaml/pyyaml/issues/193#issuecomment-402333150

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 project = pre_commit
 # These should match the travis env list
-envlist = py27,py35,py36,pypy
+envlist = py27,py35,py36,py37,pypy
```
asottile: when adding py37, we can probably remove py35 -- we don't have any python3.5 specific code and I want to keep travis-ci fast :)

setup.py
```diff
@@ -41,7 +41,7 @@
         'cfgv>=1.0.0',
         'identify>=1.0.0',
         'nodeenv>=0.11.1',
-        'pyyaml>=4.2b4',
+        'pyyaml>=3.13',
```
asottile: this should be fine to just leave unbounded -- pip will choose the latest version already as long as it isn't a pre-release

asottile:
Thanks again!

PR #790, opened 2018-07-08
Don't test python3.5 now that we test python3.7

asottile:


asottile:
currently there's no minor-version-specific code and we still test python2.  No bugs slipped while not testing python3.3 / 3.4 so I don't expect them to here either.  imo the 10-12 minutes per PR is well worth the risk

PR #803, opened 2018-07-20
Run only the specified hook even when stages exist in config.

### LINKED ISSUE #772: `stages: [commit]` hooks will run with `pre-commit run otherhookid`
minor logic bug, good new-contributor ticket

Easy to reproduce on pre-commit itself:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index a146bd2..7bb382d 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,7 @@ repos:
     rev: v1.2.3
     hooks:
     -   id: trailing-whitespace
+        stages: [commit]
     -   id: end-of-file-fixer
     -   id: autopep8-wrapper
     -   id: check-docstring-first
```

```console
$ pre-commit run end-of-file-fixer --all-files
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
```

(it should have only run `end-of-file-fixer` but also run `trailing-whitespace` due to a logic error).

mblayman:
This branches fixes the run logic so that when
`pre-commit run some_hook -a` runs when the config contains
`stages: ['commit']` for some other hook, only the hook specified as an
argument will run.

Fixes #772

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -254,11 +254,13 @@ def run(runner, store, args, environ=os.environ):
         repo_hooks = []
         for repo in repositories(runner.config, store):
             for _, hook in repo.hooks:
-                if (
-                    (not args.hook or hook['id'] == args.hook) and
-                    not hook['stages'] or args.hook_stage in hook['stages']
-                ):
-                    repo_hooks.append((repo, hook))
+                if args.hook:
+                    if args.hook == hook['id']:
+                        repo_hooks.append((repo, hook))
+                        break
+                else:
+                    if not hook['stages'] or args.hook_stage in hook['stages']:
+                        repo_hooks.append((repo, hook))
 
         if args.hook and not repo_hooks:
             output.write_line('No hook with id `{}`'.format(args.hook))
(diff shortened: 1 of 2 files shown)


pre_commit/commands/run.py
```diff
@@ -254,11 +254,13 @@ def run(runner, store, args, environ=os.environ):
         repo_hooks = []
         for repo in repositories(runner.config, store):
             for _, hook in repo.hooks:
-                if (
-                    (not args.hook or hook['id'] == args.hook) and
-                    not hook['stages'] or args.hook_stage in hook['stages']
```
asottile: I might be wrong but I think  adding parens around this second condition is sufficient (which would be a simpler fix). Let me know if that works :)

mblayman:
The negative logic sort of melts my brain, but it seems to work so I updated my PR as you requested. 😄

mblayman:
Hmm... this is probably super edge case, but I guess if you had some config like:
```
- id: some_hook
  stages: ['manual']
```
and ran something like `pre-commit run --hook-stage commit some_hook`, it would fail to run `some_hook`. Does that sound right?

asottile:
Yeah I think that's expected - - a but strange, but expected :)

asottile:
Thanks again for this PR! Really awesome :)

mblayman:
No problem. I was scratching an itch for work so I'm happy to help.

PR #811, opened 2018-08-10
Fix several ResourceWarning: unclosed file

BoboTiG:
Fixes a lot of ResourceWarning like:
```python
pre_commit/repository.py:46: ResourceWarning: unclosed file <_io.TextIOWrapper name='~/.cache/pre-commit/repozvpv73iz/py_env-python3.6/.install_state_v1' mode='r' encoding='UTF-8'>
  return json.loads(io.open(filename).read())
```
I also made changes to use the `with` context manager everywhere possible to prevent any other leaks.

### DIFF
--- a/pre_commit/repository.py
+++ b/pre_commit/repository.py
@@ -43,7 +43,8 @@ def _read_state(prefix, venv):
     if not os.path.exists(filename):
         return None
     else:
-        return json.loads(io.open(filename).read())
+        with io.open(filename) as f:
+            return json.loads(f.read())
 
 
 def _write_state(prefix, venv, state):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,6 +21,26 @@ from testing.fixtures import write_config
 from testing.util import cwd
 
 
+@pytest.fixture(autouse=True)
+def no_warnings(recwarn):
+    yield
+    warnings = []
+    for warning in recwarn:  # pragma: no cover
+        message = str(warning.message)
+        # ImportWarning: Not importing directory '...' missing __init__(.py)
+        if not (
+            isinstance(warning.message, ImportWarning)
+            and message.startswith('Not importing directory ')
+            and ' missing __init__' in message
+        ):
+            warnings.append('{}:{} {}'.format(
+                warning.filename,
+                warning.lineno,
+                message,
+            ))
+    assert not warnings
+
+
 @pytest.fixture
 def tempdir_factory(tmpdir):
     class TmpdirFactory(object):
--- a/tests/languages/all_test.py
+++ b/tests/languages/all_test.py
@@ -15,7 +15,7 @@ if six.PY2:  # pragma: no cover
         inspect.ArgSpec, varargs=None, keywords=None, defaults=None,
     )
     getargspec = inspect.getargspec
-else:
+else:  # pragma: no cover
     ArgSpec = functools.partial(
         inspect.FullArgSpec, varargs=None, varkw=None, defaults=None,
         kwonlyargs=[], kwonlydefaults=None, annotations={},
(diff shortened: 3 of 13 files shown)


pre_commit/repository.py
```diff
@@ -43,7 +43,8 @@ def _read_state(prefix, venv):
     if not os.path.exists(filename):
         return None
     else:
-        return json.loads(io.open(filename).read())
+        with io.open(filename) as fh:
```
asottile: very very minor: the rest of the code conventionally uses `f` as a variable name here -- mind changing them all to use that?

BoboTiG: Of course!

tests/conftest.py
```diff
@@ -21,6 +21,22 @@
...
+    yield
+    warnings = tuple(
+        warning for warning in recwarn
+        # python2 + pypy warn:
+        # ImportWarning: Not importing directory '...' missing __init__.py
+        if not (
+            isinstance(warning.message, ImportWarning) and
+            str(warning.message).startswith('Not importing directory ') and
+            str(warning.message).endswith(' missing __init__.py')
+        )
```
asottile: you probably don't need this filtering here (I would hope?) -- I think this was just due to testing a namespace package in the other library

BoboTiG: Nope, it is needed because on Python 3.6 and 3.7 there are a lot of these warnings.

BoboTiG: Ah. I just saw the check is not good on Python 6:
```python
python3.6/importlib/_bootstrap_external.py:426: ImportWarning: Not importing directory ~/python3.6/site-packages/aspy: missing __init__
```
There is no `.py`. I will change that.

asottile: Ah lol of course, the same name space package is still in play

BoboTiG: Tests are failing on Python 3.7. But the warnings list is not helpful. I updated to patch to have a more human message.

BoboTiG:
So, there are still ResourceWarning. But I do not if it is related to this project or third-party one. I will improve the warning logging o add the source module and lineno. I will let you knwo when the PR is OK to review again :)

BoboTiG:
I cannot run test on `python -m pytest -l -x -v tests/repository_test.py`:
```python
tests/repository_test.py:66: in <module>
    @pytest.mark.integration
../ven36/lib/python3.6/site-packages/_pytest/mark/structures.py:367: in __getattr__
    self._check(name)
../ven36/lib/python3.6/site-packages/_pytest/mark/structures.py:383: in _check
    raise AttributeError("%r not a registered marker" % (name,))
E   AttributeError: 'integration' not a registered marker
```
I installed `requirements-dev.txt`, do you know what am I missing?

asottile:
Looks like pytest strict mode? The markers are left over from when the test suite was split and don't provide any benefits now. Feel free to remove those :)

(probably check your environment for PYTEST env vars, those are probably turning on strict mode)

asottile:
whichever you want to do is fine with me :) (either this PR or new PR)

tests/languages/all_test.py
```diff
@@ -1,20 +1,34 @@
...
 import inspect
+import sys
 
 import pytest
 
 from pre_commit.languages.all import all_languages
 from pre_commit.languages.all import languages
 
 
+if sys.version_info < (3,):
```
asottile: will probably need `# pragma: no cover`s for this -- elsewhere `six.PY2` is being used, let's use that here as well

tests/languages/all_test.py
```diff
@@ -1,20 +1,34 @@
...
+    setargspec = functools.partial(
+        inspect.ArgSpec, varargs=None, keywords=None, defaults=None,
+    )
+    getargspec = inspect.getargspec
+else:
+    setargspec = functools.partial(
```
asottile: I'd probably call this `ArgSpec` or `make_argspec` personally

asottile:
Thanks for the awesome PR!

Issue #704, opened 2018-02-13
meta `check-useless-excludes` hook does not consider things excluded by `files` filter


asottile:
For instance:

https://github.com/Yelp/pyramid-hypernova/blob/ca166e698897ec46ca129e5dbce8d5ccb8d544ec/.pre-commit-config.yaml#L6-L8

```yaml
    -   id: end-of-file-fixer
        language_version: python3.6
        exclude: ^\.activate\.sh$
```

`end-of-file-fixer` uses `types: [text]` which will exclude the `.activate.sh` symlink

asottile:
Sure! Happy hacktober :)  let me know if you'd like guidance on it!

georgeyk:
From the description: `end-of-file-fixer uses files: [text] which will exclude the .activate.sh symlink`
This actually describes the `types` filter, right ?

asottile:
Yep! Good catch let me fix the description

PR #837, opened 2018-10-02
Fix `check-useless-exclude` to consider types filter

georgeyk:
Fixes https://github.com/pre-commit/pre-commit/issues/704


georgeyk:
Thanks @asottile I'll take a look o/

Issue #691, opened 2018-01-24
Improve platform command line length limits


antonbabenko:
I have a question related to #510 

When I run `pre-commit run` all hooks work as expected (processing just changed files), but `pre-commit run -a` divides a list of files into 5 bulks and run the same hook in parallel executions. This creates unnecessary checks of files which belong to the same directory.

Is it possible to control that certain hooks should not be executed in parallel but rather pass all filenames to it?

Alternatively, is there an environment variable or something like that which says that pre-commit is running a hook for all files?

asottile:
So pre-commit doesn't _yet_ run anything in parallel (as proposed in #510 -- and there will be a way to opt out of it once implemented).

What you're probably seeing is the batching of calls.  Passing all arguments to commands will not work in in all situations (there is a command length limit!).  On unixlike systems, coreutils provides a utility [`xargs`](http://man7.org/linux/man-pages/man1/xargs.1.html) to compute how many arguments to pass to avoid this limit.   pre-commit implements a similar helper but is [extremely timid](https://github.com/pre-commit/pre-commit/blob/51ac0e8bc0a1b61939283254c4bd5189c919f919/pre_commit/xargs.py#L8-L10) due to limits found on windows.

This limit can probably be increased on platforms where the limit is higher (if this can be detected?) and then you'll be much more likely to get one in a single batch.  I'd love a PR that does this :D

chriskuehl:
@antonbabenko no hooks operate in parallel yet; that issue is requesting the functionality, but it hasn't been implemented yet.

It's true that pre-commit may split files across multiple hook invocations if there are too many files, but this happens in serial and is sort of unavoidable (since there are limits on the length of `argv`, see `xargs --show-limits`; though pre-commit is a lot more conservative than what the actual max probably is on your system):

https://github.com/pre-commit/pre-commit/blob/51ac0e8bc0a1b61939283254c4bd5189c919f919/pre_commit/xargs.py#L8-L10

antonbabenko:
Thanks for the explanation. I am fine with the little overhead when running the same hook over the same directory.

asottile:
The code hints here are:

- [`os.sysconf`](https://docs.python.org/3/library/os.html#os.sysconf)
- [`SC_MAX_ARG`](ftp://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_chapter/libc_31.html#SEC648)
[...]

PR #839, opened 2018-10-05
Update xargs.partition with platform information

georgeyk:
Changes how `xargs.partition` computes the command length (including arguments) depending on the plataform.
More specifically, 'win32' uses the amount of characters while posix system uses the byte count.

I'm splitting what was discussed in https://github.com/pre-commit/pre-commit/issues/691 in two parts (the next one implements upper limit length).

Since I have no resource to test this under windows, it might be good to ask for help from other developers/users to help validate the change.

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -17,9 +17,9 @@ def _command_length(*cmd):
     full_cmd = ' '.join(cmd)
 
     # win32 uses the amount of characters, more details at:
-    # https://blogs.msdn.microsoft.com/oldnewthing/20031210-00/?p=41553/
+    # https://github.com/pre-commit/pre-commit/pull/839
     if sys.platform == 'win32':
-        return len(full_cmd)
+        return len(full_cmd.encode('utf-16le')) // 2
 
     return len(full_cmd.encode(sys.getfilesystemencoding()))
 
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -3,6 +3,8 @@ from __future__ import unicode_literals
 
 import sys
 
+import six
+
 from pre_commit import parse_shebang
 from pre_commit.util import cmd_output
 
@@ -19,9 +21,13 @@ def _command_length(*cmd):
     # win32 uses the amount of characters, more details at:
     # https://github.com/pre-commit/pre-commit/pull/839
     if sys.platform == 'win32':
-        return len(full_cmd.encode('utf-16le')) // 2
-
-    return len(full_cmd.encode(sys.getfilesystemencoding()))
+        # the python2.x apis require bytes, we encode as UTF-8
+        if six.PY2:
+            return len(full_cmd.encode('utf-8'))
+        else:
+            return len(full_cmd.encode('utf-16le')) // 2
+    else:
+        return len(full_cmd.encode(sys.getfilesystemencoding()))
 
 
 class ArgumentTooLongError(RuntimeError):
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -22,7 +22,7 @@ def _get_command_length(command, arg):
     if sys.platform == 'win32':
         return len(full_cmd)
 
-    return len(full_cmd.encode(sys.getdefaultencoding()))
+    return len(full_cmd.encode(sys.getfilesystemencoding()))
 
 
 class ArgumentTooLongError(RuntimeError):
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -13,9 +13,8 @@ def _get_platform_max_length():
     return 4 * 1024
 
 
-def _get_command_length(command, arg):
-    parts = command + (arg,)
-    full_cmd = ' '.join(parts)
+def _command_length(*cmd):
+    full_cmd = ' '.join(cmd)
 
     # win32 uses the amount of characters, more details at:
     # https://blogs.msdn.microsoft.com/oldnewthing/20031210-00/?p=41553/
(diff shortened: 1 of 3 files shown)


asottile:
I can check this on windows when I'm at home -- another option is to use the https://modern.ie VMs

georgeyk:
Thanks.
I'll fix the errors reported later and check modern.ie (timeout right now)

pre_commit/xargs.py
```diff
@@ -1,41 +1,54 @@
...
 
 
-def partition(cmd, varargs, _max_length=MAX_LENGTH):
+def partition(cmd, varargs, _max_length=None):
+    _max_length = _max_length or _get_platform_max_length()
     cmd = tuple(cmd)
     ret = []
 
     ret_cmd = []
-    total_len = len(' '.join(cmd))
```
asottile: the original code was an attempt to avoid `O(N^2)` work -- I think this pattern can still be used with something like:

```python
def _command_length(*cmd):
    cmd_s = ' '.join(cmd)
    if sys.platform == 'win32':
        return len(cmd_s)
    else:
        return len(cmd_s.encode(sys.getfilesystemencoding()))

...

    total_len = _command_length(*cmd)

    ...

    if total_len + 1 + _command_length(arg) <= _max_length:
```

asottile:
the last commit won't make a difference to the failing windows test unfortunately (the top of the file has `from __future__ import unicode_literals`)...

This is a wide vs narrow unicode problem as far as I can tell.  On wide python (python 3.x and python2.x when compiled right) `len(u'😑') == 1` -- on narrow: `len(u'😑') == 2`.

It's unclear to me whether this actually matters for the windows apis (I can try locally if you'd like me to)

georgeyk:
Yea, noticed.
It's ok to use half of limit when on win32 + narrow unicode ?
In that case, probably I just need to increase `_max_length=30` (failing test).

asottile:
It looks like this might be more complicated than I expected...

### narrow python2.7 windows

```pycon
>>> x = u'\U0001f611'
>>> len(x)
2
>>> subprocess.check_call(('echo', (x * 8190).encode('UTF-8')))
# prints bytes
>>> len((x * 8190).encode('UTF-8'))
32760
>>> subprocess.check_call(('echo', b'x' + (x * 8190).encode('UTF-8')))
# prints bytes
>>> subprocess.check_call(('echo', b'xx' + (x * 8190).encode('UTF-8')))
...
WindowsError: [Error 87] The parameter is incorrect
>>> subprocess.check_call(('echo', b'xxxx', (x * 8189).encode('UTF-8')))
# prints bytes
>>> subprocess.check_call(('echo', b'xxxxx', (x * 8189).encode('UTF-8')))
...
WindowsError: [Error 87] The parameter is incorrect
```

From this, it's actually bytes again :(

And the bytes we have:

- `echo`: 4
- `b'x' + (x * 8190).encode('UTF-8')`: 32761

total: 32765 (+1 for space +1 for null byte => 32767)

Or

- `echo`: 4
- `b'xxxx'`: 4
- `(x * 8189).encode('UTF-8')`: 32756

total:
32764 (+2 for spaces, +1 for null byte => 32767)


## wide python2.7 windows

hmmm, apparently doesn't exist?

## python3.6 windows

```pycon
>>> x = u'\U0001f611'
>>> len(x)
1
>>> subprocess.check_call(('echo', 'x' + x * 16380))
# prints smiles
>>> subprocess.check_call(('echo', 'xx' + x * 16380))
...
FileNotFoundError: [WinError 206] The filename or extension is too long
>>> subprocess.check_call(('echo', 'x' * 32761))
# prints a bunch of xs
```

Here we get a different answer, and actually I think this still makes sense to me:

- `echo`: 4 characters
- 'x' + x * 16380: 16381 characters (but most of those are emojis)

But, windows apis are UTF16 (really when converting arguments above, pre-commit should probably be using UTF16 maybe?)

to get the length of "characters" according to windows:

```pycon
>>> len(('x' + x * 16380).encode('utf-16le')) // 2
32761
```

which + 4 + 1 + 1 gets us back to 32767

### conclusion

I don't have a conclusion, but I do believe I've opened a can of worms -- I'm really sorry this is so complicated and messy.

georgeyk:
```
But, windows apis are UTF16 (really when converting arguments above, pre-commit should probably be using UTF16 maybe?)
```
Is this true for most of windows versions ? If so, I think it would be ok.
If not, maybe we can find this set on `sys.stdout.encoding` or leave it hardcoded.
Alternatively, we can set 8k for win32. The performance impact seems to be minimal, though.

Another detail, as we discussed in https://github.com/pre-commit/pre-commit/issues/691 we are never going to reach 32k (I mean, limit will be `32k - environ - headroom`).

asottile:
Looking at the options in python2, there doesn't appear to be a good choice for encoding the arguments.  The python C apis require bytes, and any encoding that I picked seems to end in mojibake :( -- it's actually probably fine that we're encoding to UTF-8 there and then querying length.

In python3 on windows though, it looks like we should first `.encode('utf-16le')` and then query `len(...) // 2`.

It also appears that the `environ` headroom requirement doesn't apply on windows 🤷‍♂️

georgeyk:
```It also appears that the environ headroom requirement doesn't apply on windows```
Yes, but I thought it would be ok to apply it anyway, but that's part 2.

For the python2 mojibake, maybe it should be a known limitation, just as it is now.
Python2 will be maintained for ~1y, are you planning to keep supporting it ?

asottile:
Support is tricky, I don't plan to support it in the framework but will continue to allow hooks to be implemented in whatever they want.  I don't think there's actually a _correct_ solution in python2 for passing non-ascii arguments to subprocesses so I'm completely ok with that being a python2 limitation (in short: mojibake fine, crash not fine).

asottile:
Oh shoot I totally forgot about this, I'll try this on windows tonight and hopefully merge!

asottile:
Thanks again for the PR, sorry this was so much more complicated than I thought :S

asottile:

PR #851, opened 2018-10-20
Implement concurrent execution of individual hooks

### LINKED ISSUE #510: Parallel execution of individual hooks?
I have this one hook that runs `puppet parser validate` which is extremely slow. Running it on just 185 files takes 105 seconds which makes my tests (which call `pre-commit run --all-files`, among other things) annoying enough to run that most people don't run them locally.

What if pre-commit's xargs implementation could do things in parallel? Here's a sketch of a patch that does that (only works on Python 3): https://i.fluffy.cc/t43V5vqd3VH9lTQfl8djnfZWBV2zDDTZ.html

This takes my test time from 105 seconds to 15 seconds.

Some thoughts:

* If any hooks write to files besides the ones they are linting, this could break. This is a problem, though pre-commit is heavily designed around operations on individual files, so the vast majority of hooks should be okay. We could offer an opt-in or opt-out at the individual hook level?
* Parallelizing different hooks (running puppet-lint at the same time as puppet-validate) would be neat but I think is way more problematic (and introduces crazy locking problems since you can't have two hooks running on the same file at once).
* Because pre-commit captures output and displays it at the end, I don't think we have any of the usual problems of interleaved/confusing output. The printing happens in the main thread when collecting statuses and shouldn't have races.
* `concurrent.futures` works great here, but is Python 3 only. I'm not sure how hard this is to do in a compatible way.

@asottile what do you think? Is this too complicated to be worth pursuing?

chriskuehl:
Fixes #510

This implements something like the interface described in the ticket. Individual hooks have partitions run in parallel. The current logic is just a sketch (it uses the number of CPUs as the target concurrency, and does not adjust partitioning logic, so partitions are usually very large and the benefit is minimal).

Mainly looking for feedback on the current interface changes. I think it makes sense for `xargs()` to remain a "hook-ignorant" function (i.e. it's basically a generic Python implementation of xargs, and does not know anything about pre-commit or hook configs), so the logic for determining the target concurrency based on a hook's config lives in language helpers instead. This leads to a bit more boilerplate, but I could imagine different languages wanting to optimize this differently anyhow. (For example, I added TODOs about whether it really makes sense to parallelize the `grep` languages.)

Any thoughts?

### TODO

* [x] Smarter partitioning to take advantage of concurrency
* [x] Better algorithm for calculating target concurrency?
* [x] Tests for require_serial / the above

### DIFF
--- a/pre_commit/languages/helpers.py
+++ b/pre_commit/languages/helpers.py
@@ -50,7 +50,7 @@ def no_install(prefix, version, additional_dependencies):
 
 
 def target_concurrency(hook):
-    if hook['require_serial']:
+    if hook['require_serial'] or 'PRE_COMMIT_NO_CONCURRENCY' in os.environ:
         return 1
     else:
         # Travis appears to have a bunch of CPUs, but we can't use them all.
--- a/pre_commit/languages/helpers.py
+++ b/pre_commit/languages/helpers.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import multiprocessing
 import shlex
 
 from pre_commit.util import cmd_output
@@ -45,3 +46,11 @@ def basic_healthy(prefix, language_version):
 
 def no_install(prefix, version, additional_dependencies):
     raise AssertionError('This type is not installable')
+
+
+def target_concurrency(hook):
+    if hook['require_serial']:
+        return 1
+    else:
+        # TODO: something smart!
+        return multiprocessing.cpu_count()
--- a/pre_commit/languages/helpers.py
+++ b/pre_commit/languages/helpers.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import multiprocessing
+import os
 import shlex
 
 from pre_commit.util import cmd_output
@@ -52,5 +53,11 @@ def target_concurrency(hook):
     if hook['require_serial']:
         return 1
     else:
-        # TODO: something smart!
-        return multiprocessing.cpu_count()
+        # Travis appears to have a bunch of CPUs, but we can't use them all.
+        if 'TRAVIS' in os.environ:
+            return 2
+        else:
+            try:
+                return multiprocessing.cpu_count()
+            except NotImplementedError:
+                return 1
--- a/pre_commit/languages/helpers.py
+++ b/pre_commit/languages/helpers.py
@@ -5,6 +5,7 @@ import os
 import shlex
 
 from pre_commit.util import cmd_output
+from pre_commit.xargs import xargs
 
 
 def run_setup_cmd(prefix, cmd):
@@ -61,3 +62,7 @@ def target_concurrency(hook):
                 return multiprocessing.cpu_count()
             except NotImplementedError:
                 return 1
+
+
+def run_xargs(hook, cmd, file_args):
+    return xargs(cmd, file_args, target_concurrency=target_concurrency(hook))
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import
+from __future__ import division
 from __future__ import unicode_literals
 
 import contextlib
+import math
 import multiprocessing.pool
 import sys
 
@@ -37,8 +39,13 @@ class ArgumentTooLongError(RuntimeError):
     pass
 
 
-def partition(cmd, varargs, _max_length=None):
+def partition(cmd, varargs, target_concurrency, _max_length=None):
     _max_length = _max_length or _get_platform_max_length()
+
+    # Generally, we try to partition evenly into at least `target_concurrency`
+    # partitions, but we don't want a bunch of tiny partitions.
+    max_args = max(4, math.ceil(len(varargs) / target_concurrency))
+
     cmd = tuple(cmd)
     ret = []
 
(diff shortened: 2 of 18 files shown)


pre_commit/languages/helpers.py
```diff
@@ -45,3 +46,11 @@ def basic_healthy(prefix, language_version):
...
 def no_install(prefix, version, additional_dependencies):
     raise AssertionError('This type is not installable')
+
+
+def target_concurrency(hook):
+    if hook['require_serial']:
+        return 1
+    else:
+        # TODO: something smart!
+        return multiprocessing.cpu_count()
```
chriskuehl: I don't really know what to do here. Maybe the number of CPUs is actually a decent heuristic? Or the number minus one? Or divided by two???

pre_commit/xargs.py
```diff
@@ -58,10 +72,17 @@ def xargs(cmd, varargs, **kwargs):
     except parse_shebang.ExecutableNotFoundError as e:
         return e.to_output()
 
-    for run_cmd in partition(cmd, varargs, **kwargs):
-        proc_retcode, proc_out, proc_err = cmd_output(
-            *run_cmd, encoding=None, retcode=None
-        )
+    # TODO: teach partition to intelligently target our desired concurrency
+    # while still respecting max_length.
+    partitions = partition(cmd, varargs, **kwargs)
```
chriskuehl: I think the eventual change to partition would have some relatively simple logic where it tries to fill up sets of N partitions (to take advantage of the N concurrency) while respecting the max_length, rather than trying to minimize the number of partitions like it does now. Probably with some simple logic to avoid silly things like 4 partitions of 1 item each which are probably non-optimal due to per-invocation overhead (we can probably just hard-code something reasonable here?).

tests/xargs_test.py
```diff
@@ -75,3 +77,20 @@ def test_xargs_retcode_normal():
...
+    ret, stdout, _ = xargs.xargs(
+        bash_cmd, print_pid * 5,
+        target_concurrency=5,
+        _max_length=len(' '.join(bash_cmd + print_pid)),
+    )
+    elapsed = time.time() - start
+    assert ret == 0
+    pids = stdout.splitlines()
+    assert len(pids) == 5
+    assert elapsed < 1
```
chriskuehl: Kind of a silly test, but asserting stuff is running in parallel is hard. idk, what do you think?

pre_commit/languages/helpers.py
asottile: probably need to do something like this: https://github.com/pytest-dev/pytest-xdist/pull/317/files

pre_commit/languages/pcre.py
```diff
@@ -19,4 +19,11 @@ def run_hook(prefix, hook, file_args):
 
     # Grep usually returns 0 for matches, and nonzero for non-matches so we
     # negate it here.
-    return xargs(cmd, file_args, negate=True)
+    return xargs(
+        cmd,
+        file_args,
+        negate=True,
+        # TODO: does it actually make sense to parallelize this hook, given
+        # it's probably I/O bound?
```
asottile: yeah, probably make this always synchronous is fine -- pcre is going away eventually anyway (for pygrep -- though I assume below I'm about to look at pygrep anyway!)

pre_commit/xargs.py
```diff
@@ -43,12 +52,23 @@ def partition(cmd, varargs, _max_length=MAX_LENGTH):
     return tuple(ret)
 
 
+@contextlib.contextmanager
+def _threadpool(size):
```
chriskuehl: @asottile let me know if you have a preference on doing this vs installing a concurrent.futures backport on python2

asottile: up to you, `concurrent.futures` is slightly more future-proof if the code is simpler (if it isn't then definitely don't bother)

chriskuehl:
Heh, on AppVeyor:

```
================================== FAILURES ===================================
___________________________ test_xargs_concurrency ____________________________
    def test_xargs_concurrency():
        bash_cmd = ('bash', '-c')
        print_pid = ('sleep 0.5 && echo $$',)
    
        start = time.time()
        ret, stdout, _ = xargs.xargs(
            bash_cmd, print_pid * 5,
            target_concurrency=5,
            _max_length=len(' '.join(bash_cmd + print_pid)),
        )
        elapsed = time.time() - start
        assert ret == 0
        pids = stdout.splitlines()
        assert len(pids) == 5
>       assert elapsed < 1
E       assert 1.0390000343322754 < 1
tests\xargs_test.py:124: AssertionError
== 1 failed, 491 passed, 6 skipped, 9 xfailed, 10 xpassed in 529.27 seconds ===
```

It's been passing most of the time on AppVeyor, probably flaky :\

Guess I should increase the elapsed duration... or just delete the test.

asottile:
> Guess I should increase the elapsed duration... or just delete the test.

🔪 😆

chriskuehl:
@asottile switched to a `concurrent.futures` backport and added an env var to force it to run in serial during tests. Let me know what you think!

asottile:
I'm excited and terrified for this feature 🎉 

pushed my cleanups, will merge after CI finishes

alexjurkiewicz:
nice work!!!

asottile:

Issue #856, opened 2018-10-29
autoupdate from a revision to a repo with latest tag missing .pre-commit-hooks.yaml errors


asottile:
It should just skip updating these repositories.

```yaml
repos:
- repo: https://github.com/python-modernize/python-modernize
  rev: a234ce4e1
  hooks:
  - id: python-modernize
```

```console
$ pre-commit  autoupdate
Updating https://github.com/python-modernize/python-modernize...[INFO] Initializing environment for https://github.com/python-modernize/python-modernize.
[INFO] Initializing environment for https://github.com/python-modernize/python-modernize.
An error has occurred: InvalidManifestError: 
=====> /home/asottile/.cache/pre-commit/repolsycg5yp/.pre-commit-hooks.yaml does not exist
Check the log at /home/asottile/.cache/pre-commit/pre-commit.log
```

```console
$ cat ~/.cache/pre-commit/pre-commit.log 
An error has occurred: InvalidManifestError: 
=====> /home/asottile/.cache/pre-commit/repolsycg5yp/.pre-commit-hooks.yaml does not exist
Traceback (most recent call last):
  File "/home/asottile/opt/venv/lib/python3.6/site-packages/cfgv.py", line 46, in reraise_as
    yield
  File "/home/asottile/opt/venv/lib/python3.6/site-packages/cfgv.py", line 344, in load_from_filename
    raise ValidationError('{} does not exist'.format(filename))
cfgv.ValidationError: 
=====> /home/asottile/.cache/pre-commit/repolsycg5yp/.pre-commit-hooks.yaml does not exist

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/asottile/opt/venv/lib/python3.6/site-packages/pre_commit/error_handler.py", line 47, in error_handler
    yield
  File "/home/asottile/opt/venv/lib/python3.6/site-packages/pre_commit/main.py", line 256, in main
    repos=args.repos,
  File "/home/asottile/opt/venv/lib/python3.6/site-packages/pre_commit/commands/autoupdate.py", line 129, in autoupdate
    new_repo_config = _update_repo(repo_config, store, tags_only)
  File "/home/asottile/opt/venv/lib/python3.6/site-packages/pre_commit/commands/autoupdate.py", line 60, in _update_repo
    hooks_missing = hooks - (hooks & set(new_repo.manifest_hooks))
[...]

PR #857, opened 2018-10-31
Reraises InvalidManifestError as RepositoryCannotBeUpdatedError

runz0rd:
For #856

### DIFF
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -11,10 +11,10 @@ from cfgv import remove_defaults
 import pre_commit.constants as C
 from pre_commit import output
 from pre_commit.clientlib import CONFIG_SCHEMA
+from pre_commit.clientlib import InvalidManifestError
 from pre_commit.clientlib import is_local_repo
 from pre_commit.clientlib import is_meta_repo
 from pre_commit.clientlib import load_config
-from pre_commit.clientlib import InvalidManifestError
 from pre_commit.commands.migrate_config import migrate_config
 from pre_commit.repository import Repository
 from pre_commit.util import CalledProcessError
@@ -54,14 +54,15 @@ def _update_repo(repo_config, store, tags_only):
     # Construct a new config with the head rev
     new_config = OrderedDict(repo_config)
     new_config['rev'] = rev
-    new_repo = Repository.create(new_config, store)
+
+    try:
+        new_hooks = Repository.create(new_config, store).manifest_hooks
+    except InvalidManifestError as e:
+        raise RepositoryCannotBeUpdatedError(e.args[0])
 
     # See if any of our hooks were deleted with the new commits
     hooks = {hook['id'] for hook in repo_config['hooks']}
-    try:
-        hooks_missing = hooks - (hooks & set(new_repo.manifest_hooks))
-    except InvalidManifestError as e:
-        raise RepositoryCannotBeUpdatedError(e.args[0])
+    hooks_missing = hooks - set(new_hooks)
     if hooks_missing:
         raise RepositoryCannotBeUpdatedError(
             'Cannot update because the tip of master is missing these hooks:\n'
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -132,7 +132,7 @@ def autoupdate(runner, store, tags_only, repos=()):
         try:
             new_repo_config = _update_repo(repo_config, store, tags_only)
         except RepositoryCannotBeUpdatedError as error:
-            output.write_line(error.args[0])
+            output.write_line(str(error))
             output_repos.append(repo_config)
             retv = 1
             continue
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import re
 from collections import OrderedDict
 
+import six
 from aspy.yaml import ordered_dump
 from aspy.yaml import ordered_load
 from cfgv import remove_defaults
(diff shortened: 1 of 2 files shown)


asottile:
Thanks for the PR! this is great :tada: 

want to add a quick test (or let me know and I can help with this)? here's a [~similar scenario](https://github.com/pre-commit/pre-commit/blob/1f1cd2bc394271540955794b3d2d3e6d8f4e65bb/tests/commands/autoupdate_test.py#L226-L241) though in this case we'll need to _not_ tag the latest commit and instead tag the commit before the `.pre-commit-hooks.yaml` gets introduced.

runz0rd:
Sorry for disappearing. Ill be able to get this done in the next 5 hours or so.

asottile:
basically just a test which verifies the behaviour in the ticket.

I think this covers it?

```diff
diff --git a/tests/commands/autoupdate_test.py b/tests/commands/autoupdate_test.py
index 3bfb62e..41addc6 100644
--- a/tests/commands/autoupdate_test.py
+++ b/tests/commands/autoupdate_test.py
@@ -260,6 +260,22 @@ def test_autoupdate_tags_only(tagged_repo_with_more_commits, in_tmpdir, store):
         assert 'v1.2.3' in f.read()
 
 
+def test_autoupdate_latest_tag_no_config(out_of_date_repo, in_tmpdir, store):
+    cmd_output('git', '-C', out_of_date_repo.path, 'rm', '-r', ':/')
+    cmd_output('git', '-C', out_of_date_repo.path, 'commit', '-m', 'rm')
+    cmd_output('git', '-C', out_of_date_repo.path, 'tag', 'v1.2.3')
+
+    config = make_config_from_repo(
+        out_of_date_repo.path, rev=out_of_date_repo.original_rev,
+    )
+    write_config('.', config)
+
+    ret = autoupdate(Runner('.', C.CONFIG_FILE), store, tags_only=False)
+    assert ret == 0
+    with open(C.CONFIG_FILE) as f:
+        assert out_of_date_repo.original_rev in f.read()
+
+
 @pytest.fixture
 def hook_disappearing_repo(tempdir_factory):
     path = make_repo(tempdir_factory, 'python_hooks_repo')
```

asottile:
ah yeah the bit about tagging, `pre-commit` by default will search for the latest tagged revision on a branch and prefer that over commits past that (unless you use `--bleeding-edge`).  In this case, it's broken whether or not there's newer commits.  Let me adjust that test :)

asottile:
I've updated the patch above :)

runz0rd:
Ah, I think I understand now. Thanks!

runz0rd:
The trouble here is, the `make_config_from_repo` will not generate a config with files missing (`.pre-commit-hooks.yaml` in this case). This happens before `autoupdate`. How is this supposed to be handled?

asottile:
I think you can just take the test I wrote above and you won't need another one.  It fails right now, and I think (?) it passes after your change

runz0rd:
I was just trying it out. It fails because my change only affects `autoupdate` command. `make_config_from_repo` calls the same validation logic:
`pre_commit.clientlib.InvalidManifestError: 
E               =====> /../test_autoupdate_latest_no_conf0/0/.pre-commit-hooks.yaml does not exist
`

runz0rd:
That did it, with a small change. Thanks again!

pre_commit/commands/autoupdate.py
```diff
@@ -128,7 +132,7 @@ def autoupdate(runner, store, tags_only, repos=()):
         try:
             new_repo_config = _update_repo(repo_config, store, tags_only)
         except RepositoryCannotBeUpdatedError as error:
-            output.write_line(error.args[0])
+            output.write_line(str(error))
```
asottile: this should still be `e.args[0]` right?  calling `str()` can lead to encoding problems in python2.x

pre_commit/commands/autoupdate.py
```diff
@@ -57,7 +58,10 @@ def _update_repo(repo_config, store, tags_only):
 
     # See if any of our hooks were deleted with the new commits
     hooks = {hook['id'] for hook in repo_config['hooks']}
-    hooks_missing = hooks - (hooks & set(new_repo.manifest_hooks))
+    try:
+        hooks_missing = hooks - (hooks & set(new_repo.manifest_hooks))
+    except InvalidManifestError as e:
+        raise RepositoryCannotBeUpdatedError(e.args[0])
```
asottile: it makes a bigger diff but it might make more sense to move this up so it reads like this:

```python
try:
    new_hooks = Repository.create(new_config, store).manifest_hooks
except InvalidManifestError as e:
   raise ...

hooks_missing = hooks - set(new_hooks)
```

(note I also adjusted the set logic, not sure why it was `and`ing with `hooks` before, I think the set logic is equivalent now)

pre_commit/commands/autoupdate.py
```diff
@@ -14,6 +14,7 @@
 from pre_commit.clientlib import is_local_repo
 from pre_commit.clientlib import is_meta_repo
 from pre_commit.clientlib import load_config
+from pre_commit.clientlib import InvalidManifestError
```
asottile: `pre-commit install` and then `pre-commit run --all-files` ;) (it'll fix the import ordering issue for you automatically, and enforce it on commits after that!)

pre_commit/commands/autoupdate.py
runz0rd: Yep, but `e.args[0]` was giving me issue with `write_line`, namely `five.to_bytes(s)`. Any suggestions?

runz0rd: After moving the error handling logic up, this is not an issue any more. Strange.

asottile:
Oops! I forgot to click the merge button!

Thanks again for the contribution it is greatly appreciated :) 🎉

asottile:

PR #874, opened 2018-11-20
Shuffle arguments before running hooks

### LINKED ISSUE #861: Concurrent execution results in uneven work per thread
I'm running `pre-commit` from current `master` to test the concurrency feature introduced with #851. While it in general seems to work, work is distributed pretty uneven. One hook we run is [`prospector`](https://github.com/guykisel/prospector-mirror) which is nice for testing, because it takes a relatively long time and it prints the time taken in its output.

Running `pre-commit run -a --verbose prospector | grep "Time Taken"` on a medium sized project (~100 Python files) results in the following distribution of work to the available 4 logical CPU cores:
```
Time Taken: 17.10 seconds
Time Taken: 8.70 seconds
Time Taken: 18.68 seconds
Time Taken: 108.02 seconds
```

Especially compared to running it with concurrency disabled (using `PRE_COMMIT_NO_CONCURRENCY`), it's pretty obvious that concurrency doesn't provide any real benefit here:
```
Time Taken: 116.95 seconds
```

I'd be happy to help debugging this further. Just tell me what other information you need. :slightly_smiling_face:

chriskuehl:
Fixes #861

### DIFF
--- a/pre_commit/languages/helpers.py
+++ b/pre_commit/languages/helpers.py
@@ -2,10 +2,16 @@ from __future__ import unicode_literals
 
 import multiprocessing
 import os
+import random
 import shlex
+
+import six
 
 from pre_commit.util import cmd_output
 from pre_commit.xargs import xargs
+
+
+FIXED_RANDOM_SEED = 1542676186
 
 
 def run_setup_cmd(prefix, cmd):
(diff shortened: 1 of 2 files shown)


pre_commit/languages/helpers.py
```diff
@@ -64,5 +70,21 @@ def target_concurrency(hook):
...
+def _shuffled(seq):
+    """Deterministically shuffle identically under both py2 + py3."""
+    fixed_random = random.Random()
+    if six.PY2:  # pragma: no cover (py2)
+        fixed_random.seed(FIXED_RANDOM_SEED)
+    else:
+        fixed_random.seed(FIXED_RANDOM_SEED, version=1)
+
+    seq = list(seq)
+    random.shuffle(seq, random=fixed_random.random)
```
chriskuehl: You have to do kind of a weird dance to get these to shuffle identically between py2 and py3 -- `fixed_random.shuffle` won't work because in py3 `random.shuffle` picks a different algorithm than what's used in py2 if you don't specify `random` as a kwarg; `random.sample` which would probably be preferred to avoid turning it into a list and mutating it has a similar problem.

We probably don't really need it to be identical between py2 and py3, but it means we don't need to go fix a bunch of tests. Idk, worth it?

tests/commands/install_uninstall_test.py
```diff
@@ -532,7 +532,7 @@ def test_pre_push_integration_failing(tempdir_factory, store):
         assert retc == 1
         assert 'Failing hook' in output
         assert 'Failed' in output
-        assert 'foo zzz' in output  # both filenames should be printed
+        assert 'zzz foo' in output  # both filenames should be printed
```
chriskuehl: lol i messed with the fixed random seed until only one test failed rather than a bunch

asottile: lol nice

chriskuehl: derp, apparently this is actually passing as-is (i needed to `pip install` into my venv again)

Issue #871, opened 2018-11-16
Add config option to ask user to continue anyway


SimonHeimberg:
When the config for some code-style (eslint, php-cs, ...) is not (yet) fine tuned, it would be handier to answer an interactive question than to redo the commit with SET=hook_id.

When there is no tty, the hook would fail.

suggestion for configuration:
```yaml
#.pre-commit-config.yaml
-   repo: https://github.com/pre-commit/pre-commit-hooks
    hooks:
    -   id: some-unreliable-check
        ask_on_fail: true
```

asottile:
[`SKIP=...`](https://pre-commit.com/#temporarily-disabling-hooks) is the intended mechanism for this.

I don't want to add any interactivity to `pre-commit` (not just because it's complex, but I just find tools that are interactive clunky) so I won't be implementing this.

Thanks nonetheless for the issue!

asottile:
Additionally, `git `hooks aren't handed a tty:

```bash
cat > .git/hooks/pre-commit <<EOF
#!/usr/bin/env python
import sys; print(sys.stdin.isatty())
EOF
chmod +x .git/hooks/pre-commit
git commit --allow-empty -m hi
```

```
...
$ git commit --allow-empty -m hi
False
[master 24ce234] hi
```

PR #876, opened 2018-11-28
support IGNORE envoronment variable to ignore failure

SimonHeimberg:


SimonHeimberg:
This is similar to SKIP, but
 * the check is run and
 * shows output if it fails

So the user sees the ignored failure and does not skip anything by accident.
This is a suggestion instead of unhandy #871.

The state `Ignore` is chosen because it has the same length as `Passed` and `Failed`.

SimonHeimberg:
When it makes sense, I can write `.<green>Passed</>`, `.<red>Failed</>` (both with leading dot first) and `<yellow>Ignored</>`. But first let's discuss the idea of this patch.

* Right, currently, IGNORE is not considered for modifying hooks. ([see here in diff](https://github.com/pre-commit/pre-commit/pull/876/files#diff-d58a98662b1c7c5efc1562af39480143R122]))
* Difference between SKIP and IGNORE: IGNORE reports "Passed" or "Ignore" depending on the output. SKIP always reports "Skipped"
* What I do not like on SKIP is that I do not see any output.
    * Am I sure that I only skipped what I wanted?
    * Does this not tempt to set SKIP when I know the file will fail without checking first if it fails on what I expect? (Me does not mean me personally.)
    We currently use [a shell commit hook by me](https://github.com/EmchBerger/cube-common-develop/blob/2a665a5/src/CodeStyle/check-commit-cube.sh) which asks to continue on nonfatal failure. I agree that that tweaking /dev/tty like this is not nice. (Querying on failure was a wish of the team leader.) But questions do not allow to say in advance to skip some parts, only after. (Unfortunately I did no pre-commit (this tool) this tool when writing the bash script...)
    Will likely be hard to convince my team to do two commit commands when expecting a failure, and use SKIP ONLY the 2nd time.

asottile:
What if your hook responded to `IGNORE` instead?  I don't really want yet-another-way to skip things for a usecase that the framework intentionally tries to forbid (interactive hooks) where `SKIP` already serves the purpose of skipping things.  Pressing the up arrow or running `fc` is pretty easy :)

The reason there aren't interactive or high-false-positive hooks is those are the types of things that make people hate git hooks (and turn them off!) -- I don't want to encourage or enable that since it hurts the ecosystem :)

Issue #730, opened 2018-03-20
Add support for requirements.txt in python additional_dependencies argument


jhg03a:
Currently the additional_dependencies parameter for python requires duplicating the contents of an existing requirements.txt.  Instead it should be able to detect that an argument it provided includes *.txt, then [adjust the pip command](https://github.com/pre-commit/pre-commit/blob/577f1093cf62502fbbd81ef770d2e29fba1e9253/pre_commit/languages/python.py#L146) to use the -r argument instead of '.'

asottile:
The problem with `requirements.txt` is it could be infinitely expanded (what if that `requirements.txt` itself contains `-r`?).  How would pre-commit know when to invalidate an environment?

`-r requirements.txt` is also purely a convention.  You could have `-r foo.bar` just as well -- I don't think it makes sense to special case "files ending in .txt".

In general, `requirements.txt` is not supposed to be used for libraries anyway (there shouldn't be duplication in `setup.py` in that case).

Do you have a specific usecase in mind?  I'd like to see something concrete before even trying to think of how this would be implemented :)

asottile:
Actually, I think you can already accomplish what you want with:

```yaml
    additional_dependencies: ['-rrequirements.txt']
```

The first caveat (about cacheability) still stands, though at least it'll be limited to the repository in question.

To validate, I tried the following configuration:

```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v1.2.3
    hooks:
    -   id: trailing-whitespace
        additional_dependencies: [-r, requirements-dev.txt]
        entry: pip freeze --all
        pass_filenames: false
```

And runtime:

```console
$ pre-commit run --all-files --verbose
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks:-r,requirements-dev.txt.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[trailing-whitespace] Trim Trailing Whitespace...........................Passed
hookid: trailing-whitespace

aspy.yaml==1.0.0
attrs==17.4.0
autopep8==1.3.4
cached-property==1.4.0
cfgv==1.0.0
coverage==4.5.1
flake8==3.5.0
identify==1.0.8
mccabe==0.6.1
mock==2.0.0
nodeenv==1.2.0
pbr==3.1.1
pip==9.0.1
pluggy==0.6.0
[...]

PR #880, opened 2018-12-11
Add support for requirements files in additional_dependencies for python

s0undt3ch:


asottile:
Sorry, I know you put a lot of work into this but I'm not interested in supporting it. Please see the discussion in https://github.com/pre-commit/pre-commit/issues/730

pre-commit currently knows nothing about the paths in additional_dependencies and I'd like to keep it that way. A solution which parses those or parses requirements.txt would therefore be out.

s0undt3ch:
I have read most of that and linked issues, and, honestly, nothing dramatically changes by supporting requirements files.
I know this implementation also takes into account options, but that could be stripped out.

Once the requirements files are pre-processed, pre-commit works exactly as it did before, detecting changes requiring re-installation, invalidating cache, etc.

The idea for us to start using pre-commit is to remove env setup from developers required tasks and also to make sure everyone runs the exact same setup, so, your local or system suggestion is not actually a suggestion to us.

Honestly, I'm not seeing why the tool needs to be shortsighted...
Got any more details as to why translating a requirements file into a requirements list is such a blocker?

asottile:
there's a lot of reasons, I'll try and get to them all but I've gotta get back to work so this might be incomplete

first and foremost is a matter of simplicity.  "simple is better than complex".  the two functions you added in this PR would be the most complex in the entire codebase.  verifying their correctness is nearly impossible given the sheer amount of branches and possibilities

```
pre_commit/languages/python.py:82:1: C901 '_parse_requirements_file' is too complex (19)
pre_commit/languages/python.py:139:1: C901 'collect_requirements' is too complex (28)
```

but also "simple is better than complex" in another regard.  pre-commit aims to be a simple bridge between tool interfaces.  It currently leaves most of the heavy lifting for installation / etc. to the language specific tools themselves.  It knows very little in this regard other than "call the tool".

next is a matter of caching.  caching is already a very difficult topic (even in pre-commit) and complicating it further would make it worse.  today, the cache has a very nice quality in that the `.pre-commit-config.yaml` contents are more than enough to represent the cache.  If that file doesn't change, the cache doesn't change.  parsing `-r` adds a fractal of possibilities in other files (especially if you consider that requirements files themselves can `-r` and then where does it end!).

next is "reinventing the wheel, poorly".  in this PR we're reinventing both (1) pip's argument parsing and (2) pip's parsing of `-r` requirements files.  pip's argument parsing changes without our control and without a way to verify we are compliant with it.  and I _really_ don't want to keep up with the pip project and adjust at every little change to their parsing, it simply isn't feasible to keep them in lockstep and any differences are going to be considered bugs by users -- bugs that I don't want to deal with.  For starters, I already see at least two issues with the current implementation that aren't handled: continued options (`-vrr.txt` and `-vr r.txt`) and prefix options (`--requirem=r.txt`) -- both of which are **really really hard** to get right!  And that doesn't even begin to handle the slough of bugs in the actual .txt file parsing!  Additionally, requirements.txt files don't even have a proper specification -- they're a mix of pip options and requirement specs (and depending on your version of pip, even that isn't specified!).

next is the "slippery slope".  while pre-commit itself is written in python, python is not the only supported language (though it may be the most widely used).  If this PR is accepted, I could reasonably see the need to implement the same complexity for pipfile, pipfile.lock, pyproject.toml, setup.py (intractable / impossible), package.json, package-lock.json, yarn.lock, gemspec, gemfile.lock, glide.yaml, glide.lock, Godeps.json, Gopkg.toml, Gopkg.lock, Cargo.toml, Package.swift, and the list goes on and on and on and on.  the "parsing package files" code would dwarf the rest of the project and they would all be buggy, all of them.

and finally (but probably the least of the problems) is the breaking change in behaviour.  "documented" [here](https://github.com/pre-commit/pre-commit/issues/730#issuecomment-374459786) there's already a way that `-r` can have meaning today *without* `pre-commit` parsing it.  I know of a nonzero number of things already using this.

I hope this helps you understand my reservation to this change -- if you have additional questions please let me know!

There's also been an eventual hope that a strong enough usecase would emerge to take the environment management code out of pre-commit and open it as a library.  There's some established value of a cross platform tool which knows how to set up a bunch of different language'd environments.

Additionally, you can probably accomplish what you want _today_ if you're willing to script a bit.  Write a thing which processes your `requirements.txt` file and adds that to the `additional_dependencies` list in `.pre-commit-config.yaml` (we're doing something similar at lyft even to better join PEP561 requirements for mypy).

asottile:
thanks for the PR nonetheless 🎉 -- it was good to get these ideas more concretely into issues and comments than stuck in my head!

s0undt3ch:
Does pre-commit have a FAQ or a recepies section?
I don't mind sharing my solution to avoid having you repeat yourself or users like me to just do the work without knowing the reasons why its not supported.
Eventually I might find the time to release it as a pre-commit hook.

asottile:
Currently not really -- there's a few "common" usecases at the bottom of https://pre-commit.com (which is sourced from https://github.com/pre-commit/pre-commit.github.io)

paw-lu:
Hi @s0undt3ch ! Curious about what your solution to this was!

s0undt3ch:
I had 2 iterations.
The first, a hook which detected changes in the requirements files and updated pre-commit config if they change. Don't know where I have that code.
The second, we use nox(kind of like tox on steroids) and created a hook which basically calls the right nox target. I'm this scenario, nox will take care of dependencies and their constant changing scenario. We actually tell nox not to create a virtualenv(pre-commit did that for us) and it just installs requirements and runs the target code.
https://github.com/saltstack/salt/blob/master/noxfile.py#L964
https://github.com/saltstack/salt/blob/master/.pre-commit-config.yaml#L605

paw-lu:
Thank you so much @s0undt3ch ! I'm familiar with nox, and this is a great use of it!

So your workaround was simply to just call nox—[which reads from to your requirements file](https://github.com/saltstack/salt/blob/51dc125d7634bdad6ea6dbdecd948a369ba98d5d/noxfile.py#L856-L857)—and handles all of your linting.

Great solution! Thanks for the explanation and the links. It's extremely appreciated and made it easy to understand.

s0undt3ch:
Anytime!

PR #883, opened 2018-12-13
Switch back to `diff` instead of `diff-index`

### LINKED ISSUE #881: "git add -N" prevents restoring stashed changes.
To reproduce, start with this simple `pre-commit-config.yaml` in an otherwise empty repo:

```yaml
repos:
  rev: master
  hooks: 
    - id: end-of-file-fixer
```

The hook used doesn't really matter. end-of-file-fixer is just an example. 

Run the following:
```bash
echo "new" > newfile
echo "\n\n\n" > needs-fixing
git add -N newfile
# newfile is now staged as an empty file
git add needs-fixing
git commit -m "fix"
```

The following output is generated: 

```
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /home/henniss/.cache/pre-commit/patch1544663784.
Fix End of Files.........................................................Failed
hookid: end-of-file-fixer

Files were modified by this hook. Additional output:

Fixing needs-fixing

[WARNING] Stashed changes conflicted with hook auto-fixes... Rolling back fixes...
An unexpected error has occurred: CalledProcessError: Command: ('/usr/lib/git-core/git', '-c', 'core.autocrlf=false', 'apply', '--whitespace=nowarn', '/home/henniss/.cache/pre-commit/patch1544663784')
Return code: 1
Expected return code: 0
Output: (none)
Errors:
    error: newfile: already exists in working directory


Check the log at /home/henniss/.cache/pre-commit/pre-commit.log
```

`cat newfile` now shows that it is empty. The unstaged changes aren't restored.

asottile:
`diff-index` appears to act strangely with `git add --intent-to-add`

Resolves #881 

Essentially a revert of #571

### DIFF
--- a/pre_commit/staged_files_only.py
+++ b/pre_commit/staged_files_only.py
@@ -29,10 +29,9 @@ def staged_files_only(patch_dir):
     context.
     """
     # Determine if there are unstaged files
-    tree = cmd_output('git', 'write-tree')[1].strip()
     retcode, diff_stdout_binary, _ = cmd_output(
-        'git', 'diff-index', '--ignore-submodules', '--binary',
-        '--exit-code', '--no-color', '--no-ext-diff', tree, '--',
+        'git', 'diff', '--ignore-submodules', '--binary',
+        '--exit-code', '--no-color', '--no-ext-diff', '--',
         retcode=None,
         encoding=None,
     )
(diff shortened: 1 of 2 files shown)


asottile:
I still need to follow up to the `git` mailing list about this one, seems like a bug or two in `git`

asottile:
seems this is not so simple after all. looks like the behaviour changes based on git version.  Something more sophisticated needs to happen I guess.

It looks like `--intent-to-add` doesn't _actually_ stage an empty file, and the various diff commands disagree on that.  I'll look into a way to not list that file and not lint that file (since it's not actually being committed without `git commit -a`)

PR #886, opened 2018-12-14
Allow aliasing a hook and calling it by it's alias

s0undt3ch:
This is the least intrusive change.

We could consider using the alias in the output instead of the ID, but I'll defer any more directions on this PR through it's review comments.

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -29,6 +29,20 @@ def _make_argparser(filenames_help):
     return parser
 
 
+class OptionalAlias(object):
+
+    def check(self, dct):
+        if 'alias' in dct:
+            cfgv.check_string(dct['alias'])
+
+    def apply_default(self, dct):
+        if 'alias' not in dct:
+            dct['alias'] = dct['id']
+
+    def remove_default(self, dct):
+        pass
+
+
 MANIFEST_HOOK_DICT = cfgv.Map(
     'Hook', 'id',
 
@@ -36,6 +50,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
     cfgv.Required('name', cfgv.check_string),
     cfgv.Required('entry', cfgv.check_string),
     cfgv.Required('language', cfgv.check_one_of(all_languages)),
+    cfgv.OptionalNoDefault('alias', cfgv.check_string),
 
     cfgv.Optional(
         'files', cfgv.check_and(cfgv.check_string, cfgv.check_regex), '',
@@ -125,6 +140,7 @@ CONFIG_HOOK_DICT = cfgv.Map(
     'Hook', 'id',
 
     cfgv.Required('id', cfgv.check_string),
+    OptionalAlias(),
 
     # All keys in manifest hook dict are valid in a config hook dict, but
     # are optional.
@@ -133,7 +149,7 @@ CONFIG_HOOK_DICT = cfgv.Map(
     *[
         cfgv.OptionalNoDefault(item.key, item.check_fn)
         for item in MANIFEST_HOOK_DICT.items
-        if item.key != 'id'
+        if item.key not in ('id', 'alias')
     ]
 )
 CONFIG_REPO_DICT = cfgv.Map(
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -29,20 +29,6 @@ def _make_argparser(filenames_help):
     return parser
 
 
-class OptionalAlias(object):
-
-    def check(self, dct):
-        if 'alias' in dct:
-            cfgv.check_string(dct['alias'])
-
-    def apply_default(self, dct):
-        if 'alias' not in dct:
-            dct['alias'] = dct['id']
-
-    def remove_default(self, dct):
-        pass
-
-
 MANIFEST_HOOK_DICT = cfgv.Map(
     'Hook', 'id',
 
@@ -50,7 +36,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
     cfgv.Required('name', cfgv.check_string),
     cfgv.Required('entry', cfgv.check_string),
     cfgv.Required('language', cfgv.check_one_of(all_languages)),
-    cfgv.OptionalNoDefault('alias', cfgv.check_string),
+    cfgv.Optional('alias', cfgv.check_string, ''),
 
     cfgv.Optional(
         'files', cfgv.check_and(cfgv.check_string, cfgv.check_regex), '',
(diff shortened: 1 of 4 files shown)


s0undt3ch:
Refs #886

pre_commit/clientlib.py
```diff
@@ -133,7 +135,7 @@ def remove_default(self, dct):
     *[
         cfgv.OptionalNoDefault(item.key, item.check_fn)
         for item in MANIFEST_HOOK_DICT.items
-        if item.key != 'id'
+        if item.key not in('id', 'alias')
     ]
```
asottile: shouldn't need any changes to `CONFIG_HOOK_DICT`

pre_commit/clientlib.py
```diff
@@ -37,6 +37,7 @@ def _make_argparser(filenames_help):
     cfgv.Required('entry', cfgv.check_string),
     cfgv.Required('language', cfgv.check_one_of(all_languages)),
 
+    cfgv.Optional('alias', cfgv.check_string, ''),
```
asottile: default this to `id`, then you can remove the `or` logic below and only use `alias`.

you'll need to write a custom validator -- you can see `cfgv` for examples or `MigrateShaToRev` below

s0undt3ch:
> I reviewed this, it's simple enough, but I'm still not convinced it's necessary if we already have a way to do this that works

We'll, we don't really support this, **unless** you call it twice, with **different** config file as you suggested on the issue filed.

Plus, it's a really simple change :)

I'll address your other review comments and update the PR accordingly.

s0undt3ch:
Updated. Is this more what you had in mind?

pre_commit/clientlib.py
s0undt3ch: My `OptionalAlias` won't a `key` not a `check_fn`, that's why I left the `key not in ()`

s0undt3ch:
hmm, if I follow your suggestion, then tests like `test_does_not_reformat` fail, because, now, there's a `alias`  key which defaults to the ID. :\

s0undt3ch:
Polite ping.

Yeah, I know it's the holidays and it can surely wait until next year :)

pre_commit/commands/run.py
```diff
@@ -255,13 +255,17 @@ def run(runner, store, args, environ=os.environ):
         for repo in repositories(runner.config, store):
             for _, hook in repo.hooks:
                 if (
-                    (not args.hook or hook['id'] == args.hook) and
+                    (not args.hook or hook['id'] == args.hook or (
+                        hook['alias'] and hook['alias'] == args.hook
```
asottile: the first part of this conditional is unnecessary because we check `not args.hook` above.

I'd write this as (emphasis on indentation here)

```python
            if (
                    (
                        not args.hook or
                        hook['id'] == args.hook or
                        hook['alias'] == args.hook
                    ) and
                    (not hook['stages'] or args.hook_stage in hook['stages'])
            ):
```

pre_commit/commands/run.py
```diff
@@ -255,13 +255,17 @@ def run(runner, store, args, environ=os.environ):
...
+                    )) and
                     (not hook['stages'] or args.hook_stage in hook['stages'])
                 ):
                     repo_hooks.append((repo, hook))
 
         if args.hook and not repo_hooks:
-            output.write_line('No hook with id `{}`'.format(args.hook))
+            output.write_line(
+                'No hook with id or alias `{}`'.format(args.hook),
+            )
```
asottile: I wouldn't change this messaging

asottile:
I didn't review because the tests are failing :P

pre_commit/commands/run.py
s0undt3ch: Fixed.

pre_commit/commands/run.py
s0undt3ch: Fixed.

s0undt3ch:
> SKIP is not handled properly, it's also unclear the "right" way SKIP should be handled -- should it exclude both the things?
It only makes sense for pre-commit to also skip based on a hook alias. Added support and test.

pre_commit/commands/run.py
```diff
@@ -85,6 +85,15 @@ def _run_single_hook(filenames, hook, repo, args, skips, cols):
             cols=cols,
         ))
         return 0
+    elif hook['alias'] and hook['alias'] in skips:
```
asottile: combine this with the condition above

tests/commands/run_test.py
```diff
@@ -390,6 +390,38 @@ def test_skip_hook(cap_out, store, repo_with_passing_hook):
...
 
 
+def test_skip_aliased_hook(cap_out, store, repo_with_passing_hook):
+    with cwd(repo_with_passing_hook):
+        # Add bash hook on there again, aliased
+        with modify_config() as config:
+            config['repos'][0]['hooks'].append(
+                {'id': 'bash_hook', 'alias': 'foo_bash'},
+            )
+        stage_a_file()
```
asottile: could maybe make an `aliased_repo` fixture and use it in both tests

I think the code is fairly self evident here, I'd drop the comment

pre_commit/commands/run.py
s0undt3ch: Fixed.

tests/commands/run_test.py
s0undt3ch: Done.

pre_commit/commands/run.py
```diff
@@ -77,7 +77,7 @@ def _run_single_hook(filenames, hook, repo, args, skips, cols):
             'replacement.'.format(hook['id'], repo.repo_config['repo']),
         )
 
-    if hook['id'] in skips:
+    if hook['id'] in skips or (hook['alias'] and hook['alias'] in skips):
```
asottile: if skips can never contain the empty string I think you can drop the first part of this condition?

if skips can contain the empty string, let's fix skips

s0undt3ch: The condition here is that, alias, by default, is an empty string...

s0undt3ch: hmm, ok, but skips can't be empty strings. OK. I get it.

s0undt3ch: fixed.

s0undt3ch:
Thanks!!!

asottile:
thanks @s0undt3ch for bearing with me :tada: 

could you please add this information to the documentation? It'll land in the next minor release so you can use `_new in 1.14.0_`

The docs live at https://github.com/pre-commit/pre-commit.github.io

s0undt3ch:
No problem. https://github.com/pre-commit/pre-commit.github.io/pull/196

asottile:
I'll be trying to get some medium sized things in before the 7th while I have some extra free time and can't stream - - there's a chance I'll release then but otherwise expect something early January

PR #889, opened 2018-12-19
Switch from deprecated docs-off args to --no-document

asottile:
this will break old versions of `gem`, but without it it breaks new versions of gem 🤷‍♂️


asottile:

PR #893, opened 2018-12-25
Pick a better python shebang for hook executable

### LINKED ISSUE #878: Incorrect shebang in .git/hooks/pre-commit for python3 only installations
The shebang for `.git/hooks/pre-commit` is `#!/usr/bin/env python`. I work with setups where `python3` is the only python in env.

Could the shebang be the install python instead? I.e. the installation under `INSTALL_PYTHON = '/usr/bin/python3'`

asottile:
Resolves #878

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -8,6 +8,7 @@ import sys
 
 from pre_commit import git
 from pre_commit import output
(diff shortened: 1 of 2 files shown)


asottile:
This worked in the few environments I had access to -- I suspect it's probably good enough!  Going to merge this

PR #924, opened 2019-01-27
Ensure that GOBIN is unset when installing a golang hook

ashanbrown:
If GOBIN is set, it will be used as the install path instead of the first item from GOPATH followed by "/bin".  If it is used, commands will not be isolated between different repos.

### DIFF
--- a/pre_commit/languages/golang.py
+++ b/pre_commit/languages/golang.py
@@ -7,6 +7,7 @@ import sys
 import pre_commit.constants as C
 from pre_commit import git
 from pre_commit.envcontext import envcontext
+from pre_commit.envcontext import UNSET
 from pre_commit.envcontext import Var
 from pre_commit.languages import helpers
 from pre_commit.util import clean_path_on_failure
@@ -21,6 +22,7 @@ healthy = helpers.basic_healthy
 
 def get_env_patch(venv):
     return (
+        ('GOBIN', UNSET),
         ('PATH', (os.path.join(venv, 'bin'), os.pathsep, Var('PATH'))),
     )
 
@@ -69,6 +71,7 @@ def install_environment(prefix, version, additional_dependencies):
         else:
             gopath = directory
         env = dict(os.environ, GOPATH=gopath)
+        env.pop('GOBIN', None)
         cmd_output('go', 'get', './...', cwd=repo_src_dir, env=env)
         for dependency in additional_dependencies:
             cmd_output('go', 'get', dependency, cwd=repo_src_dir, env=env)
--- a/pre_commit/languages/golang.py
+++ b/pre_commit/languages/golang.py
@@ -7,7 +7,6 @@ import sys
 import pre_commit.constants as C
 from pre_commit import git
 from pre_commit.envcontext import envcontext
-from pre_commit.envcontext import UNSET
 from pre_commit.envcontext import Var
 from pre_commit.languages import helpers
 from pre_commit.util import clean_path_on_failure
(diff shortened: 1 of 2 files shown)


asottile:
I imagine there's other GO env variables to look out for too? Should this unset all of them (or perhaps anything starting with GO?).

ashanbrown:
I think GOBIN might be the only one we need to be concerned about.  GOROOT points to the go installation but pre-commit doesn't install go so I don't think it should touch it.  GOBIN, GOROOT and GOPATH are the three that get billing at https://www.programming-books.io/essential/go/d6da4b8481f94757bae43be1fdfa9e73-gopath-goroot-gobin.

asottile:
Sounds good! Thanks for the research and patch! I can help write a test unless you want to take a stab at it. The current tests live in tests/repository_test.py

pre_commit/languages/golang.py
```diff
@@ -69,6 +71,7 @@ def install_environment(prefix, version, additional_dependencies):
         else:
             gopath = directory
         env = dict(os.environ, GOPATH=gopath)
+        env.pop('GOBIN', None)
```
ashanbrown: I missed this the first time.  The test found this.  ;)

ashanbrown:
@asottile I've added a test and updated the code to ensure GOBIN is also cleared when the command is installed.

pre_commit/languages/golang.py
```diff
@@ -21,6 +22,7 @@
 
 def get_env_patch(venv):
     return (
+        ('GOBIN', UNSET),
```
asottile: is this necessary now? I'd expect this env var to not matter at runtime

tests/repository_test.py
```diff
@@ -267,6 +268,16 @@ def test_golang_hook(tempdir_factory, store):
     )
 
 
+def test_golang_hook_still_works_when_gobin_is_set(tempdir_factory, store):
+    gobin_dir = tempdir_factory.get()
+    with envcontext([('GOBIN', gobin_dir)]):
+        _test_hook_repo(
```
asottile: can even just call `test_golang_hook(tempdir_factory, store)`

tests/repository_test.py
```diff
@@ -267,6 +268,16 @@ def test_golang_hook(tempdir_factory, store):
...
+            tempdir_factory, store, 'golang_hooks_repo',
+            'golang-hook', [], b'hello world\n',
+        )
+    assert os.listdir(gobin_dir) == [], "hook must not be installed in $GOBIN"
```
asottile: no need for 2-arg `assert`, the pytest error message is good enough

tests/repository_test.py
```diff
@@ -71,7 +72,7 @@ def _test_hook_repo(
     path = make_repo(tempdir_factory, repo_path)
     config = make_config_from_repo(path, **(config_kwargs or {}))
     ret = _get_hook(config, store, hook_id).run(args)
-    assert ret[0] == expected_return_code
+    assert ret[0] == expected_return_code, "output was: {}".format(ret[1])
```
asottile: I'd rather keep this as it was, right now this clobbers the pytest message

pre_commit/languages/golang.py
ashanbrown: It doesn't appear to be necessary.  Presumably the new entry in $PATH should supercede whatever is in $GOBIN.

asottile:
Thanks! :tada:

PR #935, opened 2019-02-03
Use pkg_resources instead of importlib_metadata to get installed version

sloria:
Removes importlib-metadata as a dependency.
pkg_resources ships with setuptools, so it should be available
in every reasonably-modern Python environment.

### DIFF
--- a/pre_commit/constants.py
+++ b/pre_commit/constants.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import importlib_metadata  # TODO: importlib.metadata py38?
+import pkg_resources
 
 CONFIG_FILE = '.pre-commit-config.yaml'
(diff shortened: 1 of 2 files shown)


asottile:
this would reverse #846 which intentionally factored out setuptools for performance reasons

sloria:
Ah ok, sorry, I missed that. 

I was considering replacing the [find_version hack](https://github.com/marshmallow-code/marshmallow/blob/880497912c7fe2e92d2dae43eb6b1b436e359813/setup.py#L24-L38) with `str(pkg_resources.get_distribution('...').parsed_version)` in my projects, but I'll hold off now that I know about the performance issues.

asottile:
`importlib-metadata` is supposedly the way of the future, I wasn't following too closely but there were talks of it making it into python3.8, though probably now python3.9

sloria:
Good to know. Thanks!

PR #940, opened 2019-02-06
Add osx specific limit for command-line length

ashanbrown:
Limit on OSX is 262144

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -15,6 +15,8 @@ from pre_commit.util import cmd_output
 
 # TODO: properly compute max_length value
 def _get_platform_max_length():
+    if sys.platform == "darwin":
+        return 262144
     # posix minimum
     return 4 * 1024
 
--- a/tests/xargs_test.py
+++ b/tests/xargs_test.py
@@ -36,6 +36,13 @@ def linux_mock():
             yield
 
 
+@pytest.fixture
+def osx_mock():
+    with mock.patch.object(sys, 'getfilesystemencoding', return_value='utf-8'):
+        with mock.patch.object(sys, 'platform', 'darwin'):
+            yield
+
+
 def test_partition_trivial():
     assert xargs.partition(('cmd',), (), 1) == (('cmd',),)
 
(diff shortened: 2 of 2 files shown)


asottile:
As discussed in #691, this should use `SC_ARG_MAX`: https://github.com/pre-commit/pre-commit/issues/691#issuecomment-360224845

asottile:
and even with that, it appears to take other things into account:

```python
import subprocess


def ok(s):
    try:
        subprocess.check_call(('echo', s), stdout=subprocess.DEVNULL)
    except OSError:
        return False
    else:
        return True



i = 1
while ok('x' * i):
    print(i)
    i *= 2

left = i // 2
right = i

while left < right:
    pivot = left + (right - left) // 2
    print(pivot)
    if not ok('x' * pivot):
        right = pivot
        print(f'right => {right}')
    else:
        left = pivot + 1
        print(f'left => {left}')

print(left)
```

```console
asottile@Macbook-Pro:/private/tmp$ uname -a
Darwin Macbook-Pro 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64 i386 MacBookPro11,4 Darwin
asottile@Macbook-Pro:/private/tmp$ python3 t.py 
1
2
4
8
16
32
64
128
256
512
1024
2048
4096
8192
16384
32768
65536
131072
196608
left => 196609
229376
left => 229377
245760
left => 245761
253952
left => 253953
258048
left => 258049
260096
left => 260097
261120
right => 261120
260608
right => 260608
260352
right => 260352
260224
left => 260225
260288
right => 260288
260256
left => 260257
260272
left => 260273
260280
left => 260281
260284
right => 260284
260282
left => 260283
260283
left => 260284
260284
```

ashanbrown:
Thanks for doing the additional research.  Would you be open to 260284?

I can outline the use case.  I'm passing in filenames but converting them to the directories that contain them in a script hook because I'm running some commands that I want to run per-directory rather than per-file.  With a small command-line, I end up running the command on a directory many times.

FWIW, ideally I wouldn't be passing the filenames in at all when they exceed a certain length.  A lot of this comes about because I'm trying to run the same pre-commit commands on CI that I run locally and I take advantage of how per-commit can pin the version of golangci-lint.

asottile:
that number I empirically determined isn't quite right either -- it's dynamic based on the contents of `os.environ` -- the rough algorithm is laid out in https://github.com/pre-commit/pre-commit/issues/691#issuecomment-360224845 (basically follow what `xargs` does)

ashanbrown:
Got it.  Would you prefer a more involved  algorithm or that we just pick a “safe” number like 128k?  For my purposes it doesn’t matter: 128k is orders of magnitude larger than 4k and would work great.

asottile:
I think we'd try and be as close to gnu `xargs` as possible, fortunately computing the size of `os.environ` and asking `os.sysconf` isn't too difficult.  Even 128k might be too high on some platforms so we have to be careful there

asottile:
I'm finishing this in #966 -- can you try it out on macos and see if it works well for you?

asottile:
via #966

Thanks for the PR 🎉

PR #958, opened 2019-03-03
Allow shallow cloning

### LINKED ISSUE #918: Investigate / implement shallow cloning
Now that github supports it, might be worth the speed improvements

It would have to be smart enough to fall back to slow clone when the option isn't available (either the client is too old or the server is too old)

DanielChabrowski:
Closes #918 

Needs some cleaning up, especially the name "_perform_safe_clone" doesn't really represent the function's purpose. I'm open for suggestions.

### DIFF
--- a/pre_commit/store.py
+++ b/pre_commit/store.py
@@ -154,6 +154,9 @@ class Store(object):
     def clone(self, repo, ref, deps=()):
         """Clone the given url and checkout the specific ref."""
 
+        if os.path.isdir(repo):
+            repo = os.path.abspath(repo)
+
         def clone_strategy(directory):
             env = git.no_git_env()
 
--- a/pre_commit/store.py
+++ b/pre_commit/store.py
@@ -142,7 +142,7 @@ class Store(object):
         git_cmd('checkout', ref)
         git_cmd('submodule', 'update', '--init', '--recursive')
 
-    def _shallow_clone(self, ref, git_cmd):
+    def _shallow_clone(self, ref, git_cmd):  # pragma: windows no cover
         """Perform a shallow clone of a repository and its submodules """
 
         git_config = 'protocol.version=2'
--- a/pre_commit/store.py
+++ b/pre_commit/store.py
@@ -10,6 +10,7 @@ import tempfile
 import pre_commit.constants as C
 from pre_commit import file_lock
 from pre_commit import git
+from pre_commit.util import CalledProcessError
 from pre_commit.util import clean_path_on_failure
 from pre_commit.util import cmd_output
 from pre_commit.util import resource_text
@@ -121,10 +122,7 @@ class Store(object):
                 return result
 
             logger.info('Initializing environment for {}.'.format(repo))
-
-            directory = tempfile.mkdtemp(prefix='repo', dir=self.directory)
-            with clean_path_on_failure(directory):
-                make_strategy(directory)
+            directory = make_strategy()
 
             # Update our db with the created repo
             with self.connect() as db:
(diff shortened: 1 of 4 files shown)


DanielChabrowski:
:confetti_ball: It looks like I cannot ignore those cargo/npm/go tests. I'll check it tomorrow.

pre_commit/store.py
```diff
@@ -121,10 +122,7 @@ def _get_result():
                 return result
 
             logger.info('Initializing environment for {}.'.format(repo))
-
-            directory = tempfile.mkdtemp(prefix='repo', dir=self.directory)
-            with clean_path_on_failure(directory):
-                make_strategy(directory)
```
asottile: I think you can put this back, cloning doesn't get involved for the `local` repositories (and this should happen anyway if the `make_strategy` raises)

pre_commit/store.py
```diff
@@ -134,19 +132,50 @@ def _get_result():
...
+    def _shallow_clone(self, repo, ref, directory):
+        """Perform a shallow clone of a repository and its submodules """
+        env = git.no_git_env()
+
+        def _git_cmd(*args):
+            return cmd_output('git', *args, cwd=directory, env=env)
+
+        _git_cmd('init', '.')
+        _git_cmd('remote', 'add', 'origin', repo)
+        _git_cmd('fetch', 'origin', ref, '--depth=1')
```
asottile: I think there's actually three strategies we have to try here:

1. fetch --depth=1
2. set the git protocol to v2 and then fetch --depth=1 (? or maybe just try this first since it's the most reliable for github)
3. fall back to full clone

pre_commit/store.py
DanielChabrowski: It is happening, I added "_perform_safe_clone" to make_local as well. The problem was that when you shallow clone (git init), you cannot reuse the directory for complete clone because of `fatal: destination path 'path' already exists and is not an empty directory.`. If I just created another directory in fallback, we would save invalid directory to the database so make_strategy had to return the directory that's going to be finally used.

I could change the make_strategy to a list and loop through it here. We wouldn't have to deal with error handling in make_strategy functions and multiple fallbacks would be possible (like the proposed 3). What do you think?

pre_commit/store.py
DanielChabrowski: I'll add `git -c protocol.version=2` to the `_git_cmd` for the second option.

pre_commit/store.py
asottile: another "equivalent" to full clone can reuse that though:

```bash
# non-shallow clone without using `git clone`
git init x
cd x
git remote add origin ...
git fetch origin  # this is all that changes between shallow / non-shallow clone
git checkout ...
```

asottile:
ok so I fixed the npm problem by switching the install strategy

and I played with this a little bit on my own and I have a few concerns but I think we can solve them

- The first is that `autoupdate` uses the cloned repository to `git fetch` and look for newer versions.  This _seems_ fine and I can't think of a case where this wouldn't work like I want it to -- it seems to fetch forward from the shallow clone and still succeeds at updating?
- the `test_clone` test fails on appveyor (which is running `git 2.17.1`) because shallow clone isn't implemented there
- it also fails on my computer (ubuntu 18.04) which coincidentally also runs git 2.17.1
- shallow clone of submodules can leave submodules in a permanently non-clonable state
    - I *think* this means we shouldn't attempt the protocol=1 clone

      ```console
      $ git submodule update --init --depth=1
      Submodule 'fastahack' (https://github.com/ekg/fastahack.git) registered for path 'fastahack'
      Submodule 'filevercmp' (https://github.com/ekg/filevercmp.git) registered for path 'filevercmp'
      Submodule 'fsom' (https://github.com/ekg/fsom.git) registered for path 'fsom'
      Submodule 'googletest' (https://github.com/google/googletest.git) registered for path 'googletest'
      Submodule 'intervaltree' (https://github.com/ekg/intervaltree.git) registered for path 'intervaltree'
      Submodule 'libVCFH' (https://github.com/edawson/libVCFH.git) registered for path 'libVCFH'
      Submodule 'multichoose' (https://github.com/ekg/multichoose.git) registered for path 'multichoose'
      Submodule 'smithwaterman' (https://github.com/ekg/smithwaterman.git) registered for path 'smithwaterman'
      Submodule 'tabixpp' (https://github.com/ekg/tabixpp.git) registered for path 'tabixpp'
      Cloning into '/tmp/x/vcflib/fastahack'...
      Cloning into '/tmp/x/vcflib/filevercmp'...
      Cloning into '/tmp/x/vcflib/fsom'...
      Cloning into '/tmp/x/vcflib/googletest'...
      Cloning into '/tmp/x/vcflib/intervaltree'...
      Cloning into '/tmp/x/vcflib/libVCFH'...
      Cloning into '/tmp/x/vcflib/multichoose'...
      Cloning into '/tmp/x/vcflib/smithwaterman'...
      Cloning into '/tmp/x/vcflib/tabixpp'...
      remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
      error: Server does not allow request for unadvertised object c68cebb4f2e5d5d2b70cf08fbdf1944e9ab2c2dd
      Fetched in submodule path 'fastahack', but it did not contain c68cebb4f2e5d5d2b70cf08fbdf1944e9ab2c2dd. Direct fetching of that commit failed.
      $ git submodule update --init
      error: Server does not allow request for unadvertised object c68cebb4f2e5d5d2b70cf08fbdf1944e9ab2c2dd
      Fetched in submodule path 'fastahack', but it did not contain c68cebb4f2e5d5d2b70cf08fbdf1944e9ab2c2dd. Direct fetching of that commit failed.
      ```

So we'll probably need to adjust that test to not fail when `git` can't possibly clone -- or maybe just not test the clone depth -- it is an "optional performance improvement" after all

DanielChabrowski:
Thanks.

> So we'll probably need to adjust that test to not fail when `git` can't possibly clone -- or maybe just not test the clone depth -- it is an "optional performance improvement" after all

I removed the check but maybe we could restore it later on with an `xfail` if shallow cloning is not supported.

asottile:
Stellar -- thanks for the effort on this 🎉

PR #966, opened 2019-03-09
Compute the maximum command length more accurately

### LINKED ISSUE #691: Improve platform command line length limits

asottile:
Resolves #691 

- [x] try this out on linux
- [x] try this out on macos
- [x] try this out on windows
- [x] try this out on cygwin

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import concurrent.futures
 import contextlib
 import math
+import os
 import sys
 
 import six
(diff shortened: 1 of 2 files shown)


asottile:
Sourced a lot of the information from https://www.in-ulm.de/~mascheck/various/argmax/

asottile:
cygwin and macos fall into the `posix` branch, `cygwin` sets `SC_ARG_MAX` to `32000` which performs basically the same math as the `nt` pathway.  macos seems to report a reasonable value there as well

ashanbrown:
Thanks @asottile.  Sorry not to get around to testing it.

asottile:
No big deal!  Thanks for the original work on this :)

I ended up testing it on my work computer (and last night purchased an old mac mini so hopefully I can test it more readily in the future)

PR #972, opened 2019-03-17
Added test for git.no_git_env()

javabrett:


pre_commit/git.py
```diff
@@ -18,7 +18,7 @@ def zsplit(s):
         return []
 
 
-def no_git_env():
+def no_git_env(env=os.environ):
```
asottile: this is a mutable default -- in this case it isn't a problem but it should probably be `_env=None`

```python
_env = _env if _env is not None else os.environ
```

javabrett: Fixed, thanks!

Issue #965, opened 2019-03-08
warn on unknown keys at the top level


asottile:
We're now warning on unknown keys at the hook level, this should also happen at the top level

yoavcaspi:
I wish to work on this issue.
I'm pre-commit newbie. 
can you refer me to the warning on unknown keys at the hook level?

asottile:
sure!  @mitzkia started some work on this in #980 and added a [`WarnAdditionalKeys`](https://github.com/asottile/cfgv/pull/17) helper to cfgv to make this easier to do -- but looks like they've been busy with other things :)

the extraneous keys for the hook level is done later in the process [here](https://github.com/pre-commit/pre-commit/blob/8df6b22f4fc93b30be322e62f8bf1e0886729648/pre_commit/repository.py#L105-L110) -- though for the top level and repository level this should probably get done in pre_commit/clientlib.py (on the two mapping types directly) via the new `WarnAdditionalKeys`

PR #980, opened 2019-03-25
Check unsupported TOP_LEVEL_KEYS

mitzkia:
This is a fix for: #965 

Of course I will also add unit-tests here.

Signed-off-by: Andras Mitzki <andras.mitzki@balabit.com>

### DIFF
--- a/pre_commit/repository.py
+++ b/pre_commit/repository.py
@@ -162,6 +162,7 @@ def _non_cloned_repository_hooks(repo_config, store, root_config):
 
 
 def _cloned_repository_hooks(repo_config, store, root_config):
+    _check_extra_top_level_keys(repo_config)
     repo, rev = repo_config['repo'], repo_config['rev']
     manifest_path = os.path.join(store.clone(repo, rev), C.MANIFEST_FILE)
     by_id = {hook['id']: hook for hook in load_manifest(manifest_path)}
@@ -188,6 +189,21 @@ def _cloned_repository_hooks(repo_config, store, root_config):
         )
         for hook in hook_dcts
     )
+
+
+_TOP_LEVEL_KEYS = ('repo', 'rev', 'hooks')
+
+
+def _check_extra_top_level_keys(repo_config):
+    extra_top_level_keys = set(dict(repo_config).keys()) - set(_TOP_LEVEL_KEYS)
+    if extra_top_level_keys:
+        logger.warning(
+            'Unexpected top level keys present: {}, only supported'
+            'top level keys: {}'.format(
+                extra_top_level_keys,
+                _TOP_LEVEL_KEYS,
+            ),
+        )
 
 
 def _repository_hooks(repo_config, store, root_config):


mitzkia:
Maybe the warning message not the best, not concrete.

pre_commit/repository.py
```diff
@@ -190,6 +191,21 @@ def _cloned_repository_hooks(repo_config, store, root_config):
     )
 
 
+_TOP_LEVEL_KEYS = ('repo', 'rev', 'hooks')
```
asottile: ideally this should be derived from `CONFIG_REPO_SCHEMA` in `pre_commit.clientlib` -- it probably also makes sense to implement this check as a `cfgv` checker in that module

also these aren't top level keys, just repository-level keys

pre_commit/repository.py
```diff
@@ -190,6 +191,21 @@ def _cloned_repository_hooks(repo_config, store, root_config):
...
+
+
+def _check_extra_top_level_keys(repo_config):
+    extra_top_level_keys = set(dict(repo_config).keys()) - set(_TOP_LEVEL_KEYS)
```
asottile: there's 4 unnecessary copies happening here:

1. `dict(repo_config)`
2. `.keys()` (in python 2)
3. `set(dict...)`
4. `set(_TOP_LEVEL_KEYS)`

The module constant should be `frozenset((...))`, you can use `set(repo_config)` directly

pre_commit/repository.py
```diff
@@ -190,6 +191,21 @@ def _cloned_repository_hooks(repo_config, store, root_config):
...
+    if extra_top_level_keys:
+        logger.warning(
+            'Unexpected top level keys present: {}, only supported'
+            'top level keys: {}'.format(
```
asottile: this produces a not-great error message that contains python literals, use `', '.join(sorted(...))` when displaying lists

mitzkia:
I was also thinking about the exiting if there is an unexpected top level key. 
Could you help me to decide?

pre_commit/repository.py
```diff
@@ -190,6 +191,21 @@ def _cloned_repository_hooks(repo_config, store, root_config):
...
 
 
+_TOP_LEVEL_KEYS = ('repo', 'rev', 'hooks')
+
+
+def _check_extra_top_level_keys(repo_config):
+    extra_top_level_keys = set(dict(repo_config).keys()) - set(_TOP_LEVEL_KEYS)
+    if extra_top_level_keys:
+        logger.warning(
+            'Unexpected top level keys present: {}, only supported'
```
asottile: let's use consistent messaging with the warning on hook-dict extra keys (it's in this module under `create`)

mitzkia:
Thanks, for the review notes. I will fix them. 
One more question: should not we exit if there is an unexpected key, or we can run forward?

asottile:
> Thanks, for the review notes. I will fix them.
> One more question: should not we exit if there is an unexpected key, or we can run forward?

can't error without warning first, don't want to break 5 years of behaviour

mitzkia:
Just found `cfgv.NoAdditionalKeys()`: which is a checker and proper message also displayed, maybe this is the function what we need here.

asottile:
yeah you'd want something like that but `WarnAdditionalKeys` instead

asottile:
@mitzkia with cfgv 1.6 you should be able to finish this :+1:

asottile:
via #1028

PR #981, opened 2019-03-27
Do not run legacy script again when this is the one being executed

tristan0x:
This patch fixes an infinite process recursion when `.git/hooks/pre-commit.legacy` executes itself over and over. I don't know exactly how and why this `pre-commit.legacy` file was created but it is a copy of `.git/hooks/pre-commit`

### DIFF
--- a/pre_commit/resources/hook-tmpl
+++ b/pre_commit/resources/hook-tmpl
@@ -54,8 +54,10 @@ def _run_legacy():
     else:
         stdin = None
 
-    legacy_hook = os.path.join(HERE, '{}.legacy'.format(HOOK_TYPE))
-    if os.access(legacy_hook, os.X_OK):
+    legacy_script = HOOK_TYPE + '.legacy'
+    is_legacy_executed = os.path.basename(__file__) == legacy_script
+    legacy_hook = os.path.join(HERE, legacy_script)
+    if not is_legacy_executed and os.access(legacy_hook, os.X_OK):
         cmd = _norm_exe(legacy_hook) + (legacy_hook,) + tuple(sys.argv[1:])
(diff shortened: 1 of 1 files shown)


asottile:
hmm this shouldn't be possible as the legacy script only will get created if it isn't our script (unless there's a bug in that!): https://github.com/pre-commit/pre-commit/blob/da00fa98f9e274e2d3ef8f756566f1946dc79e36/pre_commit/commands/install_uninstall.py#L85-L87

Do you have a reproduction for how that copy script is getting created?

tristan0x:
> hmm this shouldn't be possible as the legacy script only will get created if it isn't our script (unless there's a bug in that!):

yeah, the `is_our_script` script looks good to me too.

> Do you have a reproduction for how that copy script is getting created?

Unfortunately, I don't. I will ask for more information to the workmate to which it happened, but I am pretty sure he didn't create it himself.

tristan0x:
pre-commit scripts have `ID_HASH = '138fd403232d2ddd5efb44317e38bf03'`. We don't know how it happened.

asottile:
Hmmm well if you can reproduce the situation  I'd rather fix the bug than bandaiding over it :thinking:

tristan0x:
I agree a fix would definitely be a plus here, but IMOH you should prevent the fork bomb either way.

```
cp .git/hooks/pre-commit .git/hooks/pre-commit.legacy 
```

is enough to hang your computer at next commit.

asottile:
My problem with the current PR is it attempts to fix something which is impossible and would hide a bug if it did exist.  If you adjust it to `assert not __file__.endswith('.legacy'), __file__` (or equivalent) I'd be much more comfortable with this PR

PR #1001, opened 2019-04-20
Fixes #1000

### LINKED ISSUE #1000: args after -- fail
## Problem

Occasionally it is necessary to use `--` to signify [options that take an argument](https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash). When `--` is used, any file arguments are ignored.

## Description

I am using `oclint`, which identifies potential problems in c code. `--` is required to specify compiler options. This is my current configuration:

```yaml
[-- skipped for brevity --]

-   repo: local
    hooks:
    -   id: oclint
        name: oclint
        description: Find warnings/errors in C/CPP code
        entry: oclint
        types: [c]
        pass_filenames: true
        args: [
            -enable-clang-static-analyzer,
            -enable-global-analysis,
            --,
            -report-type=json,
            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON,
        ]
        language: system
```

When running on a valid c file, I get this output:

```bash
oclint...................................................................Failed
hookid: oclint

oclint: Not enough positional command line arguments specified!
Must specify at least 1 positional argument: See: /usr/local/bin/oclint -help
```

## Solution

Looking at the code, it looks like you are appending file arguments to the end instead of infixing them before `--`. A pull request that fixes this will be coming in a few.

pocc:
Passes all tests. I am not sure whether it will cause unintended behavior as I have not tested it against all possible tools (maintainers will probably have a better idea of this).

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -53,11 +53,37 @@ class ArgumentTooLongError(RuntimeError):
     pass
 
 
-def partition(cmd, varargs, target_concurrency, _max_length=None):
-    _max_length = _max_length or _get_platform_max_length()
+def infix_vararg(cmd, vararg):
+    """Insert vararg before --, if it is not the last elem of the command
 
-    # Generally, we try to partition evenly into at least `target_concurrency`
-    # partitions, but we don't want a bunch of tiny partitions.
+    This is required because if vararg comes after --, then the command will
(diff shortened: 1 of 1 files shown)


pocc:
Closing as I have created `https://github.com/pocc/pre-commit-hooks` to deal with this instead.

Issue #884, opened 2018-12-13
prepend certain characters to a commit-msg using precommit


milind-shakya-sp:
Not sure if the new `commit-msg` stage helps, may be we need a new `prepare-commit-msg` stage that can hook into ways people can add to commit messages?

asottile:
`commit-msg` should allow editing of the message, can you share what you've tried so far that hasn't worked?

asottile:
please comment back with more information and I'll reopen.  as I understand, `commit-msg` is sufficient to modify the message that will be committed

milin:
@asottile yes turns out`commit-msg` was sufficient. I've opened a pr on the pre-commit-hooks repo https://github.com/pre-commit/pre-commit-hooks/pull/347 if other people want to use as well.

Thanks for the info!

marcjay:
I do think there would be value in a `prepare-commit-msg` stage to match what's available with git: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

I would like to pre-populate the commit message with some dynamic information such as the Ticket ID and description (so a git template is not suitable). I'd rather not mess with a developer's commit message after they've written it (i.e. use the `commit-msg` stage) but instead just add a useful starting line.

If it's something that is likely to be accepted and is reasonably straightforward I may be able to offer a PR to add this

asottile:
shouldn't be too bad @marcjay -- can basically follow the PR that added `commit-msg` (though some of the internals have moved a bit since then)

https://github.com/pre-commit/pre-commit/pull/566/files

milind-shakya-sp:
@marcjay I went ahead with creating a precommit hook instead. You can checkout https://github.com/milin/giticket to see if its useful.

marcjay:
Thanks for the quick replies both!

Thank you @milind-shakya-sp I did take a look at giticket and tried it out. Nice work! For our use-case I don't really want to quietly change a dev's commit message after they've written it, just provide a relevant template to use. I might use something similar to abort the commit if it's missing the ticket ID as a quick intermediate solution

Cheers @asottile that's really helpful and looks achievable, I plan to have a go over an evening or two.

asottile:
The biggest internal change since that PR is the combination of the hook template and removal of `bash`,  hopefully (?) it's simpler now: https://github.com/pre-commit/pre-commit/blob/b33f2c40d8de0cd2c55a9b637773172198962832/pre_commit/resources/hook-tmpl#L162-L169

Issue #833, opened 2018-09-24
commit-msg hook with gitlint makes the text get lost on failure


rico-chet:
When using `gitlint` as a `commit-msg` hook with `pre-commit`, I run into issue of commit message getting lost easily when gitlint finds that it's not good enough.

Repro: on a clean minimal debian 9.5.0, I run:
```
apt install --assume-yes curl git python3
curl https://bootstrap.pypa.io/get-pip.py | python3 - pre-commit

git clone --mirror https://github.com/jorisroovers/gitlint

mkdir gitlint-test
cd gitlint-test
git init
git config user.email "you@example.com"
git config user.name "Your Name"
git commit --allow-empty --message="Initial empty commit"
cat > .pre-commit-config.yaml <<EOF
repos:
  - repo: ../gitlint.git
    rev: v0.10.0
    hooks:
    - id: gitlint
      stages:
      - commit-msg
EOF
git add .pre-commit-config.yaml
git commit --message="Add pre-commit config"
pre-commit install --hook-type commit-msg
touch foo
git add foo
git commit
*enter a lengthy commit message*
*find that gitlint keeps commit from succeeding for some valid reason*
git commit # try to fix the lengthy message
*find that all message text is gone :-(*
```
One can retrieve the text from .git/COMMIT_EDITMSG after the gitlint test, but once you hit `git commit` for the second time, all is lost. I'd expect that you get a second chance without risk of losing all your text.

IIRC, I didn't observe this issue with `gitlint` alone.

asottile:
Could be a neat little feature yeah!  Need to be sure it plays nicely if someone defines their own `prepare-commit-msg` hook and probably mesh with the rest of the framework but definitely sounds like a sharp edge in git that `pre-commit` could smooth over!

asottile:
Now that `prepare-commit-msg` will land in the next pre-commit release, it should be pretty easy to make something like this!

asottile:
Here's a small demo of this idea, this can probably be cleaned up and turned into a `repo: meta` hook

### .pre-commit-config.yaml
```yaml
repos:
-   repo: local
    hooks:
    -   id: save-commit-msg
        name: save commit message
        entry: ./bin/save-commit-msg
        verbose: true
        language: script
        stages: [commit-msg]
    -   id: restore-commit-msg
        name: restore commit message
        entry: ./bin/restore-commit-msg
        verbose: true
        language: script
        stages: [prepare-commit-msg]
    -   id: derp
[...]

PR #1004, opened 2019-04-21
Add support for prepare-commit-msg hooks

marcjay:
Adds a `prepare-commit-msg` hook stage which allows for hooks which add
dynamic suggested/placeholder text to commit messages that an author can
use as a starting point for writing a commit message.

An example being to pre-populate the first line of the commit message with some dynamic information such as the ticket ID and description (if branch name contains the ticket ID let's say). The benefit of this over a `commit-msg` hook in this scenario being that it doesn't modify an author's commit message 'behind their back'.

May help with #833 and relates to #884

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -190,7 +190,7 @@ def _compute_cols(hooks, verbose):
 def _all_filenames(args):
     if args.origin and args.source:
         return git.get_changed_files(args.origin, args.source)
-    elif args.hook_stage in ['prepare-commit-msg', 'commit-msg']:
+    elif args.hook_stage in {'prepare-commit-msg', 'commit-msg'}:
         return (args.commit_msg_filename,)
     elif args.files:
         return args.files
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -190,7 +190,7 @@ def _compute_cols(hooks, verbose):
 def _all_filenames(args):
     if args.origin and args.source:
         return git.get_changed_files(args.origin, args.source)
-    elif args.hook_stage == 'commit-msg':
+    elif args.hook_stage in ['prepare-commit-msg', 'commit-msg']:
         return (args.commit_msg_filename,)
     elif args.files:
         return args.files
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -615,7 +615,7 @@ def test_prepare_commit_msg_hook(cap_out, store, prepare_commit_msg_repo):
         stage=False,
     )
 
-    with io.open(filename, 'rt') as f:
+    with io.open(filename) as f:
         assert 'Signed off by: ' in f.read()
 
 
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -557,7 +557,12 @@ def test_stages(cap_out, store, repo_with_passing_hook):
                 'language': 'pygrep',
                 'stages': [stage],
             }
-            for i, stage in enumerate(('commit', 'push', 'manual'), 1)
+            for i, stage in enumerate(
+                (
+                    'commit', 'push', 'manual', 'prepare-commit-msg',
+                    'commit-msg',
+                ), 1,
+            )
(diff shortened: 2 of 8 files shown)


tests/conftest.py
```diff
@@ -134,6 +135,54 @@ def commit_msg_repo(tempdir_factory):
...
+
+@pytest.fixture
+def failing_prepare_commit_msg_repo(tempdir_factory):
+    path = git_dir(tempdir_factory)
+    config = {
+        'repo': 'local',
+        'hooks': [{
+            'id': 'add-signoff',
+            'name': 'Add "Signed off by:"',
+            'entry': '/usr/bin/env bash -c "exit 1"',
```
asottile: this can just be `'bash -c "exit 1"'`

tests/commands/run_test.py
```diff
@@ -593,6 +600,25 @@ def test_commit_msg_hook(cap_out, store, commit_msg_repo):
...
+        cap_out,
+        store,
+        prepare_commit_msg_repo,
+        {'hook_stage': 'prepare-commit-msg', 'commit_msg_filename': filename},
+        expected_outputs=[b'Add "Signed off by:"', b'Passed'],
+        expected_ret=0,
+        stage=False,
+    )
+
+    with io.open(filename, 'rt') as f:
```
asottile: `'rt'` is the default, (remove the second argument, same above)

pre_commit/commands/run.py
```diff
@@ -190,7 +190,7 @@ def _compute_cols(hooks, verbose):
 def _all_filenames(args):
     if args.origin and args.source:
         return git.get_changed_files(args.origin, args.source)
-    elif args.hook_stage == 'commit-msg':
+    elif args.hook_stage in ['prepare-commit-msg', 'commit-msg']:
```
asottile: very minor, this should be a set or tuple literal instead of a list literal `in {'prepare-commit-msg', 'commit-msg'}`

marcjay:
Cheers, and thanks for the pointers - nice job on building something so easily extendible!

Yup, I'll put up the beginnings of a hook I've been working on for this, one sec...

asottile:
don't worry about the windows failures, I'll fix those -- thanks for the PR 👍

marcjay:
Great, thank you!

In terms of an example, this was something quick I put together for testing - it's the reason for wanting this feature. It has no error checking or anything yet, just a dirty script at this stage!

- Uses the https://pypi.org/project/jira/ library
- Expects branch names to start with a JIRA ticket ID, e.g. `ABC-123-replace-foo`

```
def get_branch_name():
    return subprocess.check_output(
        [
            'git',
            'rev-parse',
            '--abbrev-ref',
            'HEAD',
        ],
    ).decode('UTF-8')


def get_ticket_id_from_branch_name(branch_name):
    return re.findall('^[A-Z]{1,10}-[0-9]{1,5}', branch_name)[0]


def main():
    if os.environ.get('GIT_EDITOR') == ':':
        # According to git docs, the editor will not show, e.g. because
        # of a `git commit -m "..."` supplied message. Therefore don't touch
        # the commit message as there is no opportunity for the user to edit it
        return

    jira = JIRA(server=JIRA_SERVER)

    branch_name = get_branch_name()
    ticket_id = get_ticket_id_from_branch_name(branch_name)

    issue = jira.issue(ticket_id)
    summary = issue.fields.summary

    with io.open(sys.argv[1], 'r+') as f:
        contents = f.readlines()
        contents.insert(0, u'{} - {}\n'.format(ticket_id, summary))
        f.seek(0, 0)
        f.writelines(contents)
```

`git commit` results in a commit message that starts:
```
ABC-123 - Replace Foo with Bar

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
```

asottile:
neat!

benmezger:
@marcjay Nice! I wrote something similar, which allows customization of templates and handles amends and non-jira branches

https://github.com/benmezger/gjira/tree/master/gjira

PR #1015, opened 2019-04-29
Could this fix #1013?

### LINKED ISSUE #1013: broken pcre detection is broken
https://github.com/pre-commit/pre-commit/blob/09e64e8ff6b2d31ecf6be2499bf88ed0c651e0eb/testing/util.py#L69-L71

When I moved to declarative metadata this got broken.

This'll need to be updated to look at something else.  Before it was making sure that single quotes were being handled properly inside pcre greps (since this is broken in the msys implementation of `grep -P`)

Myrheimb:
I'm still a beginner, but a single single quote looked a bit off to me.
Could adding another single quote after pre fix this issue?

### DIFF
--- a/testing/util.py
+++ b/testing/util.py
@@ -68,7 +68,7 @@ xfailif_broken_deep_listdir = pytest.mark.xfail(
 
 def platform_supports_pcre():
     output = cmd_output(GREP, '-P', "Don't", 'CHANGELOG.md', retcode=None)
-    return output[0] == 0 and "name='pre_commit'," in output[1]
+    return output[0] == 0 and "Don't use readlink -f" in output[1]
 
 
 xfailif_no_pcre_support = pytest.mark.xfail(
--- a/testing/util.py
+++ b/testing/util.py
@@ -67,7 +67,7 @@ xfailif_broken_deep_listdir = pytest.mark.xfail(
 
 
 def platform_supports_pcre():
-    output = cmd_output(GREP, '-P', "name='pre", 'setup.py', retcode=None)
+    output = cmd_output(GREP, '-P', "name='pre'", 'setup.py', retcode=None)
     return output[0] == 0 and "name='pre_commit'," in output[1]
 
 
--- a/testing/util.py
+++ b/testing/util.py
@@ -67,7 +67,7 @@ xfailif_broken_deep_listdir = pytest.mark.xfail(
 
 
 def platform_supports_pcre():
-    output = cmd_output(GREP, '-P', "name='pre'", 'setup.py', retcode=None)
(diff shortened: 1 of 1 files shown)


asottile:
ah so this thing is *intentionally* trying to grep against a quote somewhere in some file -- in #956 I moved all of the single quotes out of that file though :laughing: 

basically this callsite just needs to be updated to a different file and a different single-quote

one easy one could be `"Don't"` in `CHANGELOG.md`: https://github.com/pre-commit/pre-commit/blame/c97dfb22813eb040fafd6fa1a9be91e690ae1184/CHANGELOG.md#L53

Myrheimb:
Ah I see :-) Sorry for the inconvenience!

asottile:
@Myrheimb no inconvenience at all!  If you'd like to try and update to that file and string above that should fix it :)

when fixed there should be no `xpassed` on the linux builds

testing/util.py
```diff
@@ -67,7 +67,7 @@ def broken_deep_listdir():  # pragma: no cover (platform specific)
 
 
 def platform_supports_pcre():
-    output = cmd_output(GREP, '-P', "name='pre'", 'setup.py', retcode=None)
```
asottile: oh and remove the `name=` part too :)

Myrheimb: Ah yes, of course - thanks!

asottile: oh uh right, and the line below it needs updating too to validate the output, looks like any of these will work:

```console
$ grep -P "Don't" CHANGELOG.md 
- Don't filter `GIT_SSH_COMMAND` env variable from `git` commands
- Don't filter `GIT_EXEC_PREFIX` env variable from `git` commands
- Don't write ANSI colors on windows when color enabling fails
- Don't crash when `always_run` is `True` and `files` is not provided.
- Don't run on staged deleted files that still exist
- Don't rename across devices when creating sqlite database
- Don't use readlink -f
```

I'd do the last one since it's the shortest :)

Myrheimb: Appreciate the help as I'm learning a lot in the process :-)

asottile: No problem :) you can also run these tests locally as well -- the contributing guide goes into a little bit of detail there: https://github.com/pre-commit/pre-commit/blob/master/CONTRIBUTING.md

asottile:
looks like this is working on the linux builders :tada: 

```
========================= 575 passed in 265.91 seconds =========================
```

and properly excluding on the windows ones:

```
======== 556 passed, 6 skipped, 7 xfailed, 6 xpassed in 854.05 seconds ========
```

asottile:
Thanks!

PR #1028, opened 2019-05-11
Add warning to additional keys in config

### LINKED ISSUE #965: warn on unknown keys at the top level

yoavcaspi:
Resolves #965

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -109,6 +109,8 @@ META = 'meta'
 
 
 class MigrateShaToRev(object):
+    key = 'rev'
+
     @staticmethod
     def _cond(key):
         return cfgv.Conditional(
@@ -263,7 +265,14 @@ CONFIG_SCHEMA = cfgv.Map(
         '0',
     ),
     cfgv.WarnAdditionalKeys(
-        ('repos',),
+        (
+            'repos',
+            'default_language_version',
+            'default_stages',
+            'exclude',
+            'fail_fast',
+            'minimum_pre_commit_version',
+        ),
         warn_on_unknown_keys_at_top_level,
     ),
 )
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -149,7 +149,7 @@ def _entry(modname):
     )
 
 
-def warn_on_unknown_keys_at_top_level(extra, orig_keys):
+def warn_unknown_keys(extra, orig_keys):
     logger.warning(
         'Unexpected config key(s): {}'.format(
             ', '.join(sorted(extra)),
@@ -237,7 +237,7 @@ CONFIG_REPO_DICT = cfgv.Map(
     MigrateShaToRev(),
     cfgv.WarnAdditionalKeys(
         ('repo', 'rev', 'hooks'),
-        warn_on_unknown_keys_at_top_level,
+        warn_unknown_keys,
     ),
 )
 DEFAULT_LANGUAGE_VERSION = cfgv.Map(
@@ -273,7 +273,7 @@ CONFIG_SCHEMA = cfgv.Map(
             'fail_fast',
             'minimum_pre_commit_version',
         ),
-        warn_on_unknown_keys_at_top_level,
+        warn_unknown_keys,
     ),
 )
 
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import argparse
 import functools
+import logging
 import pipes
 import sys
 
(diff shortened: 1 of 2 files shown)


pre_commit/clientlib.py
```diff
@@ -144,6 +147,16 @@ def _entry(modname):
     )
 
 
+def warn_on_unknown_keys_at_top_level(extra, orig_keys):
+    logger.warning(
+        'Your pre-commit-config contain these extra keys: {}. '
+        'while the only valid keys are: {}.'.format(
```
asottile: let's keep this short and sweet, how about
```python
'Unexpected key(s): {}'.format(', '.join(sorted(extra)))
```

pre_commit/clientlib.py
```diff
@@ -222,6 +235,10 @@ def _entry(modname):
     ),
 
     MigrateShaToRev(),
+    cfgv.WarnAdditionalKeys(
+        {'repo', 'rev', 'hooks'},
```
asottile: let's use a tuple here instead

pre_commit/clientlib.py
```diff
@@ -222,6 +235,10 @@ def _entry(modname):
     ),
 
     MigrateShaToRev(),
+    cfgv.WarnAdditionalKeys(
```
asottile: can you add a quick test to make sure this doesn't drift from the `CONFIG_REPO_DICT` definition?

Basically will need to iterate through the things in that `Map`'s `.items` and check the keys against the keys passed in here

pre_commit/clientlib.py
yoavcaspi: why tuple?

pre_commit/clientlib.py
yoavcaspi: are you sure it will be clear to the user what is the problem here?

pre_commit/clientlib.py
asottile: immutability, WarnAdditionalKeys takes any `Sequence` and will convert it to a set when comparing the keys

pre_commit/clientlib.py
asottile: hmm then maybe `Unexpected config key(s): ...`?  The other message has context of location but it looks like we didn't think about that when adding `WarnAdditionalKeys` 🤔 https://github.com/pre-commit/pre-commit/blob/1c7ac67fba963b7a8cda852ed04bceb8a1c37063/pre_commit/repository.py#L108

pre_commit/clientlib.py
yoavcaspi: I'm not following. can you elaborate?

From what I understand you wish to add a test to validate that the new attribute does not change the definition of CONFIG_REPO_DICT and just added a kind of decorator. but I'm not sure how to do it.

asottile: ~something like this

```diff
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -109,6 +109,8 @@ META = 'meta'
 
 
 class MigrateShaToRev(object):
+    key = 'rev'
+
     @staticmethod
     def _cond(key):
         return cfgv.Conditional(
diff --git a/tests/clientlib_test.py b/tests/clientlib_test.py
index cace0f3..13b42a5 100644
--- a/tests/clientlib_test.py
+++ b/tests/clientlib_test.py
@@ -305,3 +305,12 @@ def test_minimum_pre_commit_version_failing():
 def test_minimum_pre_commit_version_passing():
     cfg = {'repos': [], 'minimum_pre_commit_version': '0'}
     cfgv.validate(cfg, CONFIG_SCHEMA)
+
+
+@pytest.mark.parametrize('schema', (CONFIG_SCHEMA, CONFIG_REPO_DICT))
+def test_warn_additional(schema):
+    allowed_keys = {item.key for item in schema.items if hasattr(item, 'key')}
+    warn_additional, = [
+        x for x in schema.items if isinstance(x, cfgv.WarnAdditionalKeys)
+    ]
+    assert allowed_keys == set(warn_additional.keys)
```

yoavcaspi: Done.
This test helped me find some more issues with the code. Thanks :-)

yoavcaspi:
Let me know if there are other issues.

pre_commit/clientlib.py
```diff
@@ -144,6 +149,14 @@ def _entry(modname):
     )
 
 
+def warn_on_unknown_keys_at_top_level(extra, orig_keys):
```
asottile: probably just `warn_unknown_keys` now that it's being used in other places

yoavcaspi: cool. thanks.

asottile:
Thanks again!  And thanks to @mitzkia for the assist via `cfgv`!

PR #1031, opened 2019-05-12
Add advice to "Files were modified…" error

katrinleinweber:
Would it make sense to include something like this? If yes, please let me know whether the formatting for a string that should not be too long in the source code, but single-line in the out-put is correct.

Afterwards, I'll:

- [ ] update [run_test.py](https://github.com/pre-commit/pre-commit/blob/b8300268bf2d8797f28edf9063d7e5659ad4d535/tests/commands/run_test.py#L126)

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -152,7 +152,9 @@ def _run_single_hook(classifier, hook, args, skips, cols):
 
         # Print a message if failing due to file modifications
         if file_modifications:
-            output.write('Files were modified by this hook.')
+            output.write('Files were modified by this hook.' \
+                         'Please review them in Git\'s staging area, ' \
+                         'and repeat the commit.')
 
             if stdout or stderr:
                 output.write_line(' Additional output:')


pre_commit/commands/run.py
```diff
@@ -152,7 +152,9 @@ def _run_single_hook(classifier, hook, args, skips, cols):
 
         # Print a message if failing due to file modifications
         if file_modifications:
-            output.write('Files were modified by this hook.')
+            output.write('Files were modified by this hook.' \
+                         'Please review them in Git\'s staging area, ' \
+                         'and repeat the commit.')
```
asottile: seems fine to add more context here

maybe something like this?

```python
output.write(
    'Files were modified by this hook.  '
    'Review the changes, `git add`, and repeat the commit.',
)
```

hmmm, but it's not terribly uncommon for this message to show multiple times with lots of fixers (it's a bit redundant in that case) and now it'll wrap on narrow terminals 🤔 

(the change won't be in the staging area, but will need to be added there)

asottile:
tl;dr I'm on the fence, it seems like a good idea but also seems a little verbose?

katrinleinweber:
True. Would a separate, final message be better, to avoid redundancy in case several files are modified?

asottile:
> True. Would a separate, final message be better, to avoid redundancy in case several files are modified?

Ooh yeah that's a good idea, we already do something similar for `--show-diff-on-failure`, we'd just need to pass along the "changed files" information somehow.

There's a perf improvement that I've been meaning to make here as well, right now execution looks roughly like this:

```
- run hook 1
    - diff = git diff
    - run hook
    - diff2 = git diff (compare against diff and report Changed files)
- run hook 2
    - diff = git diff
    - run hook
    - diff2 = git diff (compare against diff and report Changed files)
```

But it could look like this:

```
- initial_diff = git diff
- run hook 1 (take initial_diff as a parameter)
    - run hook
    - new_diff = git diff (compare against passed in diff)
   - return the new diff
- run hook 2 (take previous returned diff as a parameter)
    - run hook
    - new_diff = git diff (compare against passed in diff)
   - return the new diff
etc.
```

with that second patern, it would be easy to compare the initial diff against the last returned diff to see if there was any changes across the whole set of hooks

PR #1033, opened 2019-05-13
restore testenv:venv section

yoavcaspi:
to be able to set the dev environment as described in the [CONTRIBUTING.md](https://github.com/pre-commit/pre-commit/blob/master/CONTRIBUTING.md#setting-up-an-environment)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,4 +1,5 @@
 [tox]
+project = pre_commit
 envlist = py27,py36,py37,pypy,pypy3,pre-commit
 
 [testenv]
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -1,4 +1,5 @@
 [tox]
+project = pre_commit
```
asottile: let's inline this, it used to be used more than once but not any more

tox.ini
```diff
@@ -10,6 +11,10 @@ commands =
     coverage report --fail-under 100
     pre-commit install
 
+[testenv:venv]
+envdir = venv-{[tox]project}
+commands =
```
asottile: let's put this below the actually used environments (also conveniently alphabetical!)

PR #1051, opened 2019-05-31
Pass color option to git diff (on failure)

### LINKED ISSUE #1007: --show-diff-on-failure does not respect --color
I'm running `pre-commit` on gitlab, and I noticed that the diffs produced when `--show-diff-on-failure` is enabled do not respect `--color=always`.  

It looks like a pretty easy fix:   we just need to pass the `--color` setting to `git diff`.  It supports the same "auto", "always", "never" settings as `pre-commit`.  

`git diff --color` [documentation](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---colorltwhengt).

code in question: https://github.com/pre-commit/pre-commit/blob/master/pre_commit/commands/run.py#L227

mandarvaze:
Fixes #1007

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -224,13 +224,10 @@ def _run_hooks(config, hooks, args, environ):
                 '`pre-commit install`.',
             )
         output.write_line('All changes made by hooks:')
-        if args.color:
-            subprocess.call((
-                'git', '--no-pager', 'diff', '--no-ext-diff',
-                '--color={}'.format(args.color),
-            ))
-        else:
-            subprocess.call(('git', '--no-pager', 'diff', '--no-ext-diff'))
+        subprocess.call((
+            'git', '--no-pager', 'diff', '--no-ext-diff',
+            '--color={}'.format(args.color),
+        ))
 
     return retval
 
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -224,7 +224,14 @@ def _run_hooks(config, hooks, args, environ):
                 '`pre-commit install`.',
             )
         output.write_line('All changes made by hooks:')
-        subprocess.call(('git', '--no-pager', 'diff', '--no-ext-diff'))
+        if args.color:
+            subprocess.call((
+                'git', '--no-pager', 'diff', '--no-ext-diff',
+                '--color={}'.format(args.color),
+            ))
+        else:
+            subprocess.call(('git', '--no-pager', 'diff', '--no-ext-diff'))
+
     return retval
 
 
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -224,9 +224,11 @@ def _run_hooks(config, hooks, args, environ):
                 '`pre-commit install`.',
             )
         output.write_line('All changes made by hooks:')
+        # args.color is a boolean.
+        # See user_color function in color.py
         subprocess.call((
             'git', '--no-pager', 'diff', '--no-ext-diff',
-            '--color={}'.format(args.color),
+            '--color={}'.format({True: 'always', False: 'never'}[args.color]),
         ))
 
     return retval
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -184,7 +184,13 @@ def test_global_exclude(cap_out, store, tempdir_factory):
         (
             {
                 'show_diff_on_failure': True,
-                'color': 'auto',
+            },
+            b'All changes made by hooks:',
+        ),
+        (
+            {
+                'show_diff_on_failure': True,
+                'color': True,
             },
             b'All changes made by hooks:',
         ),
@@ -192,7 +198,6 @@ def test_global_exclude(cap_out, store, tempdir_factory):
             {
                 'show_diff_on_failure': True,
                 'all_files': True,
-                'color': 'auto',
             },
             b'reproduce locally with: pre-commit run --all-files',
         ),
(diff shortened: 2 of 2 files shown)


mandarvaze:
@asottile Build is failing on Windows, getting error : `InvocationError for command 'D:\a\1\s\.tox\py37\Scripts\coverage.EXE`

Can you help ?

pre_commit/commands/run.py
```diff
@@ -224,7 +224,14 @@ def _run_hooks(config, hooks, args, environ):
                 '`pre-commit install`.',
             )
         output.write_line('All changes made by hooks:')
-        subprocess.call(('git', '--no-pager', 'diff', '--no-ext-diff'))
+        if args.color:
```
asottile: the color option should always be present so you can just pass that through

asottile:
alright I've fixed the `rust` issues on master -- you'll want to rebase on top of that and fix the above comment and then this should be good to go! :+1:

pre_commit/commands/run.py
mandarvaze: That is what I had put in initially, but the test `test_show_diff_on_failure` fails, cause over there `args.color` is `False`.

So I had added another entry with `color: 'auto'` in the test case so that both the paths are tested.

Now, 
1. I'm passing color unconditionally, as you suggested **and**
2. I have added `color: 'auto'` both **existing** test parameter, so that the test does not fail.

asottile: ah ok yeah you're right -- I forgot that `pre_commit.color` translates the `('always', 'auto', 'never')` into booleans

Let me take another look at this then 👍

pre_commit/commands/run.py
```diff
@@ -224,7 +224,14 @@ def _run_hooks(config, hooks, args, environ):
...
+            subprocess.call((
+                'git', '--no-pager', 'diff', '--no-ext-diff',
+                '--color={}'.format(args.color),
+            ))
+        else:
+            subprocess.call(('git', '--no-pager', 'diff', '--no-ext-diff'))
```
asottile: I think basically what we're going to have to end up with is:

```python
cmd = ('git', '--no-pager', 'diff', '--no-ext-diff')
if args.color:
    cmd += ('--color=always',)
subprocess.call(cmd)
```

Since at this point `args.color` will always be a boolean

tests/commands/run_test.py
```diff
@@ -187,6 +187,13 @@ def test_global_exclude(cap_out, store, tempdir_factory):
             },
             b'All changes made by hooks:',
         ),
+        (
+            {
+                'show_diff_on_failure': True,
+                'color': 'auto',
```
asottile: so as above, this isn't a valid option here unfortunately

asottile:
ah shoot sorry, I wasn't fast enough -- you were right /o\

tests/commands/run_test.py
mandarvaze: But the tests pass :)

asottile: yeahhhh the tests pass because we ignore the return code of `git diff` -- it actually ends up printing something like this:

```console
$ git diff --color=True
error: option `color' expects "always", "auto", or "never"
```

pre_commit/commands/run.py
mandarvaze: I don't understand what this means. Can you point me to the relevant code ?

Also, if `args.color` is boolean, then how can be pass the "correct" value to git command ?

I think my latest changes are correct i.e. unconditionally pass the `color` option to git, and fix the test.

asottile: ah yes

So the color option is set up [here](https://github.com/pre-commit/pre-commit/blob/fff3ad518cb9a9bf169346aa4727fb70a67a4e30/pre_commit/main.py#L40) -- this is the argument parsing bit for pre-commit.  The `type=color.use_color` tells argparse to use that callback to determine the actual value of that argument

The `use_color` function is defined [here](https://github.com/pre-commit/pre-commit/blob/fff3ad518cb9a9bf169346aa4727fb70a67a4e30/pre_commit/color.py#L42-L54) where it translates the options into booleans

You're right you do need to always pass it and not use the code sample I put below

One quick way to translate that would be to do something like this:

```python
'--color={}'.format({True: 'always', False: 'never'}[args.color])
```

This'll map the boolean values to the string options and format that into the `--color` call.

Adding that all together I think you end up with something like this

```python
subprocess.call((
    'git', '--no-pager', 'diff', '--no-ext-diff',
    '--color={}'.format({True: 'always', False: 'never'}[args.color]),
))
```

mandarvaze:
@asottile  Now the build is failing on `tox-linux-pypy3` with similar error:

`InvocationError for command /home/vsts/work/1/s/.tox/pypy3/bin/coverage `

I don't have "write" access, so someone with write access will have to "ignore the error" and accept/merge the commit.

asottile:
I'm streaming right now if you want to chat about this PR -- https://twitch.tv/anthonywritescode :)

asottile:
the CI flakes are due to this I believe: https://github.com/nedbat/coveragepy/issues/807

hopefully I'll be able to fix them at some point :S

pre_commit/commands/run.py
mandarvaze: Done.

mandarvaze:
@asottile Hopefully this is the last iteration. `tox-linux-pypy3` still fails (as above) but for unrelated reason.

Please review and accept/merge the PR.

asottile:
Thanks!

PR #1052, opened 2019-05-31
azure pipelines now has rust by default on windows

asottile:
annoyingly it isn't on the linux agents yet -- apparently my repo will get that on 2019-06-04

I needed to add an additional variable to passenv for windows due to https://github.com/microsoft/azure-pipelines-image-generation/issues/988


asottile:
ughhhhh https://github.com/microsoft/azure-pipelines-image-generation/issues/989

PR #1054, opened 2019-06-06
Fixes for rust tests on azure pipelines linux

asottile:


asottile:
https://github.com/microsoft/azure-pipelines-image-generation/issues/996 for sqlite breakage :/

PR #1055, opened 2019-06-08
Use current version of Python for Windows by default

PeterMosmans:
By default, Windows used `python` without version identifier. This patch specifies to use the current version of Python for Windows.

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -5,6 +5,8 @@ import os.path
 import sys
 
 from pre_commit.util import cmd_output
+from pre_commit.util import fix_mingw_path
+from pre_commit.util import is_mingw
 
 
 logger = logging.getLogger(__name__)
@@ -36,7 +38,12 @@ def no_git_env(_env=None):
 
 
 def get_root():
-    return cmd_output('git', 'rev-parse', '--show-toplevel')[1].strip()
+    return fix_mingw_path(
+        cmd_output(
+            'git', 'rev-parse',
+            '--show-toplevel',
+        )[1].strip(),
+    )
 
 
 def get_git_dir(git_root='.'):
@@ -159,7 +166,8 @@ def git_path(name, repo='.'):
 
 def check_for_cygwin_mismatch():
     """See https://github.com/pre-commit/pre-commit/issues/354"""
-    if sys.platform in ('cygwin', 'win32'):  # pragma: no cover (windows)
+    if sys.platform in ('cygwin', 'win32') and \
+            not is_mingw():  # pragma: no cover (windows)
         is_cygwin_python = sys.platform == 'cygwin'
         toplevel = cmd_output('git', 'rev-parse', '--show-toplevel')[1]
         is_cygwin_git = toplevel.startswith('/')
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -49,7 +49,7 @@ def is_our_script(filename):
 
 def shebang():
     if sys.platform == 'win32':
-        py = 'python'
+        py = 'python{}'.format(sys.version_info[0])
     else:
         # Homebrew/homebrew-core#35825: be more timid about appropriate `PATH`
         path_choices = [p for p in os.defpath.split(os.pathsep) if p]
@@ -62,7 +62,7 @@ def shebang():
                 py = exe
                 break
         else:
-            py = 'python'
+            py = 'python{}'.format(sys.version_info[0])
     return '#!/usr/bin/env {}'.format(py)
 
 
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -14,6 +14,7 @@ from pre_commit.clientlib import load_config
 from pre_commit.repository import all_hooks
 from pre_commit.repository import install_hook_envs
 from pre_commit.util import cmd_output
+from pre_commit.util import is_mingw
 from pre_commit.util import make_executable
 from pre_commit.util import mkdirp
 from pre_commit.util import resource_text
@@ -48,8 +49,8 @@ def is_our_script(filename):
 
 
 def shebang():
-    if sys.platform == 'win32':
-        py = 'python{}'.format(sys.version_info[0])
+    if sys.platform == 'win32' and not is_mingw():
+        py = 'python'
     else:
         # Homebrew/homebrew-core#35825: be more timid about appropriate `PATH`
         path_choices = [p for p in os.defpath.split(os.pathsep) if p]
--- a/pre_commit/resources/hook-tmpl
+++ b/pre_commit/resources/hook-tmpl
@@ -6,6 +6,7 @@ import distutils.spawn
 import os
 import subprocess
 import sys
+import sysconfig
 
 # work around https://github.com/Homebrew/homebrew-core/issues/30445
 os.environ.pop('__PYVENV_LAUNCHER__', None)
@@ -76,7 +77,10 @@ def _run_legacy():
 
 def _validate_config():
     cmd = ('git', 'rev-parse', '--show-toplevel')
-    top_level = subprocess.check_output(cmd).decode('UTF-8').strip()
+    top_level = fix_mingw_path(
+        subprocess.check_output(cmd).decode('UTF-8').
+        strip(),
+    )
     cfg = os.path.join(top_level, CONFIG)
     if os.path.isfile(cfg):
         pass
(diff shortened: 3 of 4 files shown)


asottile:
python on windows doesn't have a pep 394 named executable so this will break that:

```console
$ ls C:/Python37/
DLLs/     Lib/         NEWS.txt      python37.dll*  tcl/
Doc/      libs/        python.exe*   pythonw.exe*   Tools/
include/  LICENSE.txt  python3.dll*  Scripts/       vcruntime140.dll*
```

asottile:
what issue are you running into?

PeterMosmans:
Hi,

I'm running Python on MSYS2 (mingw64), and that uses the 'correct' PEP 394 Python naming. Without this patch it fails, as the incorrect Python version is being referenced. I could add an exception for MSYS2 as it's easy to detect, that might do the trick.

asottile:
can you show how you installed python? I'm also running msys above as well -- cygwin has pep 394 named executables, but msys ~usually just runs the windows installed python (cygwin compiled python does not report "win32" so it does not run that branch -- see https://github.com/asottile/scratch/wiki/platforms#cygwin-cpython-2x-32-bit

Can you show some output? some crashes? I still have nothing to go on from your report

PeterMosmans:
Python is installed using pacman (`pacman -Sy mingw-w64-x86_64-python3`) - there is no "Windows Python" version installed.
```
% python3
Python 3.7.3 (default, May 28 2019, 08:37:07)  [GCC 8.3.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os, platform, sys
>>> sys.version
'3.7.3 (default, May 28 2019, 08:37:07)  [GCC 8.3.0 64 bit (AMD64)]'
>>> os.name
'nt'
>>> sys.platform
'win32'
>>> platform.system()
'Windows'
>>> platform.python_implementation()
'CPython'
>>> sys.implementation.name
'cpython'
```

If you don't have Python2 installed, the script won't work, as the binary named `python` doesn't exist. IMHO it should reference the correct version

asottile:
how did you install such a python?

asottile:
weird, I've never seen this in the wild before -- is there any differentiating characteristics between it and the standard windows pythons?

asottile:
oddly, I don't have `pacman` in my msys install -- where did you install that from?

PeterMosmans:
A check you can perform is `IS_MSYS = (sys.platform == 'win32' and 'mingw' in os.__file__.lower())`. I use MSYS2 - see https://www.msys2.org/ for the platform. A great subsystem ;)

asottile:
I don't think that check is valid, you could in theory install it anywhere right?

asottile:
I am using msys, it is great, I just must have a subset of the installation heh

PeterMosmans:
Hi @asottile , I found the correct way, see https://github.com/msys2/MINGW-packages/blob/abd06ca92d876b9db05dd65f27d71c4ebe2673a9/mingw-w64-python2/0410-MINGW-build-extensions-with-GCC.patch#L54


>>> import sysconfig
>>> sysconfig.get_platform()
'mingw'

I could add this check to the code, so that this will become dependent of using MSYS or not?
I could add that check, so that it uses the correct PEP 394 naming only when on MSYS/mingw.

asottile:
ok so I installed another copy of msys and this one has pacman, but my msys python behaves differently than yours:

```console
$ python -c '
> import os
> import platform
> import sys
>
> class NA(object):
>     def __repr__(self):
>         return "<<not available>>"
>
> def peval(s):
>     try:
>         r = eval(s)
>     except AttributeError:
>         r = NA()
>     print("- **`{}`**: `{!r}`".format(s, r))
>
>
> peval("sys.version")
> peval("os.name")
> peval("sys.platform")
> peval("platform.system()")
> peval("platform.python_implementation()")
> peval("sys.implementation.name")
> # '
- **`sys.version`**: `'3.7.3 (default, Apr 25 2019, 06:50:16) \n[GCC 7.4.0]'`
- **`os.name`**: `'posix'`
- **`sys.platform`**: `'msys'`
- **`platform.system()`**: `'MSYS_NT-10.0-17134'`
- **`platform.python_implementation()`**: `'CPython'`
- **`sys.implementation.name`**: `'cpython'`
```

this is what I expect, and the current code works fine for that... so where does this msys python that claims it is win32 come from???

asottile:
I installed mine via `pacman -S python3` let me try your install method

PeterMosmans:
Do you use the MSSY2 shell, or mingw64 shell? I'm using the latter, see also the pacman command that I used. It's a bit confusing, but there are actually three shells: the pure MSYS shell, and the mingw64 as well as mingw32 shell. The MSYS shell is mainly for compiling software for Windows, the mingw{32,64} shells are more for 'normal operations'. That's probably where the confusion comes from

asottile:
I seeeeee, I was using the msys shell which is why that installation didn't do anything for me.

asottile:
alright I've updated my platforms page: https://github.com/asottile/scratch/wiki/platforms

I think the correct fix is to change the `if sys.platform == 'win32':` to `... and not msys`

You'll probably want to check the other callsites as well to make sure those are functioning correctly as well

PeterMosmans:
👍  - I'll remove my PR and will create another one. Thanks for your feedback

asottile:
you could also amend this one, that's fine too! -- all the context is here as well I'd prefer that

PeterMosmans:
Hi @asottile I updated the PR with a more fancy check for mingw (let's call it what it is, to prevent miscommunication 😉 ), and added that check to the template as well.

pre_commit/git.py
```diff
@@ -159,7 +166,8 @@ def git_path(name, repo='.'):
 
 def check_for_cygwin_mismatch():
     """See https://github.com/pre-commit/pre-commit/issues/354"""
-    if sys.platform in ('cygwin', 'win32'):  # pragma: no cover (windows)
+    if sys.platform in ('cygwin', 'win32') and \
+            not is_mingw():  # pragma: no cover (windows)
```
asottile: hmmm I think instead this needs updating for four-way checking now since any of the following mismatches are problematic:

- windows
- cygwin
- msys
- mingw

pre_commit/git.py
```diff
@@ -36,7 +38,12 @@ def no_git_env(_env=None):
 
 
 def get_root():
-    return cmd_output('git', 'rev-parse', '--show-toplevel')[1].strip()
+    return fix_mingw_path(
+        cmd_output(
+            'git', 'rev-parse',
+            '--show-toplevel',
+        )[1].strip(),
+    )
```
asottile: I don't think this should happen, if there's adjustments happening here it sounds like you're in the mismatch case (where you've got cygwin something and windows something else) -- same goes for the `hook-tmpl`, though sadly we can't warn early enough there :/

pre_commit/git.py
PeterMosmans: From your Python overview, this shouldn't happen, as in: this check should be sufficient? As far as I know, only mingwx (under MSYS2) 'is troublesome'

asottile: the check here is to ensure that `cygpath` is never needed.  If `cygpath` is needed there's a mismatch between the flavor of `git` and `python` which is not supported

PeterMosmans:
Hi @asottile do you need more information/work from me at this point? Can this be merged?

asottile:
Shoot, I totally forgot about this, I'll try it out today

asottile:
ended up being too busy today -- but I've got this marked as unread in my email so I'll definitely get to it -- maybe this weekend

pre_commit/git.py
```diff
@@ -159,7 +166,8 @@ def git_path(name, repo='.'):
 
 def check_for_cygwin_mismatch():
     """See https://github.com/pre-commit/pre-commit/issues/354"""
-    if sys.platform in ('cygwin', 'win32'):  # pragma: no cover (windows)
+    if sys.platform in ('cygwin', 'win32') and \
```
asottile: I believe this also needs to include `'msys'` for this check to be correct

Also I believe this check needs to get copied into the hook template since we start to do some `git` things in there as well

I don't think this check should include `is_mingw()` in it -- the warning is actually correct there.

asottile:
Note that I also ran into this while investigating this, not sure if it would also be a show stopper: https://github.com/pypa/virtualenv/issues/1378

asottile:
thanks for the patch, but I think the conclusion above is that mingw python is just not supported, with the suggestion being either to use the true msys/cygwin python or the windows python

PR #1079, opened 2019-07-09
Added hook-stage print to output for missing hook id

madkinsz:
Adds the value of `hook-stage` to the error message for a missing hook id to suggest setting a different stage. 

Possible additions before merging:
- Add a prompt to set the hook-stage
- Add a search for the id in different stages to say `Did you mean to use --hook-stage bleh`
- Add tests for more than `commit` and `push`

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -285,7 +285,7 @@ def run(config_file, store, args, environ=os.environ):
         ]
 
         if args.hook and not hooks:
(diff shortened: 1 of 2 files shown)


madkinsz:
Shame on me for not installing and running the pre-commit hooks! Should be fixed now.

asottile:
> Shame on me for not installing and running the pre-commit hooks! Should be fixed now.

I'm over here having a cackle :rofl:

PR #1089, opened 2019-07-20
Make autoupdate work for non-master default branches

asottile:


pre_commit/commands/autoupdate.py
```diff
@@ -36,16 +36,16 @@ def _update_repo(repo_config, store, tags_only):
     """
     repo_path = store.clone(repo_config['repo'], repo_config['rev'])
 
-    cmd_output('git', 'fetch', cwd=repo_path)
-    tag_cmd = ('git', 'describe', 'origin/master', '--tags')
+    cmd_output('git', 'fetch', 'origin', 'HEAD', '--tags', cwd=repo_path)
```
asottile: TODO: I think older versions of `git` don't support `git fetch <remote> <branch>` and I'll need a more clever solution there

asottile: seems to work fine in git 1.7.1 which is very very old so I'm calling this good enough!

sqin-wish:
Nice!

PR #1091, opened 2019-07-22
Respect the git color.ui config value when run as git hook

geieredgar:


asottile:
this doesn't seem right, we're not `git` -- we're a utility completely separate from git

PR #1092, opened 2019-07-22
Don't use color if NO_COLOR environment variable is set

### LINKED ISSUE #1073: Globally disable color?
I cannot find the way to globally disable color in the pre-commit output. Setting only the background color to green and not changing the foreground color does not work for my terminal with the following settings in the Xt resources (as set in the `${HOME}/.Xresources` file):

````properties
Rxvt.background: black
Rxvt.foreground: deepSkyBlue
````

Is there a way? It would be great to respect https://no-color.org/ environment variable. And, while we are here, maybe understand the following git config setting:

````ini
[color]
        ui = never
````

### PR DESCRIPTION
**Pull request by geieredgar**
Resolves #1073

### DIFF
--- a/pre_commit/color.py
+++ b/pre_commit/color.py
@@ -48,7 +48,7 @@ def use_color(setting):
     if setting not in COLOR_CHOICES:
         raise InvalidColorSetting(setting)
 
-    if 'NO_COLOR' in os.environ:
+    if 'NO_COLOR' in os.environ and os.environ['NO_COLOR']:
         return False
 
     return (
--- a/pre_commit/color.py
+++ b/pre_commit/color.py
@@ -48,6 +48,9 @@ def use_color(setting):
     if setting not in COLOR_CHOICES:
         raise InvalidColorSetting(setting)
 
+    if 'NO_COLOR' in os.environ:
+        return False
+
     return (
         setting == 'always' or
         (setting == 'auto' and sys.stdout.isatty() and terminal_supports_color)
--- a/pre_commit/color.py
+++ b/pre_commit/color.py
@@ -48,7 +48,7 @@ def use_color(setting):
     if setting not in COLOR_CHOICES:
         raise InvalidColorSetting(setting)
 
-    if 'NO_COLOR' in os.environ and os.environ['NO_COLOR']:
+    if os.environ.get('NO_COLOR'):
         return False
 
     return (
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pre_commit/color.py [hunk 1]
```diff
@@ -48,6 +48,9 @@ def use_color(setting):
     if setting not in COLOR_CHOICES:
         raise InvalidColorSetting(setting)
 
+    if 'NO_COLOR' in os.environ:
```
asottile: this isn't quite right -- `NO_COLOR= git commit ...` should still have color

geieredgar: @asottile Ok, I pushed a new commit

pre_commit/color.py
```diff
@@ -48,7 +48,7 @@ def use_color(setting):
     if setting not in COLOR_CHOICES:
         raise InvalidColorSetting(setting)
 
-    if 'NO_COLOR' in os.environ:
+    if 'NO_COLOR' in os.environ and os.environ['NO_COLOR']:
```
asottile: `if os.environ.get('NO_COLOR')` would be ~slightly simpler

geieredgar:
Ok, I will do that

pre_commit/color.py
geieredgar: Ok, I will change that.

tests/color_test.py
```diff
@@ -50,3 +51,20 @@ def test_use_color_tty_without_color_support():
 def test_use_color_raises_if_given_shenanigans():
     with pytest.raises(InvalidColorSetting):
         use_color('herpaderp')
+
+
+def test_no_color_env_unset():
+    with mock.patch.dict(os.environ):
```
asottile: `clear=True` is what you want here I believe

geieredgar: Wouldn't that remove all entries from `os.environ`? It would probably still work, though.

asottile: yes, and yes

I guess if you want to be more direct you could use:

```python
with envcontext([('NO_COLOR', UNSET)]):
    ...
```

`envcotext` and `UNSET` come from `pre_commit.envcontext`

geieredgar: Ok, I will change that because otherwise the coverage is reduced.

asottile:
thanks again! and thanks @saper for the issue!

Issue #1037, opened 2019-05-17
FileNotFoundError when config installed on a different path


mitzkia:
Update:
While I wrote this issue I have found that `pre-commit autoupdate` has also a `--config` option.
My question is could `autoupdate` find my previously configured (in a proper way) config file?
Or Could `autoupdate` display a better error message: 
"please use with --config=[path to config]"?

Original Issue:
Reproduction:
1, pre-commit install
```
$ pre-commit install --config=dir/subdir/.pre-commit-config.yaml
```

2, run autoupdate
```
$ pre-commit autoupdate
An unexpected error has occurred: FileNotFoundError: [Errno 2] No such file or directory: '.pre-commit-config.yaml'
```

pre-commit version: `pre-commit 1.16.1`

asottile:
this is expected, you want `pre-commit autoupdate --config=dir/subdir/.pre-commit-config.yaml`

the `--config` argument to `install` just writes that configuration into `.git/hooks/pre-commit` -- it doesn't set any state other than that

asottile:
to your point, one thing that could be improved here is to make it an argument parsing error if `--config` doesn't exist, instead of falling through to when the config gets loaded for the first time

that would be a custom `type=...` here that does the right thing: https://github.com/pre-commit/pre-commit/blob/77947f212e7b88a479dbe6feebc60a9f773e8c13/pre_commit/main.py#L49

mitzkia:
Hi, thank you for the answer and a possible solution.
I think this argument parsing solution could help here.

Thanks

asottile:
I think for now I'm going to close this and the associated PR :( -- the case is rare enough and the change potentially breaks some legitimate workflows.  thank you again for the issue and I'm sorry it didn't work out in the end <3

PR #1093, opened 2019-07-22
Check if --config option is a path to a regular file

geieredgar:
This change implements the suggestion made by @asottile in #1037 
I tried to be compatible to the current behavior of `_adjust_args_and_chdir(args)`, therefore we also check if the path points to a file relative to the git root directory.

If there is some interest in getting this merged I will add/fix the required tests.

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -42,7 +42,11 @@ def _file_path(path):
     elif os.path.exists(path):
         msg = '{} is not a regular file'.format(path)
         raise argparse.ArgumentTypeError(msg)
-    git_path = os.path.join(git.get_root(), path)
+    try:
+        git_path = os.path.join(git.get_root(), path)
+    except CalledProcessError:
+        msg = '{} does not exist'.format(path)
+        raise argparse.ArgumentTypeError(msg)
     if os.path.isfile(git_path):
         return git_path
     else:
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -58,11 +58,11 @@ def _add_color_option(parser):
     )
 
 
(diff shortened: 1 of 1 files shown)


asottile:
hmmm this is interesting, I'll have to think about this more

the interesting / challenging part of this is there's commands that accept `--config` but don't do anything with it (`gc` / `clean` / `sample-config`, and maybe probably the brand new `init-templatedir` probably should as well) -- in these cases it isn't super important that `.pre-commit-config.yaml` resolves to a file which exists, but sometimes (rarely) that argument is used just as a string

the other problem with doing more clever parsing at the `argparse` stage is we don't properly have enough error handling set up to produce a useful trace (for instance, if a command is called when outside of a `git` repository)

I'll have to put more thought into this :thinking:

geieredgar:
@asottile Now the path checking should be disabled for `gc / clean / sample-config  and init-templatedir`. 

It now also handles the case of being run outside of a git repository by catching CalledProcessError (This is rather hacky, it would be probably better to have a utility function that checks whether or not we are inside a git repo.)

asottile:
sorry I haven't gotten back to this, still thinking about whether eagerly or waiting for failure is the right approach here (especially as it invokes git more than I'd like to in the critical path -- and for things which previously didn't actually load the configuration / validate it)

I'm leaning towards closing this and wontfixing the original issue -- thank you either way for the PR though <3

asottile:
I think for now I'm going to close this and the associated issue :( -- the case is rare enough and the change potentially breaks some legitimate workflows.  thank you again for taking the time to work on this and I'm sorry it didn't work out in the end <3

PR #1094, opened 2019-07-23
Fix docker based hooks failing on Windows

### LINKED ISSUE #1072: Docker language fails on Windows - os.getuid()
[`docker_cmd`](https://github.com/pre-commit/pre-commit/blob/0cf2638b3b6a2cfde89cb7013e61b2d4a3e12875/pre_commit/languages/docker.py#L80) in languages/docker.py uses `os.getuid()` and `os.getgid()` that don't exist in Python on Windows.

A hook that uses Docker on Windows fails to execute with

```
An unexpected error has occurred: AttributeError: module 'os' has no attribute 'getuid'
  File "C:\Python36\lib\site-packages\pre_commit\error_handler.py", line 46, in error_handler
    yield
  File "C:\Python36\lib\site-packages\pre_commit\main.py", line 294, in main
    return run(args.config, store, args)
  File "C:\Python36\lib\site-packages\pre_commit\commands\run.py", line 293, in run
    return _run_hooks(config, hooks, args, environ)
  File "C:\Python36\lib\site-packages\pre_commit\commands\run.py", line 214, in _run_hooks
    retval |= _run_single_hook(classifier, hook, args, skips, cols)
  File "C:\Python36\lib\site-packages\pre_commit\commands\run.py", line 124, in _run_single_hook
    tuple(filenames) if hook.pass_filenames else (),
  File "C:\Python36\lib\site-packages\pre_commit\repository.py", line 100, in run
    return lang.run_hook(self, file_args)
  File "C:\Python36\lib\site-packages\pre_commit\languages\docker.py", line 99, in run_hook
    cmd = docker_cmd() + entry_tag + cmd_rest
  File "C:\Python36\lib\site-packages\pre_commit\languages\docker.py", line 80, in docker_cmd
    '-u', '{}:{}'.format(os.getuid(), os.getgid()),
AttributeError: module 'os' has no attribute 'getuid'
```

The same hook works fine in Windows Subsystem for Linux.

geieredgar:
Fixes #1072

Docker for Windows 
- mounts the volume passed via `-v` always using `root:root` as owner, even if a `-u UID:GID` option is passed
-  newly created files (e.g. via `touch`) also use `root:root` as owner, even if the container is run as non-root user
- read/write access is also provided to non-root users, even if `ls -l` yields `-rwxr-x-r-x`  as file permissions
- files created inside the container appear on the host as if they were created by the user running the docker command

This is because Docker for Windows currently implements host mounted volumes via SMB (see https://docs.docker.com/docker-for-windows/troubleshoot/#volumes), which handles the access to mounted files and does not allow changing permissions from inside the container (see docker/docker.github.io#3298).

Therefore the UID and GID we pass to docker can be chosen arbitrarily, because the owner of files on the host will match the Windows user running docker and not the user in the docker container.

There are also some other caveats when it comes to running linux docker images via Docker for Windows, e.g. you have to be careful about line endings (moby/moby#24388).

With these changes I was able to successfully run a hook on Windows using `language: docker_image` for a personal project of mine, therefore I would like to see something like this getting merged.  

What are your thoughts about this?

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -14,8 +14,6 @@ from pre_commit.util import cmd_output
 
 ENVIRONMENT_DIR = 'docker'
 PRE_COMMIT_LABEL = 'PRE_COMMIT'
-FALLBACK_UID = 1000
-FALLBACK_GID = 1000
 get_default_version = helpers.basic_get_default_version
 healthy = helpers.basic_healthy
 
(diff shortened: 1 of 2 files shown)


asottile:
Thanks!

PR #1113, opened 2019-08-08
rust language: use the new cargo install command

### LINKED ISSUE #1112: rust hook requires `--path` attribute
Cargo has changed how packages get installed and requires an extra `--path <destination>` attribute.

Symptom:
```
[INFO] Initializing environment for https://github.com/nix-community/nixpkgs-fmt.
[INFO] Installing environment for https://github.com/nix-community/nixpkgs-fmt.
An unexpected error has occurred: CalledProcessError: Command: ('/nix/store/fcc3x8zwq1c0667xjs7bkn6ay8j4fdpz-rust-1.38.0-nightly-2019-08-07-ad7c55e1f/bin/cargo', 'install', '--bins', '--root', '/home/zimbatm/.cache/pre-commit/repoeft6xm6t/rustenv-default')
Return code: 101
Expected return code: 0
Output: (none)
Errors: 
    error: Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.
```

I guess the fix should be done where here: https://github.com/pre-commit/pre-commit/blob/9c6a1d80d6b94c86a1785a40a51389e83accac3e/pre_commit/languages/rust.py#L87

Do we want to make pre-commit compatible with multiple versions of cargo or just the latest one?

/cc @asottile @chriskuehl

zimbatm:
cargo install now requires an additional `--path <TARGET>` argument.

Fixes #1112

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -73,7 +73,7 @@ def install_environment(prefix, version, additional_dependencies):
         _add_dependencies(prefix.path('Cargo.toml'), lib_deps)
 
     with clean_path_on_failure(directory):
-        packages_to_install = {()}
+        packages_to_install = {('--path', '.')}
         for cli_dep in cli_deps:
             cli_dep = cli_dep[len('cli:'):]
             package, _, version = cli_dep.partition(':')


asottile:
> E               error: `/tmp/pytest-of-vsts/pytest-0/test_rust_hook0/0/.pre-commit/repoy25Xd6/rustenv-default` is not a directory. --path must point to a directory containing a Cargo.toml file.


hmm ok after reading [the docs](https://doc.rust-lang.org/cargo/commands/cargo-install.html) it seems that `--path` should be used when installing `.` essentially (and not needed otherwise) -- so I think `--path .` is what we want here?

zimbatm:
I'm not sure which path is which. Let's see what the CI has to say.

asottile:
well now I have no idea, maybe @chriskuehl knows better?

PR #1117, opened 2019-08-15
Stabilize python default version lookup

scop:
For example, for sys.executable:
>   /usr/bin/python3 -> python3.7

...the default lookup may return either `python3` or `python3.7`. Make the order deterministic by iterating over tuple, not set, of candidates.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -43,14 +43,13 @@ def _find_by_py_launcher(version):  # pragma: no cover (windows only)
             pass
 
 
-def _get_default_version():  # pragma: no cover (platform dependent)
+def _find_by_sys_executable():
     def _norm(path):
         _, exe = os.path.split(path.lower())
         exe, _, _ = exe.partition('.exe')
         if find_executable(exe) and exe not in {'python', 'pythonw'}:
             return exe
 
-    # First attempt from `sys.executable` (or the realpath)
     # On linux, I see these common sys.executables:
     #
     # system `python`: /usr/bin/python -> python2.7
@@ -59,10 +58,17 @@ def _get_default_version():  # pragma: no cover (platform dependent)
     # virtualenv v -ppython2: v/bin/python -> python2
     # virtualenv v -ppython2.7: v/bin/python -> python2.7
     # virtualenv v -ppypy: v/bin/python -> v/bin/pypy
-    for path in {sys.executable, os.path.realpath(sys.executable)}:
+    for path in (sys.executable, os.path.realpath(sys.executable)):
         exe = _norm(path)
         if exe:
             return exe
+    return None
+
+
+def _get_default_version():  # pragma: no cover (platform dependent)
+
+    # First attempt from `sys.executable` (or the realpath)
+    exe = _find_by_sys_executable()
 
     # Next try the `pythonX.X` executable
     exe = 'python{}.{}'.format(*sys.version_info)
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -66,7 +66,6 @@ def _find_by_sys_executable():
 
 
 def _get_default_version():  # pragma: no cover (platform dependent)
-
     # First attempt from `sys.executable` (or the realpath)
     exe = _find_by_sys_executable()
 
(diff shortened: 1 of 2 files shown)


scop:
An effect of the instability is that pre-commit will end up installing dependencies several times in separate redundant virtualenvs over multiple runs, depending on which version happened to be returned.

scop:
Don't know what's up with the tests, they pass for me in local tox on Linux.

asottile:
as written, those tests require everything to be on `PATH` which won't be true when we test individual pythons

they pass locally for you because you have all of those pythons available

patch looks fine -- even a patch which just changes the brackets would probably be fine to merge if this is difficult to test

tests/languages/python_test.py
```diff
@@ -44,6 +44,10 @@ def test_sys_executable_matches_does_not_match(v):
     ),
 )
 def test_find_by_sys_executable(exe, realpath, expected):
+    def mocked_find_executable(exe):
+        return exe.rpartition('/')[2]
     with mock.patch.object(sys, 'executable', exe):
         with mock.patch('os.path.realpath', return_value=realpath):
-            assert python._find_by_sys_executable() == expected
+            with mock.patch('pre_commit.parse_shebang.find_executable',
```
asottile: please use `mock.patch.object` instead, `mock.patch` has import side-effects and can lead to some nasty patch leaking issues.  `mock.patch.object` is also more explicit about the target

also, that feel when you don't run the pre-commit hooks while contributing to pre-commit :laughing:

tests/languages/python_test.py
```diff
@@ -32,3 +32,24 @@ def test_sys_executable_matches(v):
 def test_sys_executable_matches_does_not_match(v):
     with mock.patch.object(sys, 'version_info', (3, 6, 7)):
         assert not python._sys_executable_matches(v)
+
+
+@pytest.mark.parametrize(
+    'exe,realpath,expected', (
```
asottile: would prefer a tuple of strings here instead of stringly typed parametrize as well

pre_commit/languages/python.py
```diff
@@ -59,10 +58,17 @@ def _norm(path):
...
-    for path in {sys.executable, os.path.realpath(sys.executable)}:
+    for path in (sys.executable, os.path.realpath(sys.executable)):
         exe = _norm(path)
         if exe:
             return exe
+    return None
+
+
+def _get_default_version():  # pragma: no cover (platform dependent)
+
```
asottile:
```suggestion
```

tests/languages/python_test.py
asottile:
```suggestion
    ('exe', 'realpath', 'expected'), (
```

scop:
Thanks, learned something new :) And I hope I caught all of it now.

pre_commit/languages/python.py
```diff
@@ -59,10 +58,16 @@ def _norm(path):
...
+    for path in (sys.executable, os.path.realpath(sys.executable)):
         exe = _norm(path)
         if exe:
             return exe
+    return None
+
+
+def _get_default_version():  # pragma: no cover (platform dependent)
+    # First attempt from `sys.executable` (or the realpath)
+    exe = _find_by_sys_executable()
```
asottile: this is broken, the value is immediately discarded

tests/languages/python_test.py
```diff
@@ -7,6 +7,7 @@
 import mock
 import pytest
 
+import pre_commit.parse_shebang
```
asottile:
```suggestion
from pre_commit import parse_shebang
```

tests/languages/python_test.py
```diff
@@ -32,3 +33,24 @@ def test_sys_executable_matches(v):
...
+        ('v/bin/python', 'v/bin/pypy', 'pypy'),
+    ),
+)
+def test_find_by_sys_executable(exe, realpath, expected):
+    def mocked_find_executable(exe):
+        return exe.rpartition('/')[2]
+    with mock.patch.object(sys, 'executable', exe):
+        with mock.patch.object(os.path, 'realpath', return_value=realpath):
+            with mock.patch.object(
+                pre_commit.parse_shebang, 'find_executable',
```
asottile:
```suggestion
                parse_shebang, 'find_executable',
```

pre_commit/languages/python.py
scop: Wow. Nice "little detail" crept in after refactoring that out to a function for tests.

scop:
Damn, still not quite there it seems :/

PR #1118, opened 2019-08-15
Add completions for CLI

lewisacidic:
Uses[ `argcomplete`](https://github.com/kislyuk/argcomplete) to trivially add basic completions to the cli.

This allows things like:

```
$ pre-commit <TAB>
--help     -V          gc                install         run            uninstall
--version  autoupdate  help              install-hooks   sample-config
-h         clean       init-templatedir  migrate-config  try-repo
```

And
```
$ pre-commit install --hook-type <TAB>
commit-msg  pre-commit  pre-push  prepare-commit-msg
```

Which can be tab toggled through once the correct completions are registered (at least in `fish`, I didn't try other shells yet).

This is pretty basic (as might be expected from adding 2 simple lines of code!!!), it would be really nice to do a bit more, such as getting a list of the available hooks and completing `pre-commit run <TAB>` with them.

Also, at least with `click` CLIs, you should be able to get help messages showing up (in `fish` at least) which is nicer than this. Of course, this would require a rewrite of the parser code to use click...

Any opinions about this sort of functionality?

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -5,6 +5,8 @@ import logging
 import os
 import sys
 
+import argcomplete
+
 import pre_commit.constants as C
 from pre_commit import color
 from pre_commit import five
@@ -268,6 +270,9 @@ def main(argv=None):
     # argparse doesn't really provide a way to use a `default` subparser
     if len(argv) == 0:
         argv = ['run']
+
+    argcomplete.autocomplete(parser)
+
     args = parser.parse_args(argv)
 
     if args.command == 'help' and args.help_cmd:
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,6 +35,7 @@ install_requires =
     virtualenv>=15.2
     futures; python_version<"3.2"
     importlib-resources; python_version<"3.7"
+    argcomplete
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
 
 [options.entry_points]


asottile:
sorry, don't really want to add a dependency for this

asottile:
also the strategy used by argcomplete is too slow to be useful, it adds ~200ms per `<tab>` press which is well past the noticeable "slowness"

I believe the more correct way to do something like this would be to pregenerate completions as part of the packaging and provide those separately

lewisacidic:
Thanks for the quick reply.

In my opinion completions are pretty important for friendly, usable CLI tools. I understand pre-commit isn't first and foremost a CLI tool, but I found I missed completions when I was trying to remember how to install a commit-msg hook just now (typing `-h` is too much for me :))

I guess we could write completions by hand for the relatively small number of available commands and not need any new dependencies, but this is definitely the easiest and most integrated way to provide completions.

asottile:
it's not particularly difficult to generate completions scripts, here's some prior art: https://github.com/asottile/all-repos/blob/master/all_repos/complete.py

Would expose some command like:

```bash
eval "$(pre-commit complete)"
```

which you could put in your rc file

lewisacidic:
Just seen your followups, yep it's laggy (tbh I'm using WSL so I'm used to laggy terminals, didn't really notice...). I think the lag could be alleviated with a refactor to avoid importing before parsing CLIs, I've used this style of autocompletion before and it worked a charm. 

In any case, I don't really have the time to invest into writing anything substantial right now, shall I close this and open an issue about completions instead?

asottile:
yeah I don't really want to play the import whackamole game -- there's not a good reason to runtime-compute completions when we can generate them all statically

PR #1142, opened 2019-09-20
Add version information to error logs

### LINKED ISSUE #1141: include version information in error log
would be useful to include things like:

- pre-commit version
- sys.version
- sys.executable

marqueewinq:
Added pre-commit version, `sys.version` and `sys.executable` after the stack traces in error messages. 

#1141

### DIFF
--- a/pre_commit/error_handler.py
+++ b/pre_commit/error_handler.py
@@ -28,11 +28,6 @@ def _to_bytes(exc):
 
 def _log_and_exit(msg, exc, formatted):
     error_msg = b''.join((
-        _to_bytes('### version information\n'),
-        _to_bytes('pre-commit.version={}\n'.format(C.VERSION)),
-        _to_bytes('sys.version={}\n'.format(sys.version.replace('\n', ' '))),
-        _to_bytes('sys.executable={}\n'.format(sys.executable)),
-        _to_bytes('### error information\n'),
         five.to_bytes(msg), b': ',
         five.to_bytes(type(exc).__name__), b': ',
         _to_bytes(exc), b'\n',
@@ -41,9 +36,26 @@ def _log_and_exit(msg, exc, formatted):
     store = Store()
     log_path = os.path.join(store.directory, 'pre-commit.log')
     output.write_line('Check the log at {}'.format(log_path))
+
+    meta_info_msg = '### version information\n```\n'
+    meta_info_msg += 'pre-commit.version: {}\n'.format(C.VERSION)
+    meta_info_msg += 'sys.version: \n{}\n'.format(
+        '\n'.join(
+            [
+                '\t{}'.format(line)
+                for line in sys.version.splitlines()
+            ],
+        ),
+    )
+    meta_info_msg += 'sys.executable: {}\n'.format(sys.executable)
+    meta_info_msg += 'os.name: {}\n'.format(os.name)
+    meta_info_msg += 'sys.platform: {}\n```\n'.format(sys.platform)
+    meta_info_msg += '### error information\n```\n'
     with open(log_path, 'wb') as log:
+        output.write(meta_info_msg, stream=log)
         output.write(error_msg, stream=log)
         output.write_line(formatted, stream=log)
+        output.write('\n```\n', stream=log)
     raise SystemExit(1)
 
 
--- a/pre_commit/error_handler.py
+++ b/pre_commit/error_handler.py
@@ -37,22 +37,32 @@ def _log_and_exit(msg, exc, formatted):
     log_path = os.path.join(store.directory, 'pre-commit.log')
     output.write_line('Check the log at {}'.format(log_path))
 
-    meta_info_msg = '### version information\n```\n'
-    meta_info_msg += 'pre-commit.version: {}\n'.format(C.VERSION)
-    meta_info_msg += 'sys.version: \n{}\n'.format(
-        '\n'.join(
-            [
-                '\t{}'.format(line)
-                for line in sys.version.splitlines()
-            ],
-        ),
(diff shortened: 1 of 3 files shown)


pre_commit/error_handler.py
```diff
@@ -4,12 +4,14 @@
...
 import contextlib
 import os.path
+import sys
 import traceback
 
 import six
 
 from pre_commit import five
 from pre_commit import output
+from pre_commit.constants import VERSION as pre_commit_version
```
asottile: iirc the rest of the codebase uses `import pre_commit.constants as C` and then `C.VERSION`

pre_commit/error_handler.py
```diff
@@ -29,6 +31,9 @@ def _log_and_exit(msg, exc, formatted):
         five.to_bytes(msg), b': ',
         five.to_bytes(type(exc).__name__), b': ',
         _to_bytes(exc), b'\n',
+        _to_bytes('pre-commit.version={}\n'.format(pre_commit_version)),
+        _to_bytes('sys.version={}\n'.format(sys.version.replace('\n', ' '))),
+        _to_bytes('sys.executable={}\n'.format(sys.executable)),
```
asottile: let's put these at the top, perhaps something like this output:

```markdown
    ### version information
    ```
    pre-commit: {...}
    sys.executable: {...}
    sys.version:
        {...}
    ```

    ### error information
    ...
```

pre_commit/error_handler.py
```diff
@@ -29,6 +31,9 @@ def _log_and_exit(msg, exc, formatted):
...
     ))
     output.write(error_msg)
```
asottile: the version info should only go to the log file, right now it's being printed to the output as well

tests/error_handler_test.py
```diff
@@ -104,17 +104,29 @@ def test_log_and_exit(cap_out, mock_store_dir):
 
     printed = cap_out.get()
     log_file = os.path.join(mock_store_dir, 'pre-commit.log')
-    assert printed == (
-        'msg: FatalError: hai\n'
-        'Check the log at {}\n'.format(log_file)
-    )
+    printed_lines = printed.split('\n')
+    assert len(printed_lines) == 6, printed_lines
```
asottile: with pytest you can usually leave out the `, printed_lines` bit -- and pytest'll generate a cascading assertion message for you 👍

tests/error_handler_test.py
```diff
@@ -136,7 +148,7 @@ def test_error_handler_no_tty(tempdir_factory):
         pre_commit_home=pre_commit_home,
     )
     log_file = os.path.join(pre_commit_home, 'pre-commit.log')
-    assert output[1].replace('\r', '') == (
-        'An unexpected error has occurred: ValueError: ☃\n'
-        'Check the log at {}\n'.format(log_file)
-    )
+    output_lines = output[1].replace('\r', '').split('\n')
```
asottile: `.splitlines()` is usually a better alternative to `.split('\n')`

pre_commit/error_handler.py
marqueewinq: Can you please explain the motivation behind this? My thought were that such information is useful when creating an issue/debugging custom hooks, so copy-paste from console would be a common use case

marqueewinq:
@asottile Thank you for thorough comments, i've patched those small things.

I'm still not sure about not putting meta in both output and logs; i'll fix that if you insist, but i'm curious about the motivation.

pre_commit/error_handler.py
marqueewinq: The format is:
```
### version information
pre-commit.version=...
sys.version=...
sys.executable=...
### error information
An unexpected error has occurred: ...
Check the log at ...
```

pre_commit/error_handler.py
asottile: the error output is supposed to be intentionally brief (don't show debugging info / stacktraces to users) with the actual meat of the bug-reportable information being in the error log file

pre_commit/error_handler.py
```diff
@@ -26,6 +28,11 @@ def _to_bytes(exc):
 
 def _log_and_exit(msg, exc, formatted):
     error_msg = b''.join((
+        _to_bytes('### version information\n'),
```
asottile: for the literal strings you can just use `b'### version information\n'`

pre_commit/error_handler.py
```diff
@@ -26,6 +28,11 @@ def _to_bytes(exc):
...
+        _to_bytes('pre-commit.version={}\n'.format(C.VERSION)),
```
asottile: instead of the (slightly cramped) `=...` let's use a `: `

pre_commit/error_handler.py
```diff
@@ -26,6 +28,11 @@ def _to_bytes(exc):
...
+        _to_bytes('sys.version={}\n'.format(sys.version.replace('\n', ' '))),
```
asottile: for sys.version, since it can contain newlines, let's instead put the version information on the next line and indent it (instead of removing the newlines)

pre_commit/error_handler.py
```diff
@@ -26,6 +28,11 @@ def _to_bytes(exc):
...
+        _to_bytes('sys.executable={}\n'.format(sys.executable)),
+        _to_bytes('### error information\n'),
```
asottile: while we're at it (since we're markdown formatting this) let's put fixed-width code braces around the two chunks of output so it's easier to copy paste the error log into an issue (I hinted at this in the previous review but probably didn't explicitly ask for it there -- my b)

PR #1145, opened 2019-09-23
Allow --hook-type to be specified multiple times

### LINKED ISSUE #1139: Install many hooks types by one command
Now I should run next in every new repo:
```console
$ pre-commit install-hooks
$ pre-commit install --install-hooks --hook-type commit-msg
```

Maybe can be more pretty have something like 
```console
$ pre-commit install --install-hooks --all
# OR
$ pre-commit install --install-hooks --hook-type pre-commit,commit-msg
```

asottile:
Resolves #1139

took this opportunity to also make `hook_types` required to make the api less pre-commit-hook-specific

### DIFF
--- a/pre_commit/commands/init_templatedir.py
+++ b/pre_commit/commands/init_templatedir.py
@@ -8,10 +8,10 @@ from pre_commit.util import cmd_output
 logger = logging.getLogger('pre_commit')
 
 
-def init_templatedir(config_file, store, directory, hook_type):
+def init_templatedir(config_file, store, directory, hook_types):
     install(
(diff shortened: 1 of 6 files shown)


NickLarsenNZ:
@asottile (you locked the other issue #1186 so I couldn't respond there). Thanks for the hint to this PR (I just missed it in the noise of dup issues.

Anyway, was looking to see if there was an option to just install all (as has been requested by others), but it looks like there isn't. :shrug:

asottile:
I would hope you would have taken the hint that grave digging 5 year old issues is not welcome, but I guess not

Issue #1159, opened 2019-10-06
Misleading error message when entry script is present but not executable


nixjdm:
Hi! I'm getting acquainted with pre-commit (I love it by the way). I tried adding a prepare-commit-msg script with

```yaml
-   repo: local
    hooks:
      - id: some-id
        name: Some Name
        stages: [prepare-commit-msg]
        entry: ./bin/prepare-commit-msg.sh
        language: script
```

At first I neglected to `chmod +x bin/prepare-commit-msg.sh`, so when I tried it out, I got the message

```
Executable `[path]/bin/prepare-commit-msg.sh` not found
```

The file was present, it just wasn't marked executable. I squinted for a short while looking for a typo somewhere before realizing my mistake.

asottile:
would you like to submit a PR for this? it should be an easy patch + test for this function here: https://github.com/pre-commit/pre-commit/blob/ab063977ad45ead2051deefb47f01b9ea3c40641/pre_commit/parse_shebang.py#L54-L55

PR #1161, opened 2019-10-07
Handle case when executable is not executable

WillKoehrsen:
- Changed error message if executable is not executable

Closes:[1159](https://github.com/pre-commit/pre-commit/issues/1159)

### DIFF
--- a/pre_commit/parse_shebang.py
+++ b/pre_commit/parse_shebang.py
@@ -51,10 +51,12 @@ def normexe(orig):
         if exe is None:
             _error('not found')
         return exe
-    elif not os.access(orig, os.X_OK):
-        _error('not found')
     elif os.path.isdir(orig):
         _error('is a directory')
+    elif not os.path.isfile(orig):
+        _error('not found')
+    elif not os.access(orig, os.X_OK):  # pragma: windows no cover
+        _error('is not executable')
     else:
         return orig
 
--- a/tests/parse_shebang_test.py
+++ b/tests/parse_shebang_test.py
@@ -91,6 +91,14 @@ def test_normexe_does_not_exist_sep():
     assert excinfo.value.args == ('Executable `./i-dont-exist-lol` not found',)
 
 
+@pytest.mark.xfail(os.name == 'nt', reason='posix only',)
+def test_normexe_not_executable(tmpdir):  # pragma: windows no cover
+    tmpdir.join('exe').ensure()
+    with tmpdir.as_cwd(), pytest.raises(OSError) as excinfo:
+        parse_shebang.normexe('./exe')
+    assert excinfo.value.args == ('Executable `./exe` is not executable',)
+
+
 def test_normexe_is_a_directory(tmpdir):
     with tmpdir.as_cwd():
         tmpdir.join('exe').ensure_dir()


pre_commit/parse_shebang.py
```diff
@@ -52,7 +52,7 @@ def _error(msg):
             _error('not found')
         return exe
     elif not os.access(orig, os.X_OK):
-        _error('not found')
+        _error('not executable')
```
asottile: there should be a new case here, one for "does not exist" and one for "not executable"

the test you modified shows that this change is incorrect in fact

also you'd use `_error('is not executable')`

WillKoehrsen: Thanks for the review. So the check for existence would be:
```
elif not os.access(orig, os.F_OK)
```
and the check for executability would be:
```
elif not os.access(orig.os.F_OK)
```
Using the following mode codes:
```
os.F_OK: Tests existence of the path.
os.R_OK: Tests readability of the path.
os.W_OK: Tests writability of the path.
os.X_OK: Checks if path can be executed.
```

asottile: `os.path.isfile` is the usual way to check for a file

tests/parse_shebang_test.py
```diff
@@ -88,7 +88,21 @@ def test_normexe_does_not_exist():
 def test_normexe_does_not_exist_sep():
     with pytest.raises(OSError) as excinfo:
         parse_shebang.normexe('./i-dont-exist-lol')
-    assert excinfo.value.args == ('Executable `./i-dont-exist-lol` not found',)
+    assert excinfo.value.args == (
+        'Executable `./i-dont-exist-lol` not found',
+    )
```
asottile: this test shouldn't change -- can you revert the changes there?

tests/parse_shebang_test.py
```diff
@@ -88,7 +88,21 @@ def test_normexe_does_not_exist():
...
+
+
+def test_normexe_not_executable():
+    exe_path = os.path.abspath(
+        write_executable('/usr/bin/env sh', filename='run.myext'),
+    )
```
asottile: this pollutes the working directory, can you use `tmpdir` to do this?

tests/parse_shebang_test.py
```diff
@@ -88,7 +88,21 @@ def test_normexe_does_not_exist():
...
+def test_normexe_not_executable(in_tmpdir):
+    exe_path = os.path.abspath(
+        write_executable('/usr/bin/env sh', filename='run.myext'),
+    )
+    os.chmod(exe_path, 444)
+    with pytest.raises(OSError) as excinfo:
+        parse_shebang.normexe(exe_path)
+    assert excinfo.value.args == (
+        'Executable `{}` is not executable'.format(exe_path),
+    )
```
asottile: I suspect you might need to skip this test on windows given the way executables work there

you can use `# pragma: windows no cover` this test (and probably the code branch as well if it isn't going to be hit)

pre_commit/parse_shebang.py
WillKoehrsen: `os.path.isfile` will also catch a directory which is covered in `elif os.path.isdir(orig)`.

asottile: you may need to reorder the checks then

tests/parse_shebang_test.py
```diff
@@ -91,6 +91,18 @@ def test_normexe_does_not_exist_sep():
     assert excinfo.value.args == ('Executable `./i-dont-exist-lol` not found',)
 
 
+def test_normexe_not_executable(in_tmpdir):
+    exe_path = os.path.abspath(
+        write_executable('/usr/bin/env sh', filename='run.myext'),
+    )
+    os.chmod(exe_path, 444)
```
asottile: simpler:

```python
in_tmpdir.join('exe').ensure()
with pytest.raises(OSError) as excinfo:
    parse_shebang.normexe('./exe')
assert excinfo.value.args == ('Executable `./exe` is not executable',)
```

WillKoehrsen: Hmm, I get `AttributeError: 'str' object has no attribute 'ensure'` when I try to run that test.

asottile: ah right, this is before I switched to using the `tmpdir` fixture more -- so instead:

```python
def test_...(tmpdir):
    tmpdir.join('exe').ensure()
    with tmpdir.as_cwd(), pytest.raises(OSError) as excinfo:
        parse_shebang.normexe('./exe')
    assert excinfo.value.args == ('Executable `./exe` is not executable',)
```

tests/parse_shebang_test.py
```diff
@@ -91,6 +91,13 @@ def test_normexe_does_not_exist_sep():
     assert excinfo.value.args == ('Executable `./i-dont-exist-lol` not found',)
 
 
+def test_normexe_not_executable(tmpdir):
```
asottile: you'll need `@pytest.mark.xfail(os.name == 'nt', reason='posix only')  # pragma: windows no cover`

sorry forgot about the test-skipping part

tests/parse_shebang_test.py
```diff
@@ -91,6 +91,16 @@ def test_normexe_does_not_exist_sep():
     assert excinfo.value.args == ('Executable `./i-dont-exist-lol` not found',)
 
 
+@pytest.mark.xfail(
+    os.name == 'nt', reason='posix only',
+)  # pragma: windows no cover
```
asottile: here's another way to write this, nbd though:

```diff
diff --git a/tests/parse_shebang_test.py b/tests/parse_shebang_test.py
index 3c48a11..fe1cdcd 100644
--- a/tests/parse_shebang_test.py
+++ b/tests/parse_shebang_test.py
@@ -91,10 +91,8 @@ def test_normexe_does_not_exist_sep():
     assert excinfo.value.args == ('Executable `./i-dont-exist-lol` not found',)
 
 
-@pytest.mark.xfail(
-    os.name == 'nt', reason='posix only',
-)  # pragma: windows no cover
-def test_normexe_not_executable(tmpdir):
+@pytest.mark.xfail(os.name == 'nt', reason='posix only')
+def test_normexe_not_executable(tmpdir):  # pragma: windows no cover
     tmpdir.join('exe').ensure()
     with tmpdir.as_cwd(), pytest.raises(OSError) as excinfo:
         parse_shebang.normexe('./exe')
```

WillKoehrsen:
@asottile Thanks for coaching me through this one! 👍

asottile:
> @asottile Thanks for coaching me through this one! +1

happy to help! thanks for the contribution!

Issue #823, opened 2018-09-05
Question: Is there an option to only output on Failed hook


graemedavidson:
Is it possible to have it only output on Failure, otherwise leaving it clear.

Might be a good additional option (if it does not exist already) to add at the parent level a verbosity setting in which you can get the current Passed/Skipped/Failed output at the highest. a mid level providing a single line of stats --passed=5 --skipped=2 and the above with only failures.

asottile:
That's the default behaviour unless you're using `--verbose` or `verbose: true`

asottile:
no, it should always show the passed messaging

graemedavidson:
Ok, so my question becomes a feature request.

On 5 Sep 2018 18:56, "Anthony Sottile" <notifications@github.com> wrote:

no, it should always show the passed messaging

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/pre-commit/pre-commit/issues/823#issuecomment-418822945>,
or mute the thread

asottile:
What's the motivation? My initial lean is -0

asottile:
Unless you have a good reason I don't want to or plan on implementing this.

Comment back and I'll reopen though :)

aschokking:
I was looking for an option like this (silence the passed hooks) so that there's less noise in my output to scroll through.

asottile:
CC @blueyed

Even if a commandline option were added (which I'm against -- the output is almost a _branding_ at this point), there wouldn't be a way to set it since there aren't ways to pass through arguments from `git`

asottile:
oops, meant to reopen -- should be open now. let's discuss.

blueyed:
I would (like to) configure this in the config file.

blueyed:
So I assume there is no user config then?
What about an environment variable?

blueyed:
A git config option could be used for this.

asottile:
I guess another option would be to set it during `pre-commit install` I forgot about this but we already have a precedent for an option there: https://github.com/pre-commit/pre-commit/blob/518a72d7e7598b71105fb764d990438c1b6d485a/pre_commit/resources/hook-tmpl#L20

blueyed:
@asottile 
Also sounds good.

I came up with the following for me:
```diff
--- .git/hooks/pre-commit	2019-03-02 15:25:46.671907241 +0100
+++ .git/hooks/pre-commit.new	2019-03-02 15:25:35.715359949 +0100
[...]

PR #1172, opened 2019-10-15
WIP: Added config parameter hide_skipped

allburov:
Fix https://github.com/pre-commit/pre-commit/issues/823 

@asottile I thing passing and failing messages are very usefull (and a little hard to hide, cause is hook's name has printed before runs, so we need clear stdout... ).

Let's just hide the `skipped` messages from output!
What I also have to do? 
1. cli arguments?
2. Env support?
3. Change https://pre-commit.com/ and other docs?


asottile:
please discuss in the issue first, I don't want you to spend time on a feature if I'm not going to accept it

allburov:
Oh, of course, I hurry up a little

Issue #1099, opened 2019-07-23
functionality to add files to commit


lamourj:
It would be a nice feature to have the option to automatically stage new files at the end of the precommit hook.
Use case: I use a hook to generate documentation at commit time. It would be nice to have the option to add these files to the current commit, if modified. 
Is there any way to do that currently ? I guess a workaround would be to add a hook at the end which would add these files... any nicer way ?
Thanks

asottile:
pre-commit will never ever ever modify the staging area, that's not a safe operation and one of the worst anti-features of related tools

See also #1086 #879 #806 #999 #747 (please search the issue tracker next time!)

Issue #1086, opened 2019-07-20
Option for auto-commiting hook-made changes


ArturGajowy:
The docs say:
`The hook must exit nonzero on failure or modify files in the working directory (since 0.6.3).`

That is also the observed behavior. In case of a formatter hook though, that means most commits need to be attempted twice: once to re-format them, twice to commit them successfully.

For a sufficiently mature formatter tool, it should be OK to just set sth like `auto_commit_changes: true` in its hook config. This issue is about adding such an option to pre-commit. WDYT?

asottile:
pre-commit will never ever ever modify the staging area, that's not a safe operation and one of the worst anti-features of related tools

See also #879 #806 #999 #747 (please search the issue tracker next time!)

ArturGajowy:
Thanks, I found your remark in one of those issues helpful: I can always make my hook do `run_formatter && git add -u`.

(I did search the tracker)

Issue #999, opened 2019-04-19
Amend commit support


domenkozar:
There are some checks that are automatic. We're using brittany (Haskell source formatter) and my workflow goes always as following:

1) make changes
2) commit
3) pre-commit failed
4) amend commit
5) commit again

I was wondering if 3 and 4 could be merged into pre-commit showing the diff and then asking to amend it?

asottile:
It could, however pre-commit currently doesn't do either of the two things:
- modify the staging area
- have interactivity

The second is difficult as well because git hooks are not handed a tty and would have to (unreliably) fish one out of `/dev/tty` (not even sure this is possible on windows)

I use:

```bash
git add -u && !!
```

quite often fwiw

domenkozar:
I guess interactivity was a stretch goal, I would be happy with just "accept the output" as there's nothing one can really do against autoformatter.

I presume modifying the staging area is doable?

asottile:
one could `SKIP` / etc. to bypass the auto formatter.

pre-commit takes a strong stance on not modifying the staging area -- here's more reading on that subject: #747 #806 #879

domenkozar:
Alright :)

asottile:
thanks for the issue though!  :tada:

Issue #879, opened 2018-12-07
Add changes generated by pre-commit


henniss:
Loosely related to #806 . 

I sometimes find myself doing the following: 

* Edit some files. 
* Run `git add -p` and stage only some of my edits. 
* Run `git commit -m "Some message"`. 
  * pre-commit stashes all the unstaged changes,
  * runs all hooks, which change the working tree,
  * pops the stashed edits,
  * and rejects the commit.

The changes generated by the pre-commit hooks are now intermixed with the unstaged edits. I can't just run `git add -u && !!`, as that will stage too much. Instead I need to run `git add -p` and select only the unstaged changes generated by my hooks. 

Could pre-commit generate a patch of changes it generates and store that patch in `.git`? Then I could do something like the following:

```bash
git add -p
git commit -m "Some message."
cat .git/pre-commit/patch
git apply --index .git/pre-commit/patch && !-2
```

Or is there already a way to do this?

asottile:
as you've learned from #806 and #747 -- pre-commit will not modify your staging area since this is judged as an unsafe operation

as for generating a patch, there is one which is stored in `~/.cache/pre-commit` which represents the parts pre-commit unstages.  You're free to reverse-apply that, then staged, then commit, then apply the patch again and you'll be back to your desired state.

Let me know if this works for you!

Here's a sample workflow for that:

```diff
$ git diff --staged
diff --git a/t b/t
index f00c965..dc76404 100644
--- a/t
+++ b/t
@@ -1,6 +1,6 @@
 1
 2
-3
+3 # trailing whitespace 
 4
 5
 6
$ git diff
diff --git a/t b/t
index dc76404..8e6f928 100644
--- a/t
+++ b/t
@@ -4,7 +4,7 @@
 4
 5
 6
-7
+7 # don't stage me
 8
 9
 10
```

And when I commit:

```
$ git commit -m "hi"
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/asottile/.cache/pre-commit/patch1544222126.
Trim Trailing Whitespace.................................................Failed
hookid: trailing-whitespace

Files were modified by this hook. Additional output:

Fixing t

Fix End of Files.........................................................Passed
Check Yaml...........................................(no files to check)Skipped
Check for added large files..............................................Passed
[INFO] Restored changes from /Users/asottile/.cache/pre-commit/patch1544222126.
```

Now I can remove the parts I didn't want to commit:

```
[...]

PR #1188, opened 2019-10-24
Added two new options that allow for file modifications

mmabey:
This PR introduces two new options to pre-commit that do not change the program's default behavior, but allow users to either specify command-line options or set environment variables to do two things:

1. `--modified-files-ok` / `PRE_COMMIT_MODIFIED_FILES_OK`: When specified/set, pre-commit no longer classifies it as a failure when hooks have a non-zero exit status or when hooks modify files (the former is required to accommodate the latter). What this means in practice is that when pre-commit is invoked from a `git commit` command, it will not prevent the commit from proceeding if a hook modifies a file.

2. `--add-modified-files` / `PRE_COMMIT_ADD_MODIFIED_FILES`: When specified/set, this first activates the `--modified-files-ok` option and then if any files are modified by hooks, pre-commit stages those modifications. This means that when a user invokes pre-commit as part of a `git commit` command, modifications made by hooks will be included in the commit and the commit command will be allowed to continue. The down side of activating this option is that if there are any un-staged changes (stashed by pre-commit) that conflict with the changes made by the hooks, then the stashed changes can be lost. I would welcome feedback or suggestions on how to prevent this.

The objective of this work is to make using pre-commit a more seamless experience. I recently suggested to my co-workers that we start using pre-commit, and I had one particularly grumpy co-worker strongly object based on the fact that if the hooks made modifications, he would be required to manually stage those modifications and execute `git commit` a second time. I'm hoping that many others will see value in the simplified workflow enabled by these options just like I hope my co-worker doesn't have anything else to protest about... one can dream, right?

I've added the necessary changes to the test cases to ensure these changes don't break any existing tests. I've also used these options and verified they function as expected.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -129,7 +129,7 @@ def _run_single_hook(classifier, hook, args, skips, cols, use_color):
     diff_after = cmd_output_b('git', 'diff', '--no-ext-diff', retcode=None)
 
     file_modifications = diff_before != diff_after
-    modified_ok = args.modified_files_ok
+    modified_ok = args.modified_files_ok or args.add_modified
 
     # If the hook makes changes, fail the commit
     if not modified_ok and file_modifications:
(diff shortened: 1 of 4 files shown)


asottile:
nope, sorry this is intentionally not implemented -- please read the relevant issues and next time search the issue tracker and make an issue before implementing large features #1099 #1086 #879 #806 #999 #747 (please search the issue tracker next time!)

Issue #1193, opened 2019-10-28
ruby-build  tar is out of date so you can't use ruby versions later than 2016


choffee:
Can the ruby-build.tar in resources be updated?

asottile:
They absolutely can, would you like to?

Note that prebuilt versions are not limited by the `ruby-build` bits so you can usually get newer ones as long as you [pick one that exists](https://rvm.io/binaries/)

choffee:
Fair play :) I've added in the latest build from their github.

RVM stopped building debian binaries a few years ago so I always have to compile.

asottile:
via #1195

asottile:
this has been released as part of 1.20 -- thanks again for the issue and patch!

PR #1195, opened 2019-10-28
Update ruby-build to latest version

choffee:
This is just the latest version of ruby-build tarball so that you can install some of the latest rubies if they are not available for binary download.

Related to #1193


asottile:
These are built with a script, could you please update the revisions of the artifacts there? https://github.com/pre-commit/pre-commit/blob/master/pre_commit/make_archives.py

choffee:
That's probably not the right git way of doing it but the merge might be a bit cleaner :)

Does that look okay?

PR #1200, opened 2019-11-01
[xargs] Update docblock comment

ryanrhee:


pre_commit/xargs.py
```diff
@@ -106,6 +106,7 @@ def _thread_mapper(maxsize):
 def xargs(cmd, varargs, **kwargs):
     """A simplified implementation of xargs.
 
+    color: Make a TTY
```
asottile: technically it's a `pty`

asottile: (and only on platforms which support it)

ryanrhee: tru

ryanrhee:
```suggestion
    color: Make a pty
```

pre_commit/xargs.py
```diff
@@ -106,6 +106,7 @@ def _thread_mapper(maxsize):
 def xargs(cmd, varargs, **kwargs):
     """A simplified implementation of xargs.
 
+    color: Make a pty
```
ryanrhee:
```suggestion
    color: Make a pty on platforms that support it
```

pre_commit/xargs.py
ryanrhee: updated

Issue #1076, opened 2019-07-08
autoupdate: flip-flop between shallow-fetched tags


asottile:
```yaml
repos:
  - repo: https://github.com/asottile/pyupgrade
    rev: v1.20.1
    hooks:
      - id: pyupgrade
```

```console
$ pre-commit autoupdate
Updating https://github.com/asottile/pyupgrade...updating v1.20.1 -> v1.20.0.
$ pre-commit autoupdate
Updating https://github.com/asottile/pyupgrade...updating v1.20.0 -> v1.20.1.
$ pre-commit autoupdate
Updating https://github.com/asottile/pyupgrade...updating v1.20.1 -> v1.20.0.
$ pre-commit autoupdate
Updating https://github.com/asottile/pyupgrade...updating v1.20.0 -> v1.20.1.
```

Seems to be related to shallow clones as I can only reproduce there.

CC @mxr

asottile:
Here's my entry on the mailing list: https://marc.info/?l=git&m=156256221122204&w=2

PR #1205, opened 2019-11-04
Fix git reference for tags to get actual latest tag

StephenBrown2:
I've been running pre-commit with the isort hook, whose pypi package is at [4.3.21](https://pypi.org/project/isort/#history), but whose latest tag is [4.3.21-2](https://github.com/timothycrosley/isort/releases), a hotfix re-release of 4.3.21.

This causes issues with pre-commit always thinking the latest commit is the other tag, when the config is set to `4.3.21`, it wants to upgrade to tag `4.3.21-2`, and when it's set to `4.3.21-2`, it wants to "upgrade" to tag `4.3.21`. This adds a git check to list the latest tag, and bases the `git describe` command on that instead of `FETCH_HEAD`, which fixes the issue for me with isort reliably.

A reproducible `.pre-commit-config.yaml`:
```yaml
repos:
  - repo: https://github.com/timothycrosley/isort
    rev: 4.3.21-2
    hooks:
      - id: isort
        additional_dependencies: ["toml"]
```

### DIFF
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -41,7 +41,11 @@ def _update_repo(repo_config, store, tags_only):
         git.init_repo(repo_path, repo_config['repo'])
         cmd_output_b('git', 'fetch', 'origin', 'HEAD', '--tags', cwd=repo_path)
 
-        tag_cmd = ('git', 'describe', 'FETCH_HEAD', '--tags')
+        # Get the latest tagged commit
+        latest_tag_cmd = ('git', 'rev-list', '--tags', '--max-count=1')
+        latest_tagged = cmd_output(*latest_tag_cmd, cwd=repo_path)[1].strip()
+
+        tag_cmd = ('git', 'describe', latest_tagged, '--tags')
         if tags_only:
             tag_cmd += ('--abbrev=0',)
         else:


asottile:
the current logic is correct: the latest (by history) tagged revision in the mainline branch

StephenBrown2:
Can you please define "by history"? And also explain why getting the "latest" flip-flops between the two revisions each time?

asottile:
For `isort`, the mainline branch is `develop`.  The only tags which are considered are the forward history of the `develop` branch.

Here's the list of tags which are available on the `develop` branch of `isort`:

```console
$ git log --decorate --oneline  | grep -o '(tag:[^)]*)'
(tag: 4.3.21)
(tag: 4.3.14)
(tag: 4.3.9)
(tag: 4.3.8)
(tag: 4.3.7)
(tag: 4.3.6)
(tag: 4.3.5)
(tag: 4.3.4)
(tag: 4.3.3)
(tag: 4.3.2)
...
```

`4.3.21-2` is tagged on an unrelated history to the `develop` branch

by "latest" I mean as the history progresses forwards (the command in fact is "find the closest tag to the revision asked for" and the revision asked for is the tip of the default branch)

StephenBrown2:
Understood. I'll bump https://github.com/timothycrosley/isort/issues/966 to see if we can get a new version out of him.

For now, setting `rev: 4.3.21` seems to not flip-flop with the latest pre-commit, so I guess I'll leave it at that.

asottile:
curious what you mean by flip-flop?  there was a bug at one point which [was fixed](https://github.com/pre-commit/pre-commit/issues/1076) that maybe you're hitting?

StephenBrown2:
Yeah, I thought it might be that, but I don't recall what version of pre-commit I was using at the time, so I'm guessing that's the thing and it's fixed now.

Issue #1206, opened 2019-11-05
pre-commit needs a retry mecanism for network operations


ssbarnea:
It seems that pre-commit can fail in CI with what appears to be caused by network operations and it does not do any kind of retry attempt.
```
2019-11-05 09:56:13.953271 | rdo-centos-7 | [INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
2019-11-05 09:56:16.350589 | rdo-centos-7 | [INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks:flake8>=3.7.7.
2019-11-05 09:56:18.641544 | rdo-centos-7 | [INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
2019-11-05 09:56:20.778130 | rdo-centos-7 | An unexpected error has occurred: CalledProcessError: Command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
2019-11-05 09:56:20.778255 | rdo-centos-7 | Return code: 128
2019-11-05 09:56:20.778375 | rdo-centos-7 | Expected return code: 0
2019-11-05 09:56:20.778429 | rdo-centos-7 | Output: (none)
2019-11-05 09:56:20.778462 | rdo-centos-7 | Errors:
2019-11-05 09:56:20.778543 | rdo-centos-7 |     error: RPC failed; result=22, HTTP code = 4042019-11-05 09:56:20.778639 | rdo-centos-7 |     fatal: The remote end hung up unexpectedly
```

Full log at: http://logs.rdoproject.org/91/23591/2/check/tox-linters/f9bf1d6/job-output.txt

asottile:
please consider using softer language ("needs") when making reports

404 isn't a retryable error, this is due to old git on centos

the underlying tools (pip, git, etc.) already manage their own retry protocols -- adding more retries isn't going to help

ssbarnea:
I raised this bug as it was found by two of my colleagues, I knew that faced it myself few months back.

I do agree with you that 404 is not retriable and in 99/100 cases a client should respect it. Still I remember few cases where hosted services returned invalid HTTP responses and where the only option was to add some special cases (Atlassian and their CDN...).

Still the real bug behind my original request would be better described as:

- pre-commit fails to clone flake8 on centos-7 using [officially documented hook configuration](https://flake8.pycqa.org/en/latest/user/using-hooks.html).
[...]

PR #1215, opened 2019-11-08
Use official GitLab cloning locations

ssbarnea:
Avoid failure to clone repository with older versions of git by using
official cloning url (as exposed by GitLab itself in their web
interface).

References:
- https://stackoverflow.com/questions/32533379
- https://gitlab.com/gitlab-org/gitlab/issues/29629

Fixes: #1206

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
     -   id: name-tests-test
     -   id: requirements-txt-fixer
     -   id: double-quote-string-fixer
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://gitlab.com/pycqa/flake8.git
     rev: 3.7.7
     hooks:
     -   id: flake8


javierpena:
I can confirm this fixes issues when running on CentOS 7.

ssbarnea:
I also made a request to fix flake8 documentation, so we can avoid confusing other users https://gitlab.com/pycqa/flake8/merge_requests/381

asottile:
sorry, we don't use centos 7 so we don't need to work around this bug in git

webknjaz:
@asottile would you mind listing the runtime requirements somewhere in docs? And maybe having an assertion in the beginning of `pre-commit` execution pointing out if some prerequisites are missing could be a good idea.

asottile:
it's... complicated. there's a list in CONTRIBUTING.md, but there are feature which are version dependent and generally graceful segregation otherwise (for example pre-push was not always a thing, this gitlab+git bug only affects really old git, intent to add is buggy or absent in old git, etc.etc.)

ssbarnea:
Was someone who advocated the pre-commit tool in the past years to many teams, I am upset about this direction. I am curious to whom does the **we** from *we don't use centos 7* sentence refers to? Who is behind pre-commit tool? The documentation does not mention anything about project governance or what platforms are supported or not.

Does the future of pre-commit depend only on your decisions? Are your decisions set in stone? I am asking because I would be happy to offer to help with maintenance burden, especially around packaging and cross platform testing/deployment which are usually boring but also a primary source of frustration for the end-users.

To be back at the original PR: we all agree that it does avoid the mentioned error with old-git versions. The big question is **what is the downside of adding `.git` to that url?** I guess that adding 4 bytes to the size of the file does not count as a real downside, especially as this url is the one exposed by gitlab itself.

gitlab-ss

asottile:
> I am curious to whom does the we from we don't use centos 7 sentence refers to?

me, the core contributors, our CI

>  The documentation does not mention anything about project governance 

it's primarily me, but there are [other members](https://github.com/orgs/pre-commit/people) in the organization

> or what platforms are supported or not.

the [docs](https://pre-commit.com) do have `**Support:**` call outs. These cover "general" platforms such as "windows", "macos", "linux", "cygwin", etc. as enumerating every possible platform and combination is impossible (just with linux alone there are thousands of flavors shipping various combinations of software and patches)

> Does the future of pre-commit depend only on your decisions? Are your decisions set in stone?

largely yes and yes -- as it always has

> I would be happy to offer to help with maintenance burden

if you want to submit patches to fix actual bugs in the software (python files) please do so -- but this patch is not productive (pre-commit is not developed or tested on ancient distributions such as centos 7)

> what is the downside of adding .git to that url?

this patch is "noise", it doesn't actually fix anything.  additionally, `.git` is unnecessary cargo cult that I'd rather not promote

PR #1218, opened 2019-11-12
WIP: Add quiet mode to pre-commit run

allburov:
Add `--quiet` flag ot `pre-commit run` to show only failing hook's messages.

TODO:
- [X] Add `--quiet` flag to `pre-commit run`
- [ ]  Introduce env variable `PRE_COMMIT_QUIET=0|1|false|true` which allow users to enable quiet mode by default in their PC
- [ ] Add option `--output-mode=[normal|quiet]` to `pre-commit install` command 
- [ ] Add docs about flag and option to https://github.com/pre-commit/pre-commit.github.io

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -13,12 +13,13 @@ from pre_commit import git
 from pre_commit import output
 from pre_commit.clientlib import load_config
 from pre_commit.output import get_hook_message
+from pre_commit.output import NormalMode
+from pre_commit.output import QuietMode
 from pre_commit.repository import all_hooks
 from pre_commit.repository import install_hook_envs
 from pre_commit.staged_files_only import staged_files_only
 from pre_commit.util import cmd_output_b
 from pre_commit.util import noop_context
-
 
 logger = logging.getLogger('pre_commit')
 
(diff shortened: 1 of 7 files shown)


allburov:
@asottile what about this to fix https://github.com/pre-commit/pre-commit/issues/823? Could you review this solution, please.

allburov:
@asottile 
I understand your opinion, very few repositories have more than 10 hooks (I have looked to open source repositories). But a big project (without shared code base) may have a lot of hooks. However it's really less than 1% of repositories who use pre-commit. 

I tried write this with `if`-statements, but it looked hard. 
I tried rewrite with "pass args.quiet" to functions (like args.color is passed) - it did not seem pythonic and easy to read or modify.
I looked at `logging` or [ansible.display](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/display.py#L36) solutions also, but it also will have `if` statement to print the first status message with number of hooks.

So my decision is just inject new `output` to run hook function. If you will want to write new string to console in the function you just will use `write_line` or `write` and don't wary about quiet mode. 

Lets going to simplify the solution (if we will can without increase complexity)!

blueyed:
@allburov 
Maybe wrapping the whole output / run is more feasible in the long run?  77aab5e

peterjc:
I tried the other extreme in #1560, no output wrapping at all - just suppressing "skipped" message output in the first place. This seemed like the low hanging fruit, and made me wonder about having that off by default - thus #1561 which only shows the "skipped" messages in verbose mode.

PR #1228, opened 2019-11-25
`bash` on Windows

kvalkhof75:
POC or MVP: Succeeds checking `shfmt` on Windows which uses `bash`. Requires that `bash` is meaningful (WSL, MSys64) on the command prompt, and `shfmt` is installed in the `bash` environment.

pre-commit run shfmt --file bash_script.sh

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,3 +40,11 @@ repos:
     hooks:
     -   id: check-hooks-apply
     -   id: check-useless-excludes
+    # Requires to be run under `bash`.
+    # Requires installation of additional tools. Tools are available on Windows.
+    # shfmt: https://github.com/mvdan/sh/releases
+-   repo: https://github.com/syntaqx/git-hooks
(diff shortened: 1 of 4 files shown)


asottile:
pre-commit already works fine with hooks written in bash -- please make a bug report if that's not the case

kvalkhof75:
Mmm, seems that I am live now. That was unexpected. I have never done a pull request before and I thought it was just on my fork...

I'll post an issue explaining what I want to achieve later and do not make a pull request until it is finished.

kvalkhof75:
Hello Anthony,
Sorry for the pull request, I apparently set it on your repo instead of
just on my fork.
While the bash hook itself works, the tools I want it to work with (`shfmt`
and `clang-format`) do not, because the paths given to these tools are not
converted to POSIX.
Or do I misunderstand how the `bash` hook should work?
I made issue.#1229
Thanks for a great tool!
Kees Valkhof

PR #1232, opened 2019-12-02
Support for conda as a language

### LINKED ISSUE #1204: Support for conda as a language
It would be nice to have support for conda in `pre-commit` as a "language"/environment manager. This would enable setups where packages are installed through conda. The typical usage for this will be for Python and R setups (e.g. https://github.com/lorenzwalthert/precommit/issues/99) where conda is most popular but as `conda` is language independent.

The implementation would work as follows:

- A hook is expected to have a requirements.txt that is installed using `conda create -p <prefix> --file=requirements.txt`. I would not use the conda `environment.yml` as this already dictates the name of the environment.
- The hook is run using `conda run -p <prefix> <entry>`
- We will use the `conda` executable that is available on the `PATH`.

Note: I would volunteer to implement this once there are some 👍 that the proposal is sound.

xhochy:
Fixes #1204

## Example usage

`.pre-commit-hooks.yaml`:

```
- id: black-conda
  name: black-conda
  description: "Black: The uncompromising Python code formatter"
  entry: black
  language: conda
  require_serial: true
  types: [python]
```

`environment.yml`:

```
channels:
  - conda-forge
  - defaults
dependencies:
  - black=19.10b0
```

### DIFF
--- a/pre_commit/languages/conda.py
+++ b/pre_commit/languages/conda.py
@@ -1,0 +1,66 @@
+import contextlib
+import os
+
+from pre_commit.envcontext import envcontext
+from pre_commit.envcontext import UNSET
+from pre_commit.envcontext import Var
+from pre_commit.languages import helpers
+from pre_commit.util import clean_path_on_failure
+from pre_commit.util import cmd_output_b
+
+ENVIRONMENT_DIR = 'conda'
+get_default_version = helpers.basic_get_default_version
+healthy = helpers.basic_healthy
+
+
+def get_env_patch(env):
+    # On non-windows systems executable live in $CONDA_PREFIX/bin, on Windows
+    # they can be in $CONDA_PREFIX/bin, $CONDA_PREFIX/Library/bin,
+    # $CONDA_PREFIX/Scripts and $CONDA_PREFIX. Whereas the latter only
+    # seems to be used for python.exe.
+    path = (os.path.join(env, 'bin'), os.pathsep, Var('PATH'))
+    if os.name == 'nt':  # pragma: no cover (platform specific)
+        path = (env, os.pathsep) + path
+        path = (os.path.join(env, 'Scripts'), os.pathsep) + path
+        path = (os.path.join(env, 'Library', 'bin'), os.pathsep) + path
+
+    return (
+        ('PYTHONHOME', UNSET),
+        ('VIRTUAL_ENV', UNSET),
+        ('CONDA_PREFIX', env),
+        ('PATH', path),
+    )
+
+
+@contextlib.contextmanager
+def in_env(prefix, language_version):
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, language_version)
+    envdir = prefix.path(directory)
+    with envcontext(get_env_patch(envdir)):
+        yield
+
+
+def install_environment(prefix, version, additional_dependencies):
+    helpers.assert_version_default('conda', version)
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, version)
+
+    env_dir = prefix.path(directory)
+    with clean_path_on_failure(env_dir):
+        cmd_output_b(
+            'conda', 'env', 'create', '-p', env_dir, '--file',
+            'environment.yml', cwd=prefix.prefix_dir,
+        )
+        if additional_dependencies:
+            cmd_output_b(
+                'conda', 'install', '-p', env_dir, *additional_dependencies,
+                cwd=prefix.prefix_dir
+            )
+
+
+def run_hook(hook, file_args, color):
+    # TODO: Some rare commands need to be run using `conda run` but mostly we
+    #       can run them withot which is much quicker and produces a better
+    #       output.
+    # cmd = ('conda', 'run', '-p', env_dir) + hook.cmd
+    with in_env(hook.prefix, hook.language_version):
+        return helpers.run_xargs(hook, hook.cmd, file_args, color=color)
--- a/testing/resources/conda_hooks_repo/.pre-commit-hooks.yaml
+++ b/testing/resources/conda_hooks_repo/.pre-commit-hooks.yaml
@@ -1,0 +1,10 @@
+-   id: sys-exec
+    name: sys-exec
+    entry: python -c 'import os; import sys; print(sys.executable.split(os.path.sep)[-2]) if os.name == "nt" else print(sys.executable.split(os.path.sep)[-3])'
+    language: conda
+    files: \.py$
+-   id: additional-deps
+    name: additional-deps
+    entry: python
+    language: conda
+    files: \.py$
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -79,6 +79,46 @@ def _test_hook_repo(
     assert _norm_out(out) == expected
 
 
+def test_conda_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'conda_hooks_repo',
+        'sys-exec', [os.devnull],
+        b'conda-default\n',
+    )
+
+
+def test_conda_with_additional_dependencies_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'conda_hooks_repo',
+        'additional-deps', [os.devnull],
+        b'OK\n',
+        config_kwargs={
+            'hooks': [{
+                'id': 'additional-deps',
+                'args': ['-c', 'import mccabe; print("OK")'],
+                'additional_dependencies': ['mccabe'],
+            }],
+        },
+    )
+
+
+def test_local_conda_additional_dependencies(store):
+    config = {
+        'repo': 'local',
+        'hooks': [{
+            'id': 'local-conda',
+            'name': 'local-conda',
+            'entry': 'python',
+            'language': 'conda',
+            'args': ['-c', 'import mccabe; print("OK")'],
+            'additional_dependencies': ['mccabe'],
+        }],
+    }
+    ret, out = _get_hook(config, store, 'local-conda').run((), color=False)
+    assert ret == 0
+    assert _norm_out(out) == b'OK\n'
+
+
 def test_python_hook(tempdir_factory, store):
     _test_hook_repo(
         tempdir_factory, store, 'python_hooks_repo',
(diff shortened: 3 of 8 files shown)


xhochy:
This is lacking sufficient tests, I'm unsure what should be tested and am open for suggestions.

pre_commit/languages/conda.py
```diff
@@ -0,0 +1,31 @@
...
+from pre_commit.util import clean_path_on_failure
+from pre_commit.util import cmd_output_b
+
+ENVIRONMENT_DIR = 'conda'
+get_default_version = helpers.basic_get_default_version
+healthy = helpers.basic_healthy
+
+
+def install_environment(prefix, version, additional_dependencies):
+    additional_dependencies = list(additional_dependencies)
```
asottile: why convert to a list?

pre_commit/languages/conda.py
```diff
@@ -0,0 +1,31 @@
...
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, version)
```
asottile: `language_version` isn't currently being handled -- there's a helper to show that's not supported: https://github.com/pre-commit/pre-commit/blob/023b337ff0f5566f366dd85c390804208a60f7e7/pre_commit/languages/helpers.py#L27-L31

pre_commit/languages/conda.py
```diff
@@ -0,0 +1,31 @@
...
+    env_dir = prefix.path(directory)
+    with clean_path_on_failure(env_dir):
+        cmd_output_b(
+            'conda', 'env', 'create', '-p', env_dir, '--file',
+            'environment.yml', cwd=prefix.prefix_dir,
+        )
+        if additional_dependencies:
+            cmd_output_b(
+                'conda', 'install', '-p', env_dir, *additional_dependencies,
+                cwd=prefix.prefix_dir
```
asottile: the irony when committing to pre-commit and not running the pre-commit hooks :laughing:

tests/repository_test.py
```diff
@@ -79,6 +79,14 @@ def _test_hook_repo(
     assert _norm_out(out) == expected
 
 
+def test_conda_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'conda_hooks_repo',
```
asottile: did you forget to check this one in?

xhochy: Yes, fixed. We probably need to install conda in the CI.

testing/resources/conda_hooks_repo/.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,5 @@
+-   id: foo
+    name: Foo
+    entry: echo Test
```
asottile: this doesn't really show that conda works -- it doesn't exercise inside the conda environment

pre_commit/languages/conda.py
xhochy: 😭 🙈

pre_commit/languages/conda.py
```diff
@@ -26,6 +25,9 @@ def install_environment(prefix, version, additional_dependencies):
 
 def run_hook(hook, file_args, color):
     directory = helpers.environment_dir(ENVIRONMENT_DIR, hook.language_version)
-    env_dir = hook.prefix.path(directory)
-    cmd = ('conda', 'run', '-p', env_dir) + hook.cmd
+    cmd = (hook.prefix.path(directory, 'bin', hook.cmd[0]),) + hook.cmd[1:]
```
asottile: using the env patch and prefixing `PATH` is a more reliable way to accomplish this -- though if `conda run` works then maybe we just use that?

tests/repository_test.py
```diff
@@ -82,8 +82,23 @@ def _test_hook_repo(
...
+def test_conda_with_additional_dependencies_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'conda_hooks_repo',
+        'additional-deps', [os.devnull],
+        b'OK\n',
+        config_kwargs={
+            'hooks': [{
+                'id': 'additional-deps',
+                'args': ['-c', 'import requests; print("OK")'],
+                'additional_dependencies': ['requests'],
```
asottile: let's use something that isn't requests, mccabe is nice and small and no external deps for example

pre_commit/languages/conda.py
xhochy: `conda run` produces some annoying output in the failure case and depending on the environment introduces a significant runtime overhead. I would make that switchable on a per-hook basis. What would be the best approach for that? Have two languages  `conda` and `conda_run`?

With env patch and prefixing, you mean `with envcontext(get_env_patch(envdir)):`?

tests/repository_test.py
xhochy: Done.

pre_commit/languages/conda.py
asottile: maybe for now we just make `conda` and don't enable `conda run` until someone asks for it (?) -- and yes to `envcontext` :+1:

xhochy:
This has now all review comments incorporated and tests are passing except for Windows. There we need to download a lightweight conda (take an exe from https://repo.anaconda.com/pkgs/misc/conda-execs/) to get the tests passing. This needs an modification of the Azure template like https://github.com/asottile/azure-pipeline-templates/blob/f8721c757e218487ca0bb38af0201b2a04e9ed52/job--python-tox.yml#L63-L68 @asottile Should I modify the template to always install a conda.exe on Windows or make this conditional on some parameter?

asottile:
there's a `pre_test` list where you can install whatever you need 👍 

for example, here's what it does to install ruby / swift on linux: https://github.com/pre-commit/pre-commit/blob/023b337ff0f5566f366dd85c390804208a60f7e7/azure-pipelines.yml#L41-L46

pre_commit/languages/conda.py
```diff
@@ -0,0 +1,64 @@
...
+
+def get_env_patch(env):
+    # On non-windows systems executable live in $CONDA_PREFIX/bin, on Windows
+    # they can be in $CONDA_PREFIX/bin, $CONDA_PREFIX/Library/bin, and
+    # $CONDA_PREFIX/Scripts.
+    path = (os.path.join(env, 'bin'), os.pathsep, Var('PATH'))
+    if os.name == 'nt':  # pragma: no cover (platform specific)
+        path = (env, os.pathsep) + path
+        path = (os.path.join(env, 'Scripts'), os.pathsep) + path
+        path = (os.path.join(env, 'Library', 'bin'), os.pathsep) + path
```
asottile: this doesn't match the comment, looks like you're adding `env;env/Scripts;env\Library\bin;env\bin` to the path

also this can be
path += (
    os.path.join(env, 'Scripts'), os.pathsep,
    os.path.join(env, 'Library', 'bin'), os.pathsep,
)
```
or some such

the order probably matters here as well?

File: pre_commit/languages/conda.py [hunk 9]
```diff
@@ -0,0 +1,64 @@
...
+
+    return (
+        ('PYTHONHOME', UNSET),
+        ('VIRTUAL_ENV', UNSET),
```
asottile: does conda's activate usually clear these?

pre_commit/languages/conda.py
xhochy: The order matters, thus I cannot do `path +=`. Also I have adjusted the comment as the code is correct. `python.exe` is placed directly in `env`, this took a while to debug to find this problem.

pre_commit/languages/conda.py
xhochy: Yes, conda's activation does clear them but we currently bypass the activation by only settting the 
 `PATH`.

pre_commit/languages/conda.py
asottile: I think this is the only outstanding comment here

xhochy:
@asottile Adjusted all things and CI is still green. :)

pre_commit/languages/conda.py
xhochy: Ops, didn't see this. Now also fixed.

pre_commit/languages/conda.py
```diff
@@ -0,0 +1,66 @@
...
+        ('CONDA_PREFIX', env),
+        ('PATH', path),
+    )
+
+
+@contextlib.contextmanager
+def in_env(prefix, language_version):
+    helpers.assert_version_default('conda', language_version)
```
asottile: this should be part of `install_environment` I believe

xhochy:
@asottile Azure is failing with missing files in the template repo but these actually exist. Do you have an idea what is going wrong here?

pre_commit/languages/conda.py
xhochy: Should these be removed here then?

asottile: yes -- you can take precedent from the other languages that use this function

asottile:
> @asottile Azure is failing with missing files in the template repo but these actually exist. Do you have an idea what is going wrong here?

huh they must have changed the inclusion rules :/

asottile:
I'm trying to fix that in #1234 -- we'll see!

pre_commit/languages/conda.py
```diff
@@ -0,0 +1,67 @@
...
+    helpers.assert_version_default('conda', version)
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, version)
+
+    env_dir = prefix.path(directory)
+    with clean_path_on_failure(env_dir):
+        cmd_output_b(
+            'conda', 'env', 'create', '-p', env_dir, '--file',
+            'environment.yml', cwd=prefix.prefix_dir,
+        )
+        if additional_dependencies:
```
asottile: I just remembered, since this supports `additional_dependencies` we should make it work for `repo: local` -- should be as easy as adding a test and a trivial `environment.yml` to the empty repository template: https://github.com/pre-commit/pre-commit/tree/master/pre_commit/resources

asottile:
ok if you rebase on master it should fix the CI

pre_commit/languages/conda.py
xhochy: Implemented this and also discovered a `conda` bug: https://github.com/conda/conda/issues/9487 

I hope the workaround is ok.

Issue #1235, opened 2019-12-06
Running tests on a system-wide install.


Aniket-Pradhan:
Hello there!

Thanks for providing such an amazing tool. It is really helpful and simply eases out the development pipeline.

I/We were packaging pre-commit for Fedora (a Linux distribution) and were stuck in running the tests for the installation. I cannot find steps to run the tests for system-wide installation, but only for a virtual-environment. Currently, I am running the tests as:

```bash
$ # In the project directory, after installing the tool (python setup.py install)
$ git init # since the tests require a git repo for running (I am using the release tar)
$ export PYTHONPATH=/usr/lib/python3.7/site-packages
$ pytest-3
```

The installation worked out completely fine, but some of the tests fail, as described below. It would be really helpful if you could explain how can we amend running the tests.

Test failures:

```
tests/parse_shebang_test.py ..............FFFF
tests/languages/python_test.py ...........F
```

```
    def test_normalize_cmd_PATH():
        cmd = ('python', '--version')
        expected = (distutils.spawn.find_executable('python'), '--version')
>       assert parse_shebang.normalize_cmd(cmd) == expected
tests/parse_shebang_test.py:130: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pre_commit/parse_shebang.py:72: in normalize_cmd
    exe = normexe(cmd[0])
pre_commit/parse_shebang.py:52: in normexe
    _error('not found')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
msg = 'not found'
    def _error(msg):
>       raise ExecutableNotFoundError('Executable `{}` {}'.format(orig, msg))
E       pre_commit.parse_shebang.ExecutableNotFoundError: Executable `python` not found
pre_commit/parse_shebang.py:47: ExecutableNotFoundError
__________________________ test_normalize_cmd_shebang __________________________
in_tmpdir = '/tmp/pytest-of-mockbuild/pytest-0/test_normalize_cmd_shebang0/0'
    def test_normalize_cmd_shebang(in_tmpdir):
>       python = distutils.spawn.find_executable('python').replace(os.sep, '/')
E       AttributeError: 'NoneType' object has no attribute 'replace'
[...]

PR #1236, opened 2019-12-06
Use echo instead of python in parse_shebang_test

asottile:
See #1235

This makes the test succeed better when not in a virtualenv (where both `python` and `python3` executables will exist)


asottile:
oh right that's not going to work well on windows

Issue #1273, opened 2020-01-08
Calling pre-commit --version as a module confuses it about its name


ssbarnea:
I found a minor bug in `pre-commit` which reports `__main__.py` instead of `pre-commit` when is asked about its version using module calling.

    python -m pre_commit --version
    __main__.py 1.20.0
                                                                                                                                                                                     
I would be nice to find a solution for it as I observed this issue on other projects too. Still, it seems that tools like pip, virtualenv, tox or pytest are not affected by this bug.

Calling tools as python modules is quite common on CI as it avoids two common issues: exec install path not in in PATH, wrong python interpreter would be used.

asottile:
argparse by default infers the `prog` attribute from `sys.argv[0]` -- you can however pass hardcoded string to `ArgumentParser(prog=...)` to configure this

would you like to submit a patch?

orcutt989:
Bombs on Python 3.7
```shell
macbook71:pre-commit user$ python -m pre_commit --version
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/user/Projects/pre-commit/pre_commit/__main__.py", line 3, in <module>
    from pre_commit.main import main
  File "/Users/user/Projects/pre-commit/pre_commit/main.py", line 8, in <module>
    import pre_commit.constants as C
  File "/Users/user/Projects/pre-commit/pre_commit/constants.py", line 26, in <module>
    VERSION = importlib_metadata.version('pre_commit')
  File "/usr/local/lib/python3.7/site-packages/importlib_metadata/__init__.py", line 509, in version
    return distribution(distribution_name).version
  File "/usr/local/lib/python3.7/site-packages/importlib_metadata/__init__.py", line 482, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/local/lib/python3.7/site-packages/importlib_metadata/__init__.py", line 187, in from_name
    raise PackageNotFoundError(name)
importlib_metadata.PackageNotFoundError: pre_commit
```

asottile:
works fine for me, how did you install pre-commit?

[...]

PR #1276, opened 2020-01-10
fix prog arg to return correct version

orcutt989:
See issue #1273 

Fixes output when running `--version`.

Before
```bash
python -m pre_commit --version
__main__.py 1.20.0
```

After
```bash
python -m pre_commit --version
pre-commit 1.21.0
```

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -148,7 +148,7 @@ def _adjust_args_and_chdir(args):
 def main(argv=None):
     argv = argv if argv is not None else sys.argv[1:]
     argv = [five.to_text(arg) for arg in argv]
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(prog='pre_commit')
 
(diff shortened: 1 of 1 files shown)


pre_commit/main.py
```diff
@@ -148,7 +148,7 @@ def _adjust_args_and_chdir(args):
 def main(argv=None):
     argv = argv if argv is not None else sys.argv[1:]
     argv = [five.to_text(arg) for arg in argv]
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(prog='pre_commit')
```
asottile: `pre-commit` is stylized this way
```suggestion
    parser = argparse.ArgumentParser(prog='pre-commit')
```

orcutt989: Woops! Not sure where I got that from. Fixed c7d938c 👍

asottile:
by the way, if you put "Closes #...." in your PR description, it will be auto closed when the PR is merged

Issue #1279, opened 2020-01-12
support for `flake8 --diff`, which checks only the diff in the commit


koreno:
when flake8 is passed the `--diff` flag, it expects stdin to provide it with the code to lint, instead of as filenames.
It would be nice to add to pre-commit support for such hooks, in the form of a `pass_diff=True` hook parameter.

asottile:
pre-commit has no such mode for that, pre-commit operates on *files*.  doing this directly in the framework doesn't really make sense (what would `pre-commit run --all-files` do?)

however you could implement a small wrapper script which invokes flake8 in that way:

```python
# untested, probably needs tweaking (written in github comment box)
import subprocess
import sys

def main() -> int:
    diff = subprocess.run(
        ('git', 'diff', '--staged', '--', *sys.argv[1:]),
        stdout=subprocess.PIPE,
    ).stdout
    return subprocess.call(('flake8', '--diff'), input=diff)

if __name__ == '__main__':
    exit(main())
```

(additionally, `--diff` mode is meant to be a temporary measure in flake8, it's best to get linting passing and not use that option)

if you're looking for ways to gradually add linting to a repository (which is what `--diff` was designed to do) you'll probably find `files` / `exclude` more helpful (get a part of the codebase passing, add it to `files` or the opposite with `exclude`)

asottile:
also note that `flake8 --diff` isn't faster flake8 still lints everything and then after that trims down the output.  additionally, `flake8 --diff` can still allow lint errors to creep into your codebase as touching lines can cause lint errors on other lines so it's not exactly the safest gate keeper. (I'm also a maintainer of flake8)

koreno:
> pre-commit has no such mode for that, pre-commit operates on _files_. doing this directly in the framework doesn't really make sense (what would `pre-commit run --all-files` do?)
"--all-files" could ignore hooks that ask for diff, or if we really wanted, implement a more sophisticated spec that tells per-commit what to do with the hook in each mode. Might be right for "completeness", but I don't feel that's important.

> wrapper script

That might be the way to go...



> gradually add linting to a repository

I'm aware of that, but that's not enough... When people are stressed about solving bugs and adding features, and aren't big fans of style-guided, having to go over someone else's non compliant code is just too much. In addition, it make for hellish code-reviews, where the reviewers don't know which changes are there to pacify the linter, and which are the actual logic changes. git-blame also suffers.
[...]

PR #1280, opened 2020-01-12
support 'diff-only' hooks by passing diff via stdin (#1279)

koreno:
this is a poc/draft, with the intention to make minimal changes and see if this can work

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -139,10 +139,14 @@ def _run_single_hook(
 
         diff_cmd = ('git', 'diff', '--no-ext-diff')
         diff_before = cmd_output_b(*diff_cmd, retcode=None)
-        if not hook.pass_filenames:
-            filenames = ()
+        if hook.pass_diff:
+            inputs = (diff_before[1],)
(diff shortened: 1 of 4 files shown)


pre_commit/commands/run.py
```diff
@@ -117,9 +117,14 @@ def _run_single_hook(classifier, hook, skips, cols, verbose, use_color):
 
         diff_cmd = ('git', 'diff', '--no-ext-diff')
         diff_before = cmd_output_b(*diff_cmd, retcode=None)
-        filenames = tuple(filenames) if hook.pass_filenames else ()
+        if hook.pass_diff:
+            inputs = (diff_before[1],)
```
asottile: this isn't the correct value fwiw, `diff_before` is the *unstaged* diff (used to determine whether things change or not)

asottile:
if it wasn't clear, I don't want this change -- sorry

koreno:
I wrote the code before I got your response.
Thanks anyway, good luck.

asottile:
via #1279

PR #1288, opened 2020-01-16
Push remote env var details

### LINKED ISSUE #1274: Pre-push hook checks only when pushing to certain remote
I have two remotes configured in my repository (`origin` and `upstream`) and I want my pre-push hook to fire and do the configured checks only when we push to `upstream`, bypassing all checks when we push to remote `origin`. Is this possible without using git `--no-verify` option?
I was checking for ways to obtain the remote from `pre-commit`as a parameter and pass it to my local hook to bypass the checks in case remote doesn't match `upstream` but I haven't been very successful. Any ideas on how to do this?

Thank you so much for all the great work you are doing with this project.

dmbarreiro:
Work in progress. One test added (test_push_remote_environment) which is failing right now. Closes #1274


dmbarreiro:
Comes from issue #1274

pre_commit/commands/run.py
```diff
@@ -259,6 +259,11 @@ def run(config_file, store, args, environ=os.environ):
         environ['PRE_COMMIT_ORIGIN'] = args.origin
         environ['PRE_COMMIT_SOURCE'] = args.source
 
+    # Expose remote name and url as environment variables for hooks to consume
+    if args.push_remote_name and args.push_remote_url:
+        environ['PRE_PUSH_REMOTE_NAME'] = args.push_remote_name
+        environ['PRE_PUSH_REMOTE_URL'] = args.push_remote_url
```
asottile: these should be `PRE_COMMIT_...` since they come from the the tool named `pre-commit` (as per above)

pre_commit/resources/hook-tmpl
```diff
@@ -153,7 +154,8 @@ def _pre_push(stdin):
                     opts = ('--origin', local_sha, '--source', source)
 
     if opts:
-        return opts
+        remote_opts = ("--push-remote-name", remote_name, "--push-remote-url", remote_url)
```
asottile: huh weird why isn't CI running on this

tests/commands/run_test.py
```diff
@@ -633,6 +633,30 @@ def _run_for_stage(stage):
...
 
+def test_push_remote_environment(cap_out, store, repo_with_passing_hook):
+    config = {
+        'repo': 'local',
+        'hooks': [
+            {
+                'id': 'print-push-remote',
+                'name': 'Print push remote name',
+                'entry': 'python -c \'import os; print(os.environ["PRE_PUSH_REMOTE_NAME"])\'',
+                'language': 'system',
```
asottile: a (sadly) more portable alternative is to use bash here:

```
entry: bash -c 'echo "$PRE_PUSH_REMOTE_NAME"'
```

the main reason being the tests run in some environments where `python` doesn't exist (only `python3`) -- and then if you use `python3` then it doesn't work on windows which only has `python` :man_facepalming:

pre_commit/resources/hook-tmpl
```diff
@@ -136,7 +137,7 @@ def _pre_push(stdin):
             # ancestors not found in remote
             ancestors = subprocess.check_output((
                 'git', 'rev-list', local_sha, '--topo-order', '--reverse',
-                '--not', '--remotes={}'.format(remote),
+                '--not', '--remotes={}'.format(remote_name),
```
asottile: you'll want to base on the latest master -- this line has been changed recently (so you'll probably conflict)

asottile:
> Comes from issue #1274

note that if you put "Closes #1274" in your PR description this'll get automatically linked *and* github will auto-close the issue upon merging!

pre_commit/commands/run.py
dmbarreiro: You mean PRE_PUSH_REMOTE_NAME should be PRE_COMMIT_REMOTE_NAME and the same with remote url?

pre_commit/resources/hook-tmpl
dmbarreiro: I'm not sure I got this comment.

pre_commit/commands/run.py
asottile: yes

pre_commit/resources/hook-tmpl
asottile: ah a bit of two things:

- it's slightly ironic that you didn't run `pre-commit` while contributing to `pre-commit`
- I expected azure pipelines to have triggered a build against this branch and shown github status but it didn't

asottile:
I ran the hooks and moved the test (the `run` tests do some isolated `environ` management so it doesn't get forwarded to the hook executables as a normal run would)

asottile:
I renamed the options -- might be useful for some other feature later

pre_commit/resources/hook-tmpl
```diff
@@ -150,7 +151,9 @@ def _pre_push(stdin: bytes) -> Tuple[str, ...]:
                     opts = ('--origin', local_sha, '--source', source)
 
     if opts:
-        return opts
+        return (
+            *opts, '--remote-name', remote_name, '--remote-url', remote_url,
+        )
```
dmbarreiro: Why the name change?

dmbarreiro:
Sorry for that last commit and the print (I thought I could solve it just removing the byte string). I still do not understand why the pipeline was failing at some point when I pushed a new commit. Thank you so much for the support.

asottile:
no problem o7 thanks again for the help on this one!

asottile:
this has been released as part of [v2.0.0](https://github.com/pre-commit/pre-commit/releases/tag/v2.0.0) -- thanks again for the issue and patch!

dmbarreiro:
Great! Thank you for the support :-)

PR #1292, opened 2020-01-20
Move most of the actual hook script into `pre-commit hook-impl`

asottile:
this patch is mostly about portability:
- `#!/usr/bin/env {python}` would often end up with `python` on macos even if there's a python3 available
- the syntax of `.git/hooks/{hook_script}` contained py3+ syntax
- on windows, python's `exec` doesn't quite work all that well (? needs more investigation)


asottile:
hummmm bash is broken in azure pipelines -- maybe this is not the best idea 🤔

asottile:
this seems to be affected by https://github.com/microsoft/azure-pipelines-image-generation/issues/1468

asottile:
found why `os.exec*` is utterly useless on windows as well: https://bugs.python.org/issue19124

PR #1293, opened 2020-01-21
exclude GIT_SSL_NO_VERIFY env variable from getting stripped

schiermike:
Can this change please get merged? We sit behind a proxy that messes with certificates...

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -35,7 +35,10 @@ def no_git_env(_env: Optional[EnvironT] = None) -> Dict[str, str]:
     return {
         k: v for k, v in _env.items()
         if not k.startswith('GIT_') or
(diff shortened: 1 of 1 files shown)


asottile:
that's actually kinda terrifying -- doesn't this variable expose you to all sorts of malicious MITM?

I'm looking into why windows CI is failing -- I suspect a bad azure pipelines release :(

asottile:
I've opened https://github.com/microsoft/azure-pipelines-image-generation/issues/1468 for the `git` issue

schiermike:
Yeah, a bit yes. The proxy itself is pretty restricting and only allows connections to a handful of public domains such as github.com - and it does verify certs. So it's only a matter of trusting the path to the proxy inside the company network ...

asottile:
this has been released as part of [v2.0.0](https://github.com/pre-commit/pre-commit/releases/tag/v2.0.0) -- thanks again for the patch!

PR #1302, opened 2020-01-30
Fix pre-commit in python 3.6.0-3.6.1

asottile:


asottile:
fortunately I can use [flake8-typing-imports](https://github.com/asottile/flake8-typing-imports) to check this for me

PR #1303, opened 2020-01-30
Perl hook support

scop:
Here's a first cut for supporting Perl hooks. Appears to work for me with this .pre-commit-hooks.yaml locally added to a https://github.com/perltidy/perltidy clone + the usual config to use it elsewhere:
```yaml
- id: perltidy
  name: perltidy
  entry: perltidy --nostandard-output --backup-and-modify-in-place
  language: perl
  types: [perl]
```
No tests yet, haven't taken a look what it would take to add some. Throwing this here for comments already anyway.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,6 +24,11 @@ jobs:
     pre_test:
     - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
       displayName: Add conda to PATH
+    - powershell: |
+        Write-Host "##vso[task.prependpath]C:\Strawberry\perl\bin"
+        Write-Host "##vso[task.prependpath]C:\Strawberry\perl\site\bin"
+        Write-Host "##vso[task.prependpath]C:\Strawberry\c\bin"
+      displayName: Add strawberry perl to PATH
 - template: job--python-tox.yml@asottile
   parameters:
     toxenvs: [py37]
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -11,6 +11,7 @@ from pre_commit.languages import docker_image
 from pre_commit.languages import fail
 from pre_commit.languages import golang
 from pre_commit.languages import node
+from pre_commit.languages import perl
 from pre_commit.languages import pygrep
 from pre_commit.languages import python
 from pre_commit.languages import python_venv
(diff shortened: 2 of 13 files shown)


asottile:
neat! looks like a good start -- with some tests should probably be good enough as a v1 implementation!

scop:
Any hints for getting up to speed with writing the tests as easily as possible? :)

asottile:
Sure!

For this I think there's ~two main tests you'll want to add:

1. an integration test which tests a `language: perl` repository -- you can add a repository in `testing/resources/...` (there's quite a few examples there) -- and then you'd hook that up with a test in `tests/repository_test.py` (here's an example [for rust](https://github.com/pre-commit/pre-commit/blob/f0ee93c5a7ee84a895918a0c0d1bc269233ccb7f/tests/repository_test.py#L322-L326))
2. an integration test demonstrating `additional_dependencies` -- I'm less familiar with how you'd do this with perl, one easy way might be to define a `local`-like configuration and ensure that the additional dependency gets installed -- maybe something like [this golang test](https://github.com/pre-commit/pre-commit/blob/f0ee93c5a7ee84a895918a0c0d1bc269233ccb7f/tests/repository_test.py#L558-L572)

depending on what's out-of-the-box on azurepipelines we'll probably also need to install perl (?)

asottile:
if you get stuck I can probably help -- but you'll need to teach me a bit about perl package management :)

scop:
Thanks, I'll have a look. Additional deps already worked when I tested them -- there are multiple ways one can specify them for `cpan` to install, but the most usual one is just a module name.

I have no clue about perl in azure, and I'm more familiar with travis anyway, but I don't think it will be a problem at all, just something one needs to add.

One future improvement would be to support repos with `cpanfile`s in them -- for that one would first make sure either `carton` (especially if `cpanfile.snapshot` also exists) or `cpanm` is installed or install them if not. But repos with are less common, let's first get the baseline done.

scop:
Ok, got those implemented. Linux ok locally and in Azure, but tox_windows py37 fails writing a state file there or something, or maybe it's a result/side effect of something else. I guess it's a fair assumption that perl isn't installed in Windows there, but dunno if it could cause something like that.

I'm afraid I'll need someone to look into this.

asottile:
I believe I've fixed windows :crossed_fingers: -- played around this with a VM -- it also looks like azure pipelines preinstalls "strawberry perl" so perl should already be available

asottile:
:tada: thanks for the patch!

asottile:
this has been released as part of [v2.1.0](https://github.com/pre-commit/pre-commit/releases/tag/v2.1.0) -- thanks again!

PR #1322, opened 2020-02-09
Include column number in pygrep output

scop:


pre_commit/languages/pygrep.py
```diff
@@ -17,13 +17,28 @@
 install_environment = helpers.no_install
 
 
+def _column_marker(line: bytes, byte_offset: Optional[int]) -> str:
+    try:
+        char_offset = len(line[0:byte_offset].decode())
```
asottile: this was exactly why it's not implemented -- slicing bytes is nonsense and we intentionally don't consider the encoding of the file

also a python tip, you can omit the first segement of a slice and it will be `0` automatically (so `line[:byte_offset]`)

scop: Incoming regexes are converted to bytes using `.encode()`, which does to some degree imply UTF-8 treatment.

asottile: right, but the underlying files are arbitrarily encoded

scop: But applying a UTF-8 bytes pattern to a file encoded in some other encoding produces undefined results. It's more or less lucky if they match. Try for example this change to the test suite:
```diff
diff --git a/tests/languages/pygrep_test.py b/tests/languages/pygrep_test.py
index 06c24b9..71c6f2e 100644
--- a/tests/languages/pygrep_test.py
+++ b/tests/languages/pygrep_test.py
@@ -5,9 +5,9 @@ from pre_commit.languages import pygrep
 
 @pytest.fixture
 def some_files(tmpdir):
-    tmpdir.join('f1').write_binary(b'foo\nbar\n')
-    tmpdir.join('f2').write_binary(b'[INFO] hi\n')
-    tmpdir.join('f3').write_binary(b"with'quotes\n")
+    tmpdir.join('f1').write_binary('foo\nbar\n'.encode('UTF-32'))
+    tmpdir.join('f2').write_binary('[INFO] hi\n'.encode('UTF-32'))
+    tmpdir.join('f3').write_binary("with'quotes\n".encode('UTF-32'))
     with tmpdir.as_cwd():
         yield
 
```

asottile: and that's fine, that's not what I'm concerned about here

scop: Well, pygrep's results are defined against UTF-8 files only anyway. Given that premise I don't see anything wrong with attempting to display UTF-8 column offsets for its matches (and FWIW there's the fallback in case the calculation fails loudly). In the cases where the offsets could be wrong, the whole match result could be wrong too and that to me is a much bigger concern. I wonder why you aren't concerned about that?

asottile: they're not though, if your pattern is ascii it'll match any ascii-compatible encoding

scop: Yep, that would be the lucky case I was referring to earlier. A pattern in one encoding can also match something completely different in terms of characters on files in some other incompatible encoding.

PR #1324, opened 2020-02-10
Activate conda environment in pre-commit hook.

QuantCo-mr-T:
Save conda environment that was active during conda install when using
option --hooks-activate-conda. The saved environment will be activated
before calling pre-commit hooks.

Especially on Windows, more and more actions within a conda environment
require the conda environment to be activated. Thus saving just the
python executable is not enough any more.

There is currently one downside of using the option
--hooks-activate-conda. It uses "conda run" which will only show
console output after the run is completed. We have a pull request to
conda open which introduces an option to show interactive console
output in conda run. Once this is approved, it might be ok to make this
option the default behaviour.

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -73,6 +73,7 @@ def _install_hook_script(
         hook_type: str,
         overwrite: bool = False,
         skip_on_missing_config: bool = False,
+        hooks_activate_conda: bool = False,
         git_dir: Optional[str] = None,
 ) -> None:
     hook_path, legacy_path = _hook_paths(hook_type, git_dir=git_dir)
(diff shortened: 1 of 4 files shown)


QuantCo-mr-T:
I am new to pull requests in pre-commit. Any feedback how to submit a useful pull request here is highly welcome.
I needed to change the expected response of git for testing on my machine. I can revert that part if you like. I only added optional lines in the git output which occurred in my output.

asottile:
hi sorry, next time please make an issue and discuss what actual problem you're having -- sorry you've spent so much time on this but it's not something that should be done.

the golden path is hooks which work without external dependencies or intervention, anything else should be done in the hook itself or via local hooks (for example `entry: conda run ...`) -- but these are outside the supported path (pre-commit installing the tools in isolated environments and running them in isolated environments)

Issue #1120, opened 2019-08-16
Support post-checkout


domenkozar:
I know this goes beyond the original intention (and the name), but it would be really useful to be able to use same pre-commit hooks for different git hooks.

We need to run some validation before checking out. The alternative is to use https://github.com/sds/overcommit, but I'd prefer for pre-commit to get better.

asottile:
sure! there's already a few other `git` hooks supported and it shouldn't be too much work to add another one -- would you like to take a stab at it?  here's where `prepare-commit-msg` was added for example: https://github.com/pre-commit/pre-commit/pull/1004

given the others are currently file-based there might need to be some design aspects to this particular hook since `post-checkout` doesn't really operate on files

asottile:
oh right -- this got released as part of [2.2.0](https://github.com/pre-commit/pre-commit/releases/tag/v2.2.0) -- thanks again for the issue!

PR #1339, opened 2020-02-20
Add post-checkout

andrewhare:
This adds support for the `post-checkout` hook.

Fixes: https://github.com/pre-commit/pre-commit/issues/1120
Documentation PR: https://github.com/pre-commit/pre-commit.github.io/pull/307

One thing I would still like to do, if possible, would be to force hooks like this to always have `always_run` be set since that is somewhat implied by the nature of the hook (since it doesn't operate on files).

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -74,6 +74,7 @@ def _ns(
         remote_name: Optional[str] = None,
         remote_url: Optional[str] = None,
         commit_msg_filename: Optional[str] = None,
+        checkout_type: Optional[str] = None,
 ) -> argparse.Namespace:
     return argparse.Namespace(
         color=color,
@@ -84,6 +85,7 @@ def _ns(
         remote_url=remote_url,
         commit_msg_filename=commit_msg_filename,
         all_files=all_files,
+        checkout_type=checkout_type,
         files=(),
         hook=None,
         verbose=False,
@@ -157,6 +159,11 @@ def _run_ns(
         return _ns(hook_type, color, commit_msg_filename=args[0])
     elif hook_type in {'pre-merge-commit', 'pre-commit'}:
         return _ns(hook_type, color)
+    elif hook_type == 'post-checkout':
+        return _ns(
+            hook_type, color, source=args[0], origin=args[1],
+            checkout_type=args[2],
+        )
     else:
         raise AssertionError(f'unexpected hook type: {hook_type}')
 
--- a/pre_commit/constants.py
+++ b/pre_commit/constants.py
@@ -18,7 +18,7 @@ VERSION = importlib_metadata.version('pre_commit')
 # `manual` is not invoked by any installed git hook.  See #719
 STAGES = (
     'commit', 'merge-commit', 'prepare-commit-msg', 'commit-msg', 'manual',
-    'push',
+    'post-checkout', 'push',
 )
 
 DEFAULT = 'default'
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -79,7 +79,7 @@ def _add_hook_type_option(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         '-t', '--hook-type', choices=(
             'pre-commit', 'pre-merge-commit', 'pre-push',
-            'prepare-commit-msg', 'commit-msg',
+            'prepare-commit-msg', 'commit-msg', 'post-checkout',
         ),
         action=AppendReplaceDefault,
         default=['pre-commit'],
@@ -92,11 +92,17 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
     parser.add_argument('--verbose', '-v', action='store_true', default=False)
     parser.add_argument(
         '--origin', '-o',
-        help="The origin branch's commit_id when using `git push`.",
+        help=(
+            "The origin branch's commit_id when using `git push`.  "
+            'The ref of the previous HEAD when using `git checkout`.'
+        ),
     )
     parser.add_argument(
         '--source', '-s',
-        help="The remote branch's commit_id when using `git push`.",
+        help=(
+            "The remote branch's commit_id when using `git push`.  "
+            'The ref of the new HEAD when using `git checkout`.'
+        ),
     )
     parser.add_argument(
         '--commit-msg-filename',
@@ -122,6 +128,14 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
     mutex_group.add_argument(
         '--files', nargs='*', default=[],
         help='Specific filenames to run hooks on.',
+    )
+    parser.add_argument(
+        '--checkout-type',
+        help=(
+            'Indicates whether the checkout was a branch checkout '
+            '(changing branches, flag=1) or a file checkout (retrieving a '
+            'file from the index, flag=0).'
+        ),
     )
 
 
--- a/tests/commands/install_uninstall_test.py
+++ b/tests/commands/install_uninstall_test.py
@@ -20,6 +20,7 @@ from testing.fixtures import add_config_to_repo
 from testing.fixtures import git_dir
 from testing.fixtures import make_consuming_repo
 from testing.fixtures import remove_config_from_repo
+from testing.fixtures import write_config
 from testing.util import cmd_output_mocked_pre_commit_home
 from testing.util import cwd
 from testing.util import git_commit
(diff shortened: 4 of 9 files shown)


pre_commit/main.py
```diff
@@ -123,6 +123,18 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
...
         help='Specific filenames to run hooks on.',
     )
+    parser.add_argument(
+        '--prev-head-ref',
+        help='The ref of the previous HEAD.',
+    )
+    parser.add_argument(
+        '--new-head-ref',
+        help='The ref of the new HEAD (which may or may not have changed).',
+    )
```
asottile: I think we could potentially reuse `--source` / `--origin` for these arguments?

tests/conftest.py
```diff
@@ -180,6 +180,49 @@ def failing_prepare_commit_msg_repo(tempdir_factory):
...
+            'always_run': True,
+            'verbose': True,
+            'stages': ['post-checkout'],
+        }],
+    }
+    write_config(path, config)
+    with cwd(path):
+        cmd_output('git', 'add', '.')
+        git_commit()
+        yield path
```
asottile: since these are only used once, it probably makes sense to inline them adjacent to the test itself

pre_commit/main.py
andrewhare: I can do that if you want to keep the arg count down. However, I don't think that source/origin maps directly to new-head-ref/prev-head-ref so it might create some confusion. I'll leave that decision up to you.

tests/conftest.py
andrewhare: Do you want me to inline the entire fixture or just the `git add` / `git commit` part of this fixture?

andrewhare:
Thanks for the feedback, I will fix the CI issues later today.

pre_commit/main.py
asottile: `source` / `origin` is a "from ref" -> "to ref" relation, so I think it fits well enough

tests/conftest.py
asottile: ah mostly that there isn't a reason to put it in `conftest.py` unless it is being shared across multiple tests

andrewhare: Sounds good to me, I will move it.

pre_commit/main.py
andrewhare: Ok cool, I will do that.

andrewhare:
@asottile I made the changes you requested and CI is now green as well, please take another look when you can.

pre_commit/commands/hook_impl.py
```diff
@@ -157,6 +159,11 @@ def _run_ns(
         return _ns(hook_type, color, commit_msg_filename=args[0])
     elif hook_type in {'pre-merge-commit', 'pre-commit'}:
         return _ns(hook_type, color)
+    elif hook_type == 'post-checkout':
+        return _ns(
+            hook_type, color, origin=args[0], source=args[1],
```
asottile: origin and source are actually the opposite direction, the forward diff is `ORIGIN...SOURCE`

pre_commit/constants.py
```diff
@@ -18,7 +18,7 @@
 # `manual` is not invoked by any installed git hook.  See #719
 STAGES = (
     'commit', 'merge-commit', 'prepare-commit-msg', 'commit-msg', 'manual',
-    'push',
+    'push', 'post-checkout',
```
asottile: po < pu (alphabetize)

pre_commit/main.py
```diff
@@ -92,11 +92,13 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
     parser.add_argument('--verbose', '-v', action='store_true', default=False)
     parser.add_argument(
         '--origin', '-o',
-        help="The origin branch's commit_id when using `git push`.",
+        help="The origin branch's commit_id when using `git push`. "
+        + 'The ref of the previous HEAD when using `git checkout`.',
```
asottile: instead of ugly `+` use parens to continue lines (see below there's a few examples -- search for `help=(`)

pre_commit/main.py
```diff
@@ -123,6 +125,12 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
         '--files', nargs='*', default=[],
         help='Specific filenames to run hooks on.',
     )
+    parser.add_argument(
+        '--checkout-type',
+        help='A flag indicating whether the checkout was a branch checkout '
```
asottile: "A flag" is redundant, of course it's a flag :)

tests/commands/install_uninstall_test.py
```diff
@@ -725,6 +726,53 @@ def test_commit_msg_legacy(commit_msg_repo, tempdir_factory, store):
     assert second_line.startswith('Must have "Signed off by:"...')
 
 
+def test_post_checkout_integration_failing(tempdir_factory, store):
```
asottile: hmmm I'm pretty sure the exit code of `post-checkout` is ignored right? so we can probably skip this test 

> This hook cannot affect the outcome of git switch or git checkout.

tests/commands/run_test.py
```diff
@@ -466,6 +466,12 @@ def test_all_push_options_ok(cap_out, store, repo_with_passing_hook):
     assert b'Specify both --origin and --source.' not in printed
 
 
+def test_checkout_type(cap_out, store, repo_with_passing_hook):
+    args = run_opts(origin='', source='', checkout_type='1')
+    ret, printed = _do_run(cap_out, store, repo_with_passing_hook, args)
+    assert ret == 0
```
asottile: this doesn't really assert anything?

tests/commands/install_uninstall_test.py
andrewhare: The exit code is ignored by Git but it is still reported by `pre-commit` to the user. I can remove the test but I think it's still a useful thing to have because the hook can actually return non-zero.

tests/commands/run_test.py
andrewhare: This test is here to provide coverage of this line: https://github.com/pre-commit/pre-commit/pull/1339/files#diff-d58a98662b1c7c5efc1562af39480143R320. Without this test, coverage goes down to 99%.

asottile: the test should probably demonstrate something about that then (perhaps print out in the hook and show that it was passed through)

tests/commands/install_uninstall_test.py
asottile: I'd probably remove it, it doesn't cause a change in behaviour for the git command that's run so the test doesn't really demonstrate anything

andrewhare:
Yeah, your welcome! It’s was fun, you’ve got a great project. Thanks for
all your input!

asottile:
I'm also going to follow up and add aliases for `--source` and `--origin`, probably `--from-ref` and `--to-ref` since the current names are definitely confusing 😆

asottile:
just played with this a bit, not sure if it's intentional but the `post-checkout` hooks are currently getting the list of files different between the two branches

I *think* they're not supposed to get any files at all so I'm making a change to fix that 👍

asottile:
```console
$ ~/workspace/pre-commit/venv/bin/pre-commit install -t post-checkout
pre-commit installed at .git/hooks/post-checkout
$ git add .
$ git commit -m 'add config'
[master (root-commit) d15280a] add config
 1 file changed, 5 insertions(+)
 create mode 100644 .pre-commit-config.yaml
$ git status
On branch master
nothing to commit, working tree clean
$ git status
On branch master
nothing to commit, working tree clean
$ git checkout master -b wat
Switched to a new branch 'wat'
identity.............................................(no files to check)Skipped
- hook id: identity
$ touch foo.py
$ git status
On branch wat
Untracked files:
	foo.py

nothing added to commit but untracked files present
$ git add .
$ git commit -m "add foo"
[wat 1ca3ef0] add foo
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo.py
$ git checkout master 
Switched to branch 'master'
identity.............................................(no files to check)Skipped
- hook id: identity
$ git checkout wat
Switched to branch 'wat'
identity.................................................................Passed
- hook id: identity
- duration: 0.04s

foo.py

$ cat .pre-commit-config.yaml 
repos:
-   repo: meta
    hooks:
    -   id: identity
        stages: [post-checkout]        
```

asottile:
#1344 for that one!

andrewhare:
@asottile Nice catch on not needing the files, I missed that!

https://github.com/pre-commit/pre-commit/pull/1344 looks like a great fix.

asottile:
oh right -- this got released as part of [2.2.0](https://github.com/pre-commit/pre-commit/releases/tag/v2.2.0) -- thanks again for the patch!

PR #1369, opened 2020-03-14
Fix test failure of test_installed_from_venv when running with pytest-runner

felixonmars:
Passing PYTHONPATH here fixes the module not found error when running with pytest-runner without pre-commit installed first.

### DIFF
--- a/tests/commands/install_uninstall_test.py
+++ b/tests/commands/install_uninstall_test.py
@@ -530,6 +530,8 @@ def test_installed_from_venv(tempdir_factory, store):
                 'GIT_COMMITTER_NAME': os.environ['GIT_COMMITTER_NAME'],
                 'GIT_AUTHOR_EMAIL': os.environ['GIT_AUTHOR_EMAIL'],
                 'GIT_COMMITTER_EMAIL': os.environ['GIT_COMMITTER_EMAIL'],
+                # Pass PYTHONPATH to support running tests using pytest-runner
+                'PYTHONPATH': os.environ.get('PYTHONPATH', ''),
             },
         )
         assert ret == 0


asottile:
[pytest-runner is deprecated](https://github.com/pytest-dev/pytest-runner#deprecation-notice) and not a supported way to run our tests -- see the [contributing guide](https://github.com/pre-commit/pre-commit/blob/master/CONTRIBUTING.md)

rharish101:
Hi @asottile I'm facing this error even when using bare pytest, i.e. without tox or pytest-runner, to test a build of pre-commit in a clean Arch Linux chroot. The proposed fix works for me, so could you reconsider this pull request?

asottile:
if PYTHONPATH is fixing things then you haven't installed properly and aren't running the tests in a support manner -- please see the contributing guide

I will not accept this as a pull request, PYTHONPATH is always the wrong solution

rharish101:
According to the contributing guide, I should set up a virtual environment with dependencies installed from PyPI and run the tests. However, when creating a package in a chroot, I want to use the dependencies installed in the chroot through the distro's repos instead of what is installed in the virtual environment through PyPI.

This should work fine, but the problem is that the tests require `pre-commit` to be "installed" (albeit through an egg-link to the built package) in the virtual environment, since some tests use `importlib` to get info about `pre-commit`. Since I'm not using a virtual environment, I'm creating an egg-info using `python setup.py egg_info` in the repo's root directory. I'm making `importlib` read this by setting the repo's root in `PYTHONPATH`.

Clearly, you pointed out that my approach is wrong, but I'm not finding a solution online. Could you help me out?

asottile:
no sorry, I don't have the time budget to help you do an unsupported thing for free

rharish101:
Ah okay, I'll try to figure it out myself. Thanks anyway.

PR #1377, opened 2020-03-27
fix CJK characters width in output

sophgn:


sophgn:
emmm, imports are reordered by pre-commit... I will close this pr and open another one

PR #1378, opened 2020-03-27
fix CJK characters width in output

sophgn:


sophgn:
5a03bc20f247c58c
before and after

pre_commit/commands/run.py
```diff
@@ -33,8 +34,12 @@
 logger = logging.getLogger('pre_commit')
 
 
+def _len_cjk(msg: str) -> int:
+    return len(msg) + sum(east_asian_width(c) in 'FW' for c in msg)
```
asottile: I would prefer `in {'F', 'W'}`

pre_commit/commands/run.py
```diff
@@ -13,6 +13,7 @@
 from typing import Sequence
 from typing import Set
 from typing import Tuple
+from unicodedata import east_asian_width
```
asottile: I tend to prefer `import unicodedata` and then `unicodedata.east_asian_width` for stdlib modules

pre_commit/commands/run.py
```diff
@@ -13,9 +13,9 @@
 from typing import Sequence
 from typing import Set
 from typing import Tuple
-from unicodedata import east_asian_width
 
 from identify.identify import tags_from_path
+from wcwidth import wcswidth
```
asottile: `wcwidth` introduces a dependency, I'd rather not -- `unicodedata` is fine for now (this is only accidentally working due to `pytest`)

PR #1380, opened 2020-04-01
Remove virtualenv dependency.

thequicksort:
Hi there, 
First of all, thank you for all your hard work building and maintaining this package. It's already been a tremendous help to my workflow. 

I might be missing something, but from my understanding of the code, it seems like all the virtualenv functionalities being used by pre-commit are already taken care of by venv. Venv has been a core Python module [since v3.3](https://docs.python.org/3/library/venv.html), which means it's safe to assume in this project  (whose minimum version is 3.6 and  Python 2 is no longer supported).

With all that in mind, is there a specific reason for keeping the virtualenv dependency around around? If not, can it be removed?

### Advantages 

- Smaller package size (potentially?)

- One less dependency to be managed, which is great for those of us using less package managers like Nix. 

### Disadvantages 

- None that occur to me immediately (happy to hear otherwise!) 

## Testing 

To test, I ran the test suite on Mac OS X (10.14.6) and Windows 10. 

```bash
pre-commit run --all-files --show-diff-on-failure
```

### Mac OS 

- Success 

Screen Shot 2020-03-31 at 10 42 30 PM

### Windows 10

- All hooks pass except MyPy. 

- However
  - The files that are failing are unchanged in this commit 
  - This MyPy check seems to fail in master as well (proof in screenshot)

Screen Shot 2020-03-31 at 10 17 27 PM

Screen Shot 2020-03-31 at 11 10 10 PM

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -202,7 +202,7 @@ def py_interface(
 
 def make_venv(envdir: str, python: str) -> None:
     env = dict(os.environ, VIRTUALENV_NO_DOWNLOAD='1')
-    cmd = (sys.executable, '-mvirtualenv', envdir, '-p', python)
+    cmd = (sys.executable, '-mvenv', envdir)
     cmd_output_b(*cmd, env=env, cwd='/')
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,6 @@ install_requires =
     nodeenv>=0.11.1
     pyyaml>=5.1
     toml
-    virtualenv>=15.2
     importlib-metadata;python_version<"3.8"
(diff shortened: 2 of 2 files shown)


asottile:
no sorry, this is intentional -- `virtualenv` allows us to do cross version installations in a more portable way than venv -- if you search through the issue tracker there's a bit about venv in a couple of places

the actual intention is to eventually remove `venv` support because `virtualenv` does it better in every way (faster, more portable, cross version, etc.)

asottile:
the most important thing this patch would break is the default installation on debian and ubuntu which is a no go

thequicksort:
Ah gotcha, thank you for explaining!

PR #1391, opened 2020-04-13
Provide an environment variable to override the pre-commit command.

chrisgilmerproj:
In this PR I'm modifying the pre-commit hook template to provide a new environment variable. Using the variable `PRE_COMMIT_CMD` will override both methods of running: `INSTALL_PYTHON` and the fallback of `which('pre-commit')`. 

---

Why would I want this? In my case I am wrapping the `pre-commit` utility in a shell script which runs the command inside a docker container where I've pre-installed the `pre-commit` utility and mounted the directory that I'm running against. The wrapper works fine, allowing me to run `pre-commit run -a` both in and outside that docker container. But when the pre-commit hook for git is run I still find that it's asking for the locally installed python or the locally installed pre-commit utility. 

Another way to do this would be to allow me to choose to use `which('pre-commit')` instead of the `INSTALL_PYTHON` should I set a similar environment variable. I didn't write a PR for that one but I could.

I am also not committed to the choice of the environment variable name, that could definitely be changed if it helps folks understand its meaning better.

Thanks for the review and I'm happy to answer any questions.


asottile:
I'd rather not support this, it opens the door for an unsupported and difficult to debug situation.  I'd suggest if you're going to set up a special situation that you write out the scripts for your special situation and support it 👍

an aside: it's pretty funny to submit a PR to pre-commit, _but not run the pre-commit hooks_ 😆

chrisgilmerproj:
> I'd rather not support this, it opens the door for an unsupported and difficult to debug situation. I'd suggest if you're going to set up a special situation that you write out the scripts for your special situation and support it 👍
> 
> an aside: it's pretty funny to submit a PR to pre-commit, _but not run the pre-commit hooks_ 😆

That's completely fair. Was curious if other folks would find utility in this. Like I said, I wish I could choose between the options in the existing hook. If you have an interest in that PR I'd be happy to write it.

asottile:
by the way, pre-commit has a builtin way to support containerized hooks: https://pre-commit.com/#docker https://pre-commit.com/#docker https://pre-commit.com/#docker_image

chrisgilmerproj:
> by the way, pre-commit has a builtin way to support containerized hooks: https://pre-commit.com/#docker https://pre-commit.com/#docker https://pre-commit.com/#docker_image

It's more that I want to containerize pre-commit itself, not just the hooks. I'll keep noodling on this :)

asottile:
I'm curious why you would want to do that?

chrisgilmerproj:
> I'm curious why you would want to do that?

The goal for my group is to move all development into docker containers, ultimately to make distribution of tools (and versions of tools) much easier. But for some folks they can't commit their changes inside the docker containers because they have GPG signing of their commits enabled. So I'm trying to make it work so they can commit outside the docker container with `git` while the `pre-commit` utility is run inside the container. 

I actually was able to do this with not a lot of elbow grease. But it would be easier if I could just modify an environment variable for some of the folks (not all of them) and enable this behavior I want.

asottile:
You could [mount the gpg keys](https://github.com/deadsnakes/runbooks/blob/fbedbe2b7d5a32c6d067d8784db54244d843fd6e/build#L60-L61) fwiw

chrisgilmerproj:
> You could [mount the gpg keys](https://github.com/deadsnakes/runbooks/blob/fbedbe2b7d5a32c6d067d8784db54244d843fd6e/build#L60-L61) fwiw

Nice! I may try that. I'm certainly not at the end of this journey, its been ... interesting. FWIW thank you for working with me on this, appreciate the time you put into it. And I *love* pre-commit!

PR #1402, opened 2020-04-18
Preserve line ending when running autoupdate

utek:
This PR aims to make pre-commit's `autoupdate` option preserve the line endings.

This is slight inconvenience when using windows but having all code use Unix type line endings.

### DIFF
--- a/tests/commands/autoupdate_test.py
+++ b/tests/commands/autoupdate_test.py
@@ -263,6 +263,45 @@ def test_does_not_reformat(tmpdir, out_of_date, store):
     assert cfg.read() == expected
 
 
+def test_does_not_change_mixed_endlines_read(up_to_date, tmpdir, store):
+    fmt = (
+        'repos:\n'
+        '-   repo: {}\n'
+        '    rev: {}  # definitely the version I want!\r\n'
+        '    hooks:\r\n'
+        '    -   id: foo\n'
+        '        # These args are because reasons!\r\n'
+        '        args: [foo, bar, baz]\r\n'
+    )
+    cfg = tmpdir.join(C.CONFIG_FILE)
+
+    expected = fmt.format(up_to_date, git.head_rev(up_to_date)).encode()
+    cfg.write_binary(expected)
+
+    assert autoupdate(str(cfg), store, freeze=False, tags_only=False) == 0
+    assert cfg.read_binary() == expected
+
+
+def test_does_not_change_mixed_endlines_write(tmpdir, out_of_date, store):
+    fmt = (
+        'repos:\n'
+        '-   repo: {}\n'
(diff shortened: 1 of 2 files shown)


tests/commands/autoupdate_test.py
```diff
@@ -263,6 +263,27 @@ def test_does_not_reformat(tmpdir, out_of_date, store):
...
+        'repos:\n'
+        '-   repo: {}\n'
+        '    rev: {}  # definitely the version I want!\r\n'
+        '    hooks:\r\n'
+        '    -   id: foo\n'
+        '        # These args are because reasons!\r\n'
+        '        args: [foo, bar, baz]\r\n'
+    )
+    cfg = tmpdir.join(C.CONFIG_FILE)
+    expected = bytes(
```
asottile, line 278: generally you should never call `bytes` on a `str` -- use the `.encode()` method (you also don't need a parameter to encode as python3+ defaults to UTF-8

tests/commands/autoupdate_test.py
```diff
@@ -263,6 +263,27 @@ def test_does_not_reformat(tmpdir, out_of_date, store):
     assert cfg.read() == expected
 
 
+def test_does_not_change_mixed_endlines(up_to_date, tmpdir, store):
+    """Test that line endings are not being changed."""
```
asottile, line 267: this docstring is redundant with the test name, I would remove it

tests/commands/autoupdate_test.py
```diff
@@ -263,6 +263,27 @@ def test_does_not_reformat(tmpdir, out_of_date, store):
...
+        '        args: [foo, bar, baz]\r\n'
+    )
+    cfg = tmpdir.join(C.CONFIG_FILE)
+    expected = bytes(
+        fmt.format(up_to_date, git.head_rev(up_to_date)), 'utf-8',
+    )
+    cfg.write_binary(expected)
+
+    assert autoupdate(str(cfg), store, freeze=False, tags_only=False) == 0
+    assert cfg.read_binary() == expected
```
asottile, line 284: this test only covers half of your patch -- it would need to trigger the write case to cover the full patch

tests/commands/autoupdate_test.py
utek: Thank  you. Fixed

tests/commands/autoupdate_test.py
utek: I like docstrings in every method. But I see your point here. Removed

tests/commands/autoupdate_test.py
utek: You are right. It's me trying to simplify test. It did test write before I tried to "clean it up"

PR #1413, opened 2020-04-24
Update CONTRIBUTING.md

ModischFabrications:
add windows venv path to installation guide

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,8 @@ This is useful for running specific tests.  The easiest way to set this up
 is to run:
 
 1. `tox --devenv venv`  (note: requires tox>=3.13)
-2. `. venv/bin/activate`
+2. `. venv/bin/activate` (or follow the [activation instructions] for your
+   platform)
 
 This will create and put you into a virtualenv which has an editable
 installation of pre-commit.  Hack away!  Running `pre-commit` will reflect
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ This is useful for running specific tests.  The easiest way to set this up
 is to run:
 
 1. `tox --devenv venv`  (note: requires tox>=3.13)
-2. `. venv/bin/activate`
+2. `. venv/bin/activate` (POSIX) or `venv\Scripts\activate` (Windows)
```
asottile, line 20: specifically in `cmd.exe` -- if you're using `git bash` the former works

ModischFabrications: alright,  I was not aware that the generation from 2 differs by shell, not OS. Might want to add a sentence somewhere that everything should be executed with bash instead then?

asottile: I mean this patch is fine, I'm just suggesting it becomes a little bit more specific

the directions were written before most of the execution worked well on windows so it doesn't surprise me that there's parts that are missing

ModischFabrications: I'm back, sorry for the wait. I copied the definition from [the python documentation](https://docs.python.org/3/library/venv.html) , which won't specify it any further than "Windows". Powershell will act the same as cmd.exe, ksh & more like bash. Listing interpreters will probably have more edge cases than just calling git bash a POSIX-alike.

asottile: welllll we're using `virtualenv` and not `venv` so those docs don't precisely apply -- a better solution might be to do:

```markdown
[activate your virtualenv](link to docs...)
```

ModischFabrications: I'm not entirely convinced it's easier that way, but it's not worth fighting over. I changed it to your suggestions, feel free to merge if everything is alright.

PR #1422, opened 2020-04-29
fix hooks firing during staged_files_only

### LINKED ISSUE #1418: post-checkout hook prevents pre-commit hooks from working with unstaged files
Hi there!

Most important thing first: pre-commit is great, it makes managing git hooks trivial and when combined with static analysis hooks, helps our team immeasurably - **thanks for all the hard work**.

We have pre-commit hooks that run a bunch of checks, and thanks to #1339,  a post-checkout hook that does some workspace analysis and prints a few warnings if needed. 

Unfortunately this seems to have broken the "stash unstaged changes" behaviour when committing - the stash [runs a git checkout](https://github.com/pre-commit/pre-commit/blob/5a625013071877a5b5dc8eba5b25c791c9c9c6ec/pre_commit/staged_files_only.py#L60) to clear the working directory before running the hooks, and that fails because of the post-checkout hook. I'd expect that the working directory to be cleared as normal, commit OK and then the working directory to be restored as it does without the post-checkout hook, as the actual mechanism behind removing the unstaged changes and restoring them is hidden from the user.

Below is a quick test to reproduce it:

```bash
#!/usr/bin/env bash

mkdir test
cd test || exit
git init

# Configure a pre-commit and post-checkout hook
cat << EOF > .pre-commit-config.yaml
repos:
  - repo: local
    hooks:
      - id: pre-commit-ok
        name: Pre-commit hook that always passes
        stages: [commit]
        language: system
        entry: sh -c 'exit 0'
        always_run: true

  - repo: local
    hooks:
      - id: post-checkout-fail
        name: Post-checkout hook that fails
        stages: [post-checkout]
        language: system
        entry: sh -c 'exit 1'
        always_run: true

EOF

echo "bananas" > great-things.txt
git add .
git commit -m "init"

# Install the hooks
pre-commit install -t pre-commit
pre-commit install -t post-checkout

# Do a checkout - this is expected to print an error from the lint
git checkout -b test

# Change the original file
echo "platanos" >> great-things.txt

# Add another file and stage it
echo "42" > answer.txt
git add answer.txt

# Attempt to commit the new, staged file.
#
# Without the post-checkout hook, this causes the unstaged changes to be
# stashed and the commit works as normal.
echo ""
echo ""
echo "    Attempting to commit with unstaged changes"
echo "    This will fail when attempting to checkout as part of the commit hook:"
echo ""
echo ""
git commit -m "more info"
```

Unfortunately there seems to be no easy way to skip the post-checkout hook when clearing the unstaged files with `git checkout`.

Happy to open a PR to help fix this, just unsure what to do - I could swap the `git checkout` for a `git stash save --keep-index --include-untracked` but I assume it is using checkout for reason.

domodwyer:
pass through an environment variable to prevent hooks running

Fixes #1418.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -324,6 +324,12 @@ def run(
             f'`--hook-stage {args.hook_stage}`',
         )
         return 1
+    # prevent recursive post-checkout hooks (#1418)
+    if (
+            args.hook_stage == 'post-checkout' and
+            environ.get('_PRE_COMMIT_SKIP_POST_CHECKOUT')
+    ):
+        return 0
 
     # Expose from-ref / to-ref as environment variables for hooks to consume
     if args.from_ref and args.to_ref:
--- a/pre_commit/staged_files_only.py
+++ b/pre_commit/staged_files_only.py
@@ -56,8 +56,10 @@ def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
         with open(patch_filename, 'wb') as patch_file:
             patch_file.write(diff_stdout_binary)
 
-        # Clear the working directory of unstaged changes
-        cmd_output_b('git', 'checkout', '--', '.')
+        # prevent recursive post-checkout hooks (#1418)
+        no_checkout_env = dict(os.environ, _PRE_COMMIT_SKIP_POST_CHECKOUT='1')
+        cmd_output_b('git', 'checkout', '--', '.', env=no_checkout_env)
+
         try:
             yield
         finally:
@@ -72,8 +74,9 @@ def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
                 # We failed to apply the patch, presumably due to fixes made
                 # by hooks.
                 # Roll back the changes made by hooks.
-                cmd_output_b('git', 'checkout', '--', '.')
+                cmd_output_b('git', 'checkout', '--', '.', env=no_checkout_env)
                 _git_apply(patch_filename)
+
             logger.info(f'Restored changes from {patch_filename}.')
     else:
         # There weren't any staged files so we don't need to do anything
--- a/testing/util.py
+++ b/testing/util.py
@@ -103,10 +103,12 @@ def cwd(path):
         os.chdir(original_cwd)
 
 
-def git_commit(*args, fn=cmd_output, msg='commit!', **kwargs):
+def git_commit(*args, fn=cmd_output, msg='commit!', all_files=True, **kwargs):
     kwargs.setdefault('stderr', subprocess.STDOUT)
 
-    cmd = ('git', 'commit', '--allow-empty', '--no-gpg-sign', '-a') + args
+    cmd = ('git', 'commit', '--allow-empty', '--no-gpg-sign', *args)
+    if all_files:  # allow skipping `-a` with `all_files=False`
+        cmd += ('-a',)
     if msg is not None:  # allow skipping `-m` with `msg=None`
         cmd += ('-m', msg)
     ret, out, _ = fn(*cmd, **kwargs)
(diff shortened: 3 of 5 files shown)


domodwyer:
Running `git checkout-index --all --force` is not equivalent, so changed to using the env var to skip hooks triggered by the `git checkout` call.

pre_commit/commands/run.py
```diff
@@ -245,6 +245,11 @@ def _run_hooks(
         environ: EnvironT,
 ) -> int:
     """Actually run the hooks."""
+
+    # Skip running any hooks if _PRE_COMMIT_SKIP_POST_CHECKOUT is set (#1418)
+    if os.getenv('_PRE_COMMIT_SKIP_POST_CHECKOUT') is not None:
```
asottile, line 250: this should probably also check hook_stage to make sure it is `post-checkout`

pre_commit/staged_files_only.py
```diff
@@ -57,6 +57,12 @@ def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
             patch_file.write(diff_stdout_binary)
 
         # Clear the working directory of unstaged changes
+        #
+        # Do not use "git checkout" to avoid triggering post-checkout hooks.
+        #
+        #       https://github.com/pre-commit/pre-commit/issues/1418
+        #
+        os.environ['_PRE_COMMIT_SKIP_POST_CHECKOUT'] = '1'
```
asottile, line 65: there's `pre_commit.envcontext` to do this sort of thing -- but probably just need to pass it to the individual `checkout` calls instead of applying to the whole thing

can probably do `env = (**os.environ, _PRE_COMMIT...)`

domodwyer: I hadn't seen this, thanks!

pre_commit/commands/run.py
```diff
@@ -245,6 +245,15 @@ def _run_hooks(
...
 ) -> int:
     """Actually run the hooks."""
+
+    # Skip running any hooks if _PRE_COMMIT_SKIP_POST_CHECKOUT is set (#1418)
+    # and this is an execution for post-checkout hooks.
+    if (
+        args.hook_stage == 'post-checkout' and
+        os.getenv('_PRE_COMMIT_SKIP_POST_CHECKOUT') is not None
+    ):
+        return False
```
asottile, line 255: while `False` is a valid integer in python, the intention here is to be the program exit code so it would be more expected to `return 0`

pre_commit/staged_files_only.py
```diff
@@ -57,7 +58,14 @@ def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
             patch_file.write(diff_stdout_binary)
 
         # Clear the working directory of unstaged changes
-        cmd_output_b('git', 'checkout', '--', '.')
+        #
+        # Do not use "git checkout" to avoid triggering post-checkout hooks.
+        #
+        #       https://github.com/pre-commit/pre-commit/issues/1418
+        #
```
asottile, line 65: the whitespace and formatting is super unnecessary here.  this should be at most one line, and you can refer to an issue using its number like `#1418`

`cmd_output` / `cmd_output_b` take an `env` environment dictionary which would be better I think than `envcontext` here

asottile:
needs a test as well

pre_commit/staged_files_only.py
domodwyer: Sorry, didn't you say use `envcontext` above?

I can change it to use a dict instead, and I've made the comment one line.

domodwyer:
> needs a test as well

As I said, I'm trying to help without really knowing python - the tests are particularly hard to pick apart.

pre_commit/staged_files_only.py
asottile: I indicated that your before/after thing is better done with `envcontext`, but that in this case an env dictionary is best

asottile:
oh ok, I can help wiht the test

domodwyer:
Great cat, thanks for your help!

asottile:
thanks again for the fix and report, this has been released as part of v2.4.0 🎉

PR #1424, opened 2020-05-01
Allow local config only hooks

scop:
Some authors of very useful tools are reluctant to add and host .pre-commit-hooks.yml to their repos for various reasons.

As we can already override everything in local configs anyway, we could actually operate solely on local configs. This way, no .pre-commit-hooks.yaml is needed in a target repo. I think making this possible would be very valuable.

This is more or less a PoC in its current state; I haven't dug deeper to look for dragons, and docs and some polishing are required (the access to cfgv might be  bit icky the way it is now, and we print the "Proceeding with local configs" info all the time when it could do so only on install/update), but in its current form I can for example Just Use shfmt with this:
```yaml
  - repo: https://github.com/mvdan/sh
    rev: v3.1.0
    hooks:
      - id: shfmt
        name: shfmt
        language: golang
        entry: shfmt
        args: [-w]
        types: [shell]
```
...and perlcritic with:
```yaml
  - repo: https://github.com/Perl-Critic/Perl-Critic
    rev: v1.138
    hooks:
      - id: perlcritic
        name: perlcritic
        language: perl
        entry: perlcritic
        args: [--quiet, --verbose, "5"]
        types: [perl]
```
...in my .pre-commit-config.yaml even though the target repos have no .pre-commit-hooks.yaml at all.

### DIFF
--- a/pre_commit/repository.py
+++ b/pre_commit/repository.py
@@ -10,9 +10,11 @@ from typing import Set
 from typing import Tuple
 
 import pre_commit.constants as C
+from pre_commit.clientlib import apply_manifest_defaults
 from pre_commit.clientlib import load_manifest
 from pre_commit.clientlib import LOCAL
 from pre_commit.clientlib import META
+from pre_commit.clientlib import validate_manifest
 from pre_commit.hook import Hook
 from pre_commit.languages.all import languages
 from pre_commit.languages.helpers import environment_dir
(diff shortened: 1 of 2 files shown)


pre_commit/repository.py
```diff
@@ -146,21 +148,37 @@ def _cloned_repository_hooks(
...
-
-    for hook in repo_config['hooks']:
-        if hook['id'] not in by_id:
-            logger.error(
-                f'`{hook["id"]}` is not present in repository {repo}.  '
-                f'Typo? Perhaps it is introduced in a newer version?  '
-                f'Often `pre-commit autoupdate` fixes this.',
-            )
-            exit(1)
+    if os.path.isfile(manifest_path):
```
scop, line 151: This is currently all or nothing: if target repo has a manifest, we require all configured hooks to be found in it. I think it could be better so that we use whatever is there, but allow additional local-config-only ones (i.e. if a hook is not found in manifest, use local-only config for it).

scop: Also, we could make all of this new stuff optional on per repo basis if this does not seem like something we'd also want to unconditionally everywhere, e.g. add a repo level `allow_local_only` or the like boolean option to trigger the new behavior with default being whatever we think is best.

asottile:
no sorry, this is what `repo: local` is for already -- I'm not making a second thing for that

asottile:
please discuss large issues in features before working on them to save the effort -- thanks!

scop:
Oh wow, didn't realize that local can be used to achieve this. And I agree adding overlap such as in this PR should not be added in that case, if it works (seems at least to some extent). No worries with the throwaway work, wasn't much at all and left it at unpolished poc state on purpose.

By discussing in features, do you mean filing issues here?

Anyway, I ran into some issues trying to configure https://github.com/mvdan/cc with a `local` to use `shfmt`, seems there may be some generic golang ones lurking. Will see if I can get around those and file issues/PR's accordingly.

asottile:
probably need to set up another `language` specifically for go modules -- though there are some other go mod tools already in use so it's unclear to me why shfmt is different there 🤔

PR #1435, opened 2020-05-04
Maintain scalar quoting style when autoupdate re-writes rev

### LINKED ISSUE #1434: Any scalar quotes around rev are lost during autoupdate
#### Steps to reproduce

Example `.pre-commit-config.yaml`:
```
repos:
-   repo: 'https://github.com/pre-commit/pre-commit-hooks'
    rev: 'v2.4.0'
    hooks:
    -   id: 'trailing-whitespace'
    -   id: 'end-of-file-fixer'
    -   id: 'check-yaml'
    -   id: 'check-added-large-files'
```

Run 
```
$ pre-commit autoupdate
Updating https://github.com/pre-commit/pre-commit-hooks ... updating v2.4.0 -> v2.5.0.
```

`.pre-commit-config.yaml` after:
```
repos:
  - repo: 'https://github.com/pre-commit/pre-commit-hooks'
    rev: v2.5.0
    hooks:
      - id: 'trailing-whitespace'
      - id: 'end-of-file-fixer'
      - id: 'check-yaml'
      - id: 'check-added-large-files'
```
Any wrapping quotes (single or double) around the `rev` value are lost during the update

I noticed this when I turned on the `quoted-strings` [yamllint rule](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings) and set it to `required` to be consistent with our internal YAML coding conventions

marcjay:
If `rev` is wrapped in single or double quotes (e.g. due to a yamllint quoted-strings rule), when
re-writing the `rev` to update it, honour the existing quotation style

Fixes #1434

### DIFF
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -84,7 +84,9 @@ def _check_hooks_still_exist_at_rev(
         )
 
 
-REV_LINE_RE = re.compile(r'^(\s+)rev:(\s*)([^\s#]+)(.*)(\r?\n)$', re.DOTALL)
+REV_LINE_RE = re.compile(
+    r'^(\s+)rev:(\s*)([\'"]?)([^\s#]+)(.*)(\r?\n)$', re.DOTALL,
+)
 
 
 def _original_lines(
@@ -116,15 +118,15 @@ def _write_new_config(path: str, rev_infos: List[Optional[RevInfo]]) -> None:
             continue
         match = REV_LINE_RE.match(lines[idx])
         assert match is not None
-        new_rev_s = yaml_dump({'rev': rev_info.rev})
+        new_rev_s = yaml_dump({'rev': rev_info.rev}, default_style=match[3])
         new_rev = new_rev_s.split(':', 1)[1].strip()
         if rev_info.frozen is not None:
             comment = f'  # frozen: {rev_info.frozen}'
-        elif match[4].strip().startswith('# frozen:'):
+        elif match[5].strip().startswith('# frozen:'):
             comment = ''
         else:
-            comment = match[4]
-        lines[idx] = f'{match[1]}rev:{match[2]}{new_rev}{comment}{match[5]}'
+            comment = match[5]
+        lines[idx] = f'{match[1]}rev:{match[2]}{new_rev}{comment}{match[6]}'
 
     with open(path, 'w', newline='') as f:
         f.write(''.join(lines))
--- a/pre_commit/util.py
+++ b/pre_commit/util.py
@@ -36,10 +36,11 @@ yaml_load = functools.partial(yaml.load, Loader=Loader)
 Dumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
 
 
-def yaml_dump(o: Any) -> str:
+def yaml_dump(o: Any, **kwargs: Any) -> str:
     # when python/mypy#1484 is solved, this can be `functools.partial`
     return yaml.dump(
         o, Dumper=Dumper, default_flow_style=False, indent=4, sort_keys=False,
+        **kwargs,
     )
 
 
--- a/tests/commands/autoupdate_test.py
+++ b/tests/commands/autoupdate_test.py
@@ -474,3 +474,23 @@ def test_updates_old_format_to_new_format(tmpdir, capsys, store):
     )
     out, _ = capsys.readouterr()
     assert out == 'Configuration has been migrated.\n'
+
+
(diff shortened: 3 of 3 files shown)


pre_commit/util.py
```diff
@@ -36,10 +36,11 @@
 Dumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
 
 
-def yaml_dump(o: Any) -> str:
+def yaml_dump(o: Any, style: Optional[str] = None) -> str:
```
asottile, line 39: this should be named-only if anything and match the name of the argument from yaml

pre_commit/util.py
```diff
@@ -36,10 +36,11 @@
...
     # when python/mypy#1484 is solved, this can be `functools.partial`
```
asottile, line 40: this comment would now be incorrect with this change

tests/commands/autoupdate_test.py
```diff
@@ -474,3 +474,23 @@ def test_updates_old_format_to_new_format(tmpdir, capsys, store):
...
+
+def test_maintains_rev_quoting_style(tmpdir, out_of_date, store):
+    fmt = (
+        'repos:\n'
+        '-   repo: {0}\n'
+        '    rev: "{1}"\n'
+        '    hooks:\n'
+        '    -   id: foo\n'
+        '-   repo: {0}\n'
+        '    rev: \'{1}\'\n'
```
asottile, line 487: can avoid escapes by doing this:
```suggestion
        "    rev: '{1}'\n"
```

tests/commands/autoupdate_test.py
```diff
@@ -474,3 +474,23 @@ def test_updates_old_format_to_new_format(tmpdir, capsys, store):
     )
     out, _ = capsys.readouterr()
     assert out == 'Configuration has been migrated.\n'
+
+
+def test_maintains_rev_quoting_style(tmpdir, out_of_date, store):
+    fmt = (
+        'repos:\n'
+        '-   repo: {0}\n'
```
asottile, line 482: would prefer named formats if there's repeats

pre_commit/commands/autoupdate.py
```diff
@@ -84,7 +84,9 @@ def _check_hooks_still_exist_at_rev(
         )
 
 
-REV_LINE_RE = re.compile(r'^(\s+)rev:(\s*)([^\s#]+)(.*)(\r?\n)$', re.DOTALL)
+REV_LINE_RE = re.compile(
+    r'^(\s+)rev:(\s*)([\'"]?)([^\s#]+)([\'"]?)(.*)(\r?\n)$', re.DOTALL,
```
asottile, line 88: git tags can contain and end with quotes so this regex isn't quite correct

marcjay: Ah, definitely hadn't considered that. On the basis that leading quotes in the tag would need to be escaped and so correctly not match, and that there's no benefit to trying (and currently failing due to greedy regex) to match any closing quotes, did you have this in mind? (dropping the second quote match)
```
^(\s+)rev:(\s*)([\'"]?)([^\s#]+)(.*)(\r?\n)$
```

asottile: yeah I think that would work, I can't think of any problems with that off the top of my head 🤔

pre_commit/util.py
```diff
@@ -36,10 +36,10 @@
 Dumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
 
 
-def yaml_dump(o: Any) -> str:
-    # when python/mypy#1484 is solved, this can be `functools.partial`
+def yaml_dump(o: Any, *, default_style: Optional[str] = None) -> str:
```
asottile, line 39: oh, can this just be `**kwargs: Any`?

marcjay: Makes more sense!

pre_commit/util.py
```diff
@@ -36,10 +36,10 @@
 Dumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
 
 
-def yaml_dump(o: Any) -> str:
-    # when python/mypy#1484 is solved, this can be `functools.partial`
```
asottile, line 40: oh, I think the comment is valid again, I'll just fix it when I merge it don't worry

marcjay:
Thanks for your time in improving the PR and cheers for the merge 👍

asottile:
thanks again for the issue and patch, this has been released as part of v2.4.0 🎉

PR #1437, opened 2020-05-05
Add Changelog project URL

scop:
Background info at
  https://github.com/pypa/warehouse/pull/7882#issue-412444446

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,8 @@ classifiers =
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: Implementation :: CPython
     Programming Language :: Python :: Implementation :: PyPy
+project_urls =
+    Changelog = https://github.com/pre-commit/pre-commit/blob/master/CHANGELOG.md
 
 [options]
 packages = find:


asottile:
I'd rather send people through https://pre-commit.com or https://github.com/pre-commit/pre-commit

scop:
Hmph, unfortunate. But pre-commit.com would benefit from a direct link to the change log.

scop:
Also, in case you're not aware, the pre-commit PyPI listing already contains both links you mentioned. This is for benefit of people doing an upgrade. It's great that there is a well maintained change log, but it's a shame it's not as prominent as it could/should.

Also #2 FWIW the mentioned warehouse issue shows that the reception for this has so far been solely positive across the bunch of projects I've filed this for, except pre-commit(-hooks) and your comment in the flake8 PR.

asottile:
I'd be fine adding a link to pre-commit.com but it seems a little bit unnecessary to add this (and imo borderline spam to send PRs to a bunch of projects adding this).  We don't (for example) list separately the issues link or the PRs link (though those are also know by warehouse).  In my opinion, this is just more unnecessary configuration that people can already figure out by clicking through the links that are provided.

PR #1444, opened 2020-05-09
Expose verbosity to called hooks

ssbarnea:
Allows hooks to check if pre-commit was called with verbose mode
in order to change their behavior.

Fixes: 1199

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -347,6 +347,9 @@ def run(
     if args.checkout_type:
         environ['PRE_COMMIT_CHECKOUT_TYPE'] = args.checkout_type
 
+    if args.verbose:
+        environ['PRE_COMMIT_VERBOSE'] = '1'
+
     with contextlib.ExitStack() as exit_stack:
         if stash:
             exit_stack.enter_context(staged_files_only(store.directory))


asottile:
needs a use case and a test -- why do you need this?

sfuerte:
> needs a use case and a test -- why do you need this?

here is a use-case for that: 
- using `pre-commit` in its normal, intended way - running local unit tests, SAST and linting on dev workstations,
- at the same time, have a shared GitHub Workflow, that is marked as "Required" in an org-wide ruleset to repos;
- the Workflow is simply re-running `pre-commit` checks on GitHub runners;
- so instead of creating multiple GitHub Actions and/or Workflow and have a high possibility of getting unsynchronized configs and/or exec args, the above makes sure that the checks and their configs are unified and in one place.

- now, to the problem and use-case, exposing verbosity flag to the hooks can make easier troubleshooting for the failed check(s) through and on top of GitHub's `IS_DEBUG` env var.

asottile:
> at the same time, have a shared GitHub Workflow, that is marked as "Required" in an org-wide ruleset to repos;

yeah you're holding it wrong.  the point is to always run exactly what you're running locally.  anything deviating from that you're doing yourself and your developers a disservice by making it *harder* to develop and reproduce.

verbosity has nothing to do with this self inflicted problem -- and you don't even need pre-commit to do anything for you at that point just set your own environment variable

PR #1467, opened 2020-05-24
Expose an env variable that indicates if pre-commit is running

### LINKED ISSUE #1426: Expose an env variable that indicates if pre-commit is running
In some circumstances, it could be helpful to know inside an executable / function / script if it was invoked from pre-commit or from some other process. I wrote a hook that just printed all env variables set but I could not find an env variable like `PRECOMMIT_RUNNING` or similar. Would you consider setting such an env variable during running pre-commit? I searched old issues and documentation but I could not find anything.

tech-chad:
Resolves  #1426


asottile:
thanks!

PR #1484, opened 2020-06-02
Detect rootless mode

### LINKED ISSUE #1243: Permission issue with rootless containers
I tried to use a `docker_image` hook on a RHEL7.7 system using `podman` with the `podman-docker` package installed [This setup allows for rootless containers].  The hook attempts to modify the file, but gets a "permission denied" error.  From looking at the source code, I see that `pre-commit` is roughly trying to execute:

```
docker run -u $(id -u):$(id -g)  -v $(pwd):/src:rw,Z --workdir /src -it <ENTRY> <FILE>
```
The hook does try to run, but results in permission error.
image

If, however, I remove the `-u` option from the source code (locally, languages/docker.py, docker_cmd()), then the hook runs fine:
image

themr0c:
fix #1243 - the -u option is not necessary on podman

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -78,23 +78,28 @@ def install_environment(
         os.mkdir(directory)
 
 
-@functools.lru_cache(maxsize=1)
-def docker_is_rootless() -> bool:
+def _docker_is_rootless() -> bool:
     returncode, stdout, stderr = cmd_output(
-        ('docker', 'system', 'info'),
+        'docker', 'system', 'info',
     )
     for line in stdout.splitlines():
         # rootless docker has "rootless"
         # rootless podman has "rootless: true"
         if line.strip().startswith('rootless'):
             if 'false' not in line:
+                print(line)
                 return True
             break
     return False
 
 
+@functools.lru_cache(maxsize=1)
+def docker_is_rootless() -> bool:
+    return _docker_is_rootless()
+
+
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
-    if: docker_is_rootless():
+    if docker_is_rootless():
         return ()
     try:
         return ('-u', f'{os.getuid()}:{os.getgid()}')
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -78,11 +78,11 @@ def install_environment(
 
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
+    output = subprocess.check_output(
+        ('docker', 'system', 'info'),
+        text=True,
+    )
     try:
-        output = subprocess.check_output(
-            ('docker', 'system', 'info'),
-            text=True,
-        )
         for line in output.splitlines():
             # rootless docker has "rootless"
             # rootless podman has "rootless: true"
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -78,7 +78,7 @@ def install_environment(
         os.mkdir(directory)
 
 
-def _docker_is_rootless() -> bool:
+def _docker_is_rootless() -> bool:  # pragma: win32 no cover
     returncode, stdout, stderr = cmd_output(
         'docker', 'system', 'info',
     )
@@ -93,7 +93,7 @@ def _docker_is_rootless() -> bool:
 
 
 @functools.lru_cache(maxsize=1)
-def docker_is_rootless() -> bool:
+def docker_is_rootless() -> bool:  # pragma: win32 no cover
     return _docker_is_rootless()
 
 
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -79,11 +79,19 @@ def install_environment(
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
     try:
-        ver_cmd = 'docker', '--version'
-        if subprocess.check_output(ver_cmd).startswith(b'podman version '):
-            return ()
-        else:
+        rootless = False
+        output = subprocess.check_output(('docker', 'system', 'info'), text=True)
+        for line in output.splitlines():
+            # rootless docker has "rootless"
+            # rootless podman has "rootless: true"
+            if line.strip().startswith('rootless'):
+                if not 'false' in line:
+                    rootless = True
+                break
+        if not rootless:
             return ('-u', f'{os.getuid()}:{os.getgid()}')
+        else:
+            return()
     except AttributeError:
         return ()
 
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -87,7 +87,6 @@ def _docker_is_rootless() -> bool:
         # rootless podman has "rootless: true"
         if line.strip().startswith('rootless'):
             if 'false' not in line:
-                print(line)
                 return True
             break
     return False
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -78,22 +78,15 @@ def install_environment(
 
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
-    try:
-        rootless = False
-        output = subprocess.check_output(('docker', 'system', 'info'), text=True)
-        for line in output.splitlines():
-            # rootless docker has "rootless"
-            # rootless podman has "rootless: true"
-            if line.strip().startswith('rootless'):
-                if not 'false' in line:
-                    rootless = True
-                break
-        if not rootless:
-            return ('-u', f'{os.getuid()}:{os.getgid()}')
-        else:
-            return()
-    except AttributeError:
-        return ()
+    output = subprocess.check_output(('docker', 'system', 'info'), text=True)
+    for line in output.splitlines():
+        # rootless docker has "rootless"
+        # rootless podman has "rootless: true"
+        if line.strip().startswith('rootless'):
+            if not 'false' in line:
+                return ()  # no -u for rootless
+            break
+    return ('-u', f'{os.getuid()}:{os.getgid()}')
 
 
 def docker_cmd() -> Tuple[str, ...]:  # pragma: win32 no cover
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -78,15 +78,21 @@ def install_environment(
 
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
-    output = subprocess.check_output(('docker', 'system', 'info'), text=True)
-    for line in output.splitlines():
-        # rootless docker has "rootless"
-        # rootless podman has "rootless: true"
-        if line.strip().startswith('rootless'):
-            if not 'false' in line:
-                return ()  # no -u for rootless
-            break
-    return ('-u', f'{os.getuid()}:{os.getgid()}')
+    try:
+        output = subprocess.check_output(
+            ('docker', 'system', 'info'),
+            text=True,
+        )
+        for line in output.splitlines():
+            # rootless docker has "rootless"
+            # rootless podman has "rootless: true"
+            if line.strip().startswith('rootless'):
+                if 'false' not in line:
+                    return ()  # no -u for rootless
+                break
+        return ('-u', f'{os.getuid()}:{os.getgid()}')
+    except AttributeError:
+        return ()
 
 
 def docker_cmd() -> Tuple[str, ...]:  # pragma: win32 no cover
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -1,5 +1,6 @@
 import hashlib
 import os
+import subprocess
 from typing import Sequence
 from typing import Tuple
 
@@ -78,7 +79,11 @@ def install_environment(
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
     try:
-        return ('-u', f'{os.getuid()}:{os.getgid()}')
+        ver_cmd = 'docker', '--version'
+        if subprocess.check_output(ver_cmd).startswith(b'podman version '):
+            return ()
+        else:
+            return ('-u', f'{os.getuid()}:{os.getgid()}')
     except AttributeError:
         return ()
 
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -82,14 +82,14 @@ def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
         ('docker', 'system', 'info'),
         text=True,
     )
+    for line in output.splitlines():
+        # rootless docker has "rootless"
+        # rootless podman has "rootless: true"
+        if line.strip().startswith('rootless'):
+            if 'false' not in line:
+                return ()  # no -u for rootless
+            break
     try:
-        for line in output.splitlines():
-            # rootless docker has "rootless"
-            # rootless podman has "rootless: true"
-            if line.strip().startswith('rootless'):
-                if 'false' not in line:
-                    return ()  # no -u for rootless
-                break
         return ('-u', f'{os.getuid()}:{os.getgid()}')
     except AttributeError:
         return ()
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -1,6 +1,6 @@
+import functools
 import hashlib
 import os
-import subprocess
 from typing import Sequence
 from typing import Tuple
 
@@ -10,6 +10,7 @@ from pre_commit.languages import helpers
 from pre_commit.prefix import Prefix
 from pre_commit.util import CalledProcessError
 from pre_commit.util import clean_path_on_failure
+from pre_commit.util import cmd_output
 from pre_commit.util import cmd_output_b
 
 ENVIRONMENT_DIR = 'docker'
@@ -77,18 +78,24 @@ def install_environment(
         os.mkdir(directory)
 
 
-def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
-    output = subprocess.check_output(
+@functools.lru_cache(maxsize=1)
+def docker_is_rootless() -> bool:
+    returncode, stdout, stderr = cmd_output(
         ('docker', 'system', 'info'),
-        text=True,
     )
-    for line in output.splitlines():
+    for line in stdout.splitlines():
         # rootless docker has "rootless"
         # rootless podman has "rootless: true"
         if line.strip().startswith('rootless'):
             if 'false' not in line:
-                return ()  # no -u for rootless
+                return True
             break
+    return False
+
+
+def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
+    if: docker_is_rootless():
+        return ()
     try:
         return ('-u', f'{os.getuid()}:{os.getgid()}')
     except AttributeError:
--- a/tests/languages/docker_test.py
+++ b/tests/languages/docker_test.py
@@ -28,9 +28,9 @@ Server:
  Runtimes: runc
  Default Runtime: runc
  Init Binary: /usr/libexec/docker/docker-init
- containerd version: 
+ containerd version:
  runc version: fbdbaf85ecbc0e077f336c03062710435607dbf1
- init version: 
+ init version:
  Security Options:
   seccomp
    Profile: default
@@ -51,7 +51,7 @@ Server:
  Insecure Registries:
   127.0.0.0/8
  Live Restore Enabled: true
-'''
+'''  # noqa
 
 DOCKER_ROOTLESS_SYSTEM_INFO = '''
 Client:
@@ -99,7 +99,7 @@ Server:
   127.0.0.0/8
  Live Restore Enabled: false
  Product License: Community Engine
-'''
+'''  # noqa
 
 PODMAN_SYSTEM_INFO = '''
 host:
@@ -183,7 +183,9 @@ store:
     number: 23
   runRoot: /tmp/105971
   volumePath: /home/remote/user/.local/share/containers/storage/volumes
-'''
+'''  # noqa
+
+
 def test_docker_is_running_process_error():
     with mock.patch(
         'pre_commit.languages.docker.cmd_output_b',
@@ -204,16 +206,24 @@ def test_docker_fallback_user():
 
 
 def test_docker_is_not_rootless():
-    with mock.patch.object(docker, 'cmd_output', return_value=(0, DOCKER_NOT_ROOTLESS_SYSTEM_INFO, '')):
-        assert docker.docker_is_rootless() == False
+    with mock.patch.object(
+        docker, 'cmd_output',
+        return_value=(0, DOCKER_NOT_ROOTLESS_SYSTEM_INFO, ''),
+    ):
+        assert docker._docker_is_rootless() is False
 
 
 def test_docker_is_rootless():
-    with mock.patch.object(docker, 'cmd_output', return_value=(0, DOCKER_ROOTLESS_SYSTEM_INFO, '')):
-        assert docker.docker_is_rootless() == True
+    with mock.patch.object(
+        docker, 'cmd_output',
+        return_value=(0, DOCKER_ROOTLESS_SYSTEM_INFO, ''),
+    ):
+        assert docker._docker_is_rootless() is True
 
 
 def test_podman_is_rootless():
-    with mock.patch.object(docker, 'cmd_output', return_value=(0, PODMAN_SYSTEM_INFO, '')):
-        assert docker.docker_is_rootless() == True
-
+    with mock.patch.object(
+        docker, 'cmd_output',
+        return_value=(0, PODMAN_SYSTEM_INFO, ''),
+    ):
+        assert docker._docker_is_rootless() is True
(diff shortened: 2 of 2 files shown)


themr0c:
It's my first attempt to write a python patch ever. I apologize if I did it wrong.

themr0c:
Build failing on coverage, I am afraid it is beyond my knowledge to fix this :/.

pre_commit/languages/docker.py
```diff
@@ -78,7 +79,11 @@ def install_environment(
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
     try:
-        return ('-u', f'{os.getuid()}:{os.getgid()}')
+        ver_cmd = 'docker', '--version'
+        if subprocess.check_output(ver_cmd).startswith(b'podman version '):
```
asottile, line 83: I'm still fairly certain this is a _bug_ in podman -- also we can't incur this overhead repeatedly so this patch as written won't be accepted

at most `('docker', '--version')` should be called once per `pre-commit` run (instead of every call) -- you can use `@functools.lru_cache(maxsize=1)` to do something like that

there's also a `cmd_output` helper which does some improved things over `subprocess.check_call` which is preferable here

the coverage failure is because your new code is untested

is there a link to the podman issue?

themr0c: Sorry @asottile, I am not a developer, I reached my limits of copy-paste programming. I am trying my best. This not working is currently a blocker to start using the tool on some project, but I have only tech writer skills.

The podman issue is there: https://github.com/containers/libpod/issues/6431 
The -u option exists in podman, but in my case, the UID of my user is too high, so that change of UID won't work.

The whole story is, there is no point using this `-u` option when using podman, because it is the default behaviour of podman to map to the current user. As I understand it, this `-u` option is used here to overcome the default behaviour of docker which is not doing this mapping. Therefore, it makes more sense not adding this option when running with podman, than trying to make it work for no benefits.

ssbarnea: With the risk of upsetting @themr0c I have support @asottile here, that is not a bug in pre-commit itself, more of a side effect of faking `docker` cli with another container engine.

Still, reading this change I realized that the way is currently implemented this code will fail to run with remote docker hosts, where all execution runs on another host. In that case there is no mapping to perform because a  volume mount is not even possible.

AFAIK, the only way to make this work in a portable way would be to build the source code inside the container and run it. An approach without any mounts would also avoid this issue, and likely work with podman too.

themr0c: No worries for upsetting me @ssbarnea :) What I am looking for is a solution, not my solution :) 

As a Fedora 32 user, to use pre-commit I need to make pre-commit and podman interact in a friendly way. All what I have now the the `podman-docker` package. On Fedora 32 installing docker is not an option. See https://docs.docker.com/engine/install/fedora/. 

As an user with id 105971, I would also like the -u option behave friendly with me of course. Given that comment: https://github.com/containers/libpod/issues/6431#issuecomment-636119379, I would like to reproduce on docker, but ...

Given your last remark, I wonder how pre-commit will run on kubernetes. It is our next step:  we need to make it work on https://che.openshift.io.

themr0c: New commit attempting to detect a rootless environment. Is that better?

pre_commit/languages/docker.py
```diff
@@ -78,6 +79,17 @@ def install_environment(
 
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
     try:
+        output = subprocess.check_output(
```
webknjaz, line 82: This new block shouldn't be inside try/except

themr0c: Where should I put it to avoid the `AttributeError` error like: https://asottile.visualstudio.com/asottile/_build/results?buildId=3829&view=logs&j=ae542f9d-96d9-59d2-9a06-a2a85a17a811&t=6deecb4d-4f67-5e8d-1829-1f5cd5833176

pre_commit/languages/docker.py
webknjaz: @asottile is this cacheable?

pre_commit/languages/docker.py
webknjaz: That is a part of the test. The code expects `os.getuid()` or `os.getgid()` raise that exception and you make it expect that your new code may also raise that exception which is not true. But even if it was, you shouldn't put more than one line inside of a try/except block because the expectations would be too broad.

themr0c: OMG! I did that?

webknjaz: Most likely. Just move it outside, before `try`.

themr0c: Done this. If I manage to get something useful out of this PR, I will owe some beers to half of the python planet ...

themr0c:
Now I am facing a coverage error: https://asottile.visualstudio.com/asottile/_build/results?buildId=3831&view=logs&j=291e3f77-befc-520d-9779-f5b46c027190&t=037f03e6-93d9-56f0-123e-44a4e724aa8c&l=73

I am clueless on how to fix it:

```
coverage report
Name                             Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------
pre_commit/languages/docker.py      64      3      8      1    92%   89->90, 90-92
```

pre_commit/languages/docker.py
```diff
@@ -77,7 +78,18 @@ def install_environment(
...
+        text=True,
+    )
     try:
+        for line in output.splitlines():
+            # rootless docker has "rootless"
+            # rootless podman has "rootless: true"
+            if line.strip().startswith('rootless'):
+                if 'false' not in line:
+                    return ()  # no -u for rootless
+                break
```
webknjaz, line 92: Not just output, move this whole thing out.

themr0c: Done. Thanks @webknjaz!

themr0c:
I spent some time trying to understand how to write a test to make the build pass. I just don't understand how to get it. Too much to learn in one go. May I request some help (again)?

webknjaz:
@themr0c you can use a built-in fixture called [`monkeypatch`](https://docs.pytest.org/en/latest/monkeypatch.html) (for this, just add `monkeypatch` as an argument of your test method) to mock `subprocess.check_output` with a fake function that returns a sample output of podman (multiline).

The simple one case test would be
```
_ROOTLESS_DOCKER_OUTPUT = """
some garbage
   rootless: true
      even more garbage
"""

def test_rootless_docker(monkeypatch):
    monkeypatch.setattr('subprocess.check_output', lambda cmd, text: _ROOTLESS_DOCKER_OUTPUT)
    assert get_docker_user() == ()
```

webknjaz:
```python
import os

import pytest

_CURRENT_UID = os.getuid()
_CURRENT_GID = os.getgid()

_ROOTLESS_DOCKER_OUTPUT = """
some garbage
   rootless: true
      even more garbage
"""

@pytest.mark.parametrize(
    ('docker_sys_info', 'expected_args'),
    (
        pytest.param(
            """
            some garbage
               rootless: true
                  even more garbage
            """,
            (),
            id='rootless podman',
        ),
        pytest.param(
            """
            some garbage
               rootless: false
                  even more garbage
            """,
            (_CURRENT_UID, _CURRENT_GID),
            id='non-rootless podman',
        ),
        pytest.param(
            """
            some garbage
               rootless
                  even more garbage
            """,
            (),
            id='rootless docker',
        ),
        pytest.param(
            """
            some garbage
               nothing good
                  even more garbage
            """,
            (_CURRENT_UID, _CURRENT_GID),
            id='non-rootless docker',
        ),
    ),
)
def test_rootless_docker(docker_sys_info, expected_args, monkeypatch):
    """Verify that only rootless docker/podman doesn't add args."""
    monkeypatch.setattr('subprocess.check_output', lambda cmd, text: docker_sys_info)
    assert docker.get_docker_user() == expected_args
```

themr0c:
I am completely illiterate in writing python code, it seems I chose a task  too complex for me as first attempt to contribute :/.

* Is there some example somewhere where I could learn to use `@functools.lru_cache(maxsize=1)` by copy-pasting?
* What does the replacement helper for `subprocess.check_output` look like?
* I have read the other tests ans seen  some `unittest.mock.patch.object`. I still don't understand how to write a proper test.

webknjaz:
@themr0c you literally can just `import functools` and paste `@functools.lru_cache(maxsize=1)` right before `def get_docker_user(): ...` (in the previous line). LRU cache only works during the same runtime session. So if you run the program (pre-commit) again, it'll not hit the cache and execute the function. I suppose Anthony wants the cached result to be preserved across pre-commit runs which is more complicated. Also, I'd factor out the logic of identifying the rootless install into a separate helper function and apply caching just to that function.

asottile:
> 
> 
> @themr0c you literally can just `import functools` and paste `@functools.lru_cache(maxsize=1)` right before `def get_docker_user(): ...` (in the previous line). LRU cache only works during the same runtime session. So if you run the program (pre-commit) again, it'll not hit the cache and execute the function. I suppose Anthony wants the cached result to be preserved across pre-commit runs which is more complicated. Also, I'd factor out the logic of identifying the rootless install into a separate helper function and apply caching just to that function.

in-process cache is fine -- there's currently no precedent for across-process caching in pre-commit, though I plan to do that eventually for virtualenv invalidation

asottile:
> I am completely illiterate in writing python code, it seems I chose a task too complex for me as first attempt to contribute :/.

you're doing great! we'll help you through it :)

>     * Is there some example somewhere where I could learn to use `@functools.lru_cache(maxsize=1)` by copy-pasting?
> 

in this case it would just be importing `functools`  and adding that decorator to the function that's being modified here

>     * What does the replacement helper for `subprocess.check_output` look like?

the helper is `cmd_output` and you can use it like `cmd_output('docker', '--version')` (it returns `(retcode, stdout, stderr)`)

>     * I have read the other tests ans seen  some `unittest.mock.patch.object`. I still don't understand how to write a proper test.

taking the test above, you'd use

```python
with mock.patch.object(docker, 'cmd_output', return_value=(0, ..., '')):
    ...
```

instead of the `monkeypatch` of `check_output`

hroncok:
Side note: We have a videocall scheduled with @themr0c for early next week (this week it didn't work out) and we'll go trough the requested changes together.

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,24 @@ def install_environment(
...
+    returncode, stdout, stderr = cmd_output(
+        ('docker', 'system', 'info'),
+    )
+    for line in stdout.splitlines():
+        # rootless docker has "rootless"
+        # rootless podman has "rootless: true"
+        if line.strip().startswith('rootless'):
+            if 'false' not in line:
+                return True
+            break
```
webknjaz, line 92: this is probably slightly more elegant:
```suggestion
        if not line.strip().startswith('rootless'):
            continue
        return 'false' not in line
```
(though, you don't have to apply it if you prefer the current solution)

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,24 @@ def install_environment(
...
+    return False
+
+
 def get_docker_user() -> Tuple[str, ...]:  # pragma: win32 no cover
+    if: docker_is_rootless():
```
webknjaz: This small typo makes the CI explode:
```suggestion
    if docker_is_rootless():
```
(just accepting it via GH UI should fix that)

tests/languages/docker_test.py
```diff
@@ -21,3 +201,19 @@ def invalid_attribute():
         getgid=invalid_attribute,
     ):
         assert docker.get_docker_user() == ()
+
+
+def test_docker_is_not_rootless():
+    with mock.patch.object(docker, 'cmd_output', return_value=(0, DOCKER_NOT_ROOTLESS_SYSTEM_INFO, '')):
+        assert docker.docker_is_rootless() == False
```
webknjaz: The idiomatic way is to do
```suggestion
        assert not docker.docker_is_rootless()
```

webknjaz: Or, at least
```suggestion
        assert docker.docker_is_rootless() is False
```

pre_commit/languages/docker.py
themr0c: done

tests/languages/docker_test.py
webknjaz: (because one shouldn't use `==` comparisons with `True`, `False` and `None` — `is` should be used instead)

tests/languages/docker_test.py
```diff
@@ -21,3 +201,19 @@ def invalid_attribute():
...
+
+
+def test_docker_is_rootless():
+    with mock.patch.object(docker, 'cmd_output', return_value=(0, DOCKER_ROOTLESS_SYSTEM_INFO, '')):
+        assert docker.docker_is_rootless() == True
```
webknjaz:
```suggestion
        assert docker.docker_is_rootless()
```

tests/languages/docker_test.py
```diff
@@ -21,3 +201,19 @@ def invalid_attribute():
...
+
+
+def test_podman_is_rootless():
+    with mock.patch.object(docker, 'cmd_output', return_value=(0, PODMAN_SYSTEM_INFO, '')):
+        assert docker.docker_is_rootless() == True
```
webknjaz:
```suggestion
        assert docker.docker_is_rootless()
```

tests/languages/docker_test.py
hroncok: Yes, we were doing this later...

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,29 @@ def install_environment(
...
+def _docker_is_rootless() -> bool:
+    returncode, stdout, stderr = cmd_output(
+        'docker', 'system', 'info',
+    )
+    for line in stdout.splitlines():
+        # rootless docker has "rootless"
+        # rootless podman has "rootless: true"
+        if line.strip().startswith('rootless'):
+            if 'false' not in line:
+                print(line)
```
webknjaz, line 90: `print()` shouldn't be used for logging (unless you're debugging something right now)

pre_commit/languages/docker.py
hroncok: IMO I found this harder to understand.

pre_commit/languages/docker.py
hroncok: this was debugging.

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,28 @@ def install_environment(
...
+        if line.strip().startswith('rootless'):
+            if 'false' not in line:
+                return True
+            break
+    return False
+
+
+@functools.lru_cache(maxsize=1)
+def docker_is_rootless() -> bool:
+    return _docker_is_rootless()
```
hroncok, line 97: Note: We had to do this to be able to test this with multiple mocked outputs. The cached version made that very hard.

pre_commit/languages/docker.py
themr0c: OK, let's keep the first one then.

pre_commit/languages/docker.py
webknjaz: could also be something like `docker_is_rootless = functools.lru_cache(maxsize=1)(_docker_is_rootless)`

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,28 @@ def install_environment(
...
+        # rootless podman has "rootless: true"
+        if line.strip().startswith('rootless'):
+            if 'false' not in line:
+                return True
+            break
+    return False
+
+
+@functools.lru_cache(maxsize=1)
+def docker_is_rootless() -> bool:
```
webknjaz, line 96: This should probably have a PEP257-compliant docstring

pre_commit/languages/docker.py
hroncok: Could. Not is if more or less readable.

pre_commit/languages/docker.py
hroncok: To be fair, non of the other functions in this file have docstrings.

webknjaz: Fair enough. I guess it's just my personal habit kicking in before the linters :)

hroncok:
> I think you could also add a test calling docker.docker_is_rootless() twice and checking that cmd_output got called only once.

This seems a bit too far fetched to me. Isn't it rather testing lru_cache?

themr0c:
@asottile Is it satisfactory with the latest changes?
@hroncok A big thank you for the help!

webknjaz:
> This seems a bit too far fetched to me. Isn't it rather testing lru_cache?

It's a regression test: when somebody decides to remove the decorator, the test should explode.

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,28 @@ def install_environment(
...
 
+def _docker_is_rootless() -> bool:  # pragma: win32 no cover
+    returncode, stdout, stderr = cmd_output(
+        'docker', 'system', 'info',
+    )
+    for line in stdout.splitlines():
+        # rootless docker has "rootless"
+        # rootless podman has "rootless: true"
+        if line.strip().startswith('rootless'):
+            if 'false' not in line:
```
hroncok, line 89: Note: Technically, the case where `false` is present is not tested. However in practice, we haven't found a case like this, this check is present as precaution.

webknjaz: This would be useful as a code comment

themr0c: Is that what is causing the coverage error?

```
Name                             Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------
pre_commit/languages/docker.py      72      2     10      2    95%   89->91, 91, 101->102, 102
```

webknjaz: This is weird. If tests include example outputs containing `rootless`, they should also hit this condition check.

pre_commit/languages/docker.py
asottile: there's an example of testing the inner part of an `lru_cache` function in pre_commit/languages/node.py (tests/languages/node_test.py) -- basically, access the `__wrapped__` attribute of the cached object in the tests, then you don't need this indirection

pre_commit/languages/docker.py
webknjaz: @asottile any ideas about this?

webknjaz:
@themr0c I think this PR's title/description should be updated.

pre_commit/languages/docker.py
asottile: there's no test for `rootless: false` so line 90 is never hit -- perhaps simpler would be `if line.strip() == 'rootless: true': ...`

tests/languages/docker_test.py
```diff
@@ -3,6 +3,188 @@
...
+ ID: IUGE:4L5B:VVTV:JGIA:IIJN:G7MG:TVGF:XBVW:YHF7:MYRK:L524:4HBK
+ Docker Root Dir: /var/lib/docker
+ Debug Mode: false
+ Registry: https://index.docker.io/v1/
+ Labels:
+ Experimental: false
+ Insecure Registries:
+  127.0.0.0/8
+ Live Restore Enabled: true
+'''  # noqa
```
webknjaz, line 54: it's usually better to specify specific violation codes instead of ignoring everything

pre_commit/languages/docker.py
webknjaz: @themr0c @hroncok this will need to be addressed ^

pre_commit/languages/docker.py
webknjaz: That would probably not catch docker's behavior. Better to just improve the test matrix

asottile:
let me know if you'd like me to finish this one, I've finally gotten around to setting up podman to reproduce this

hroncok:
Depends on @themr0c. I am available to meet again and address the review comments, but if they prefer you to handle it, I don't mind.

themr0c:
@asottile @hroncok If you manage to finish this one, I would be very happy. I am not sure I would find time during the next coming month.

asottile:
looks like `--userns=keep-id` works slightly better on podman, no idea if this works well for docker though 🤔 https://stackoverflow.com/a/63767259/812183

themr0c:
The user option is not necessary, podman does the correct mapping.

```
$ ls -ltr toto
ls: cannot access 'toto': No such file or directory
$ podman  run --rm -ti   -v $(pwd):/docs:Z antora/antora touch /docs/toto
/docs/tox.ini-rw-rw-r--    1 root     root          7966 Sep 14 11:22 /docs/tox.ini
$ ls -ltr toto
-rw-r--r--. 1 ffloreth ffloreth 0 14 sep 14:36 toto
```

On the contrary, specifying user and userns can lead to trouble:

``` 
$ podman  run --rm -ti   --userns=keep-id --user $(id -u):$(id -g)   -v $(pwd):/docs:Z antora/antora ls -ltr /docs/tox.ini
Error: requested user's UID 105971 is too large for the rootless user namespace
```

One of the main advantages of podman is to get rid of the -u option that you need when you run docker if you don't want to see your workspaces filled by files owned by root... So why insist on keeping some unnecessary complexity?

asottile:
the default mapping gives too much permission and can create undeletable files outside:

```console
$ podman run --rm -ti -v $PWD:/z:rw ubuntu:focal bash -c 'mkdir -p /z/1/2/3 && chown -R nobody /z/1'
$ rm -rf 1
rm: cannot remove '1/2/3': Permission denied
```

themr0c:
Now I undertand better.

macoMarv86 (approved):
good for issue

PhilipTrauner:
> looks like `--userns=keep-id` works slightly better on podman, no idea if this works well for docker though 🤔 [stackoverflow.com/a/63767259/812183](https://stackoverflow.com/a/63767259/812183)

Came here from #2173 and would be interested in taking this over. 

Would it be sufficient to return `('-u', f'{os.getuid()}:{os.getgid()}', '--userns=keep-id')` in [`get_docker_user()`](https://github.com/themr0c/pre-commit/blob/9d4fe3a00b89bdbfce5708d4433c5d7c4f7aadb2/pre_commit/languages/docker.py#L100-L106) if (a) `docker` is an alias for `podman` and (b) rootless mode is detected?

Svgbotpay (approved):

themr0c:
@PhilipTrauner please take it over. I am happy if this pull request gets closed, even unresolved. I apologize for having started something that I am not able to finish.

pre_commit/languages/docker.py
```diff
@@ -76,7 +78,28 @@ def install_environment(
         os.mkdir(directory)
 
 
+def _docker_is_rootless() -> bool:  # pragma: win32 no cover
+    returncode, stdout, stderr = cmd_output(
+        'docker', 'system', 'info',
+    )
```
Kurt-von-Laven, line 84: It would be even faster and simplify both the logic below and the mock output used in testing to only query Docker/Podman for the field we care about. From [the rootless-docker GitHub Action's README](https://github.com/ScribeMD/rootless-docker#rootless-docker), `docker info --format "{{ .ClientInfo.Context }}"` outputs `rootless` in Docker when in rootless mode, and other strings (most commonly `default`, I believe) otherwise. I suspect, based on [Podman's documentation](https://docs.podman.io/en/latest/markdown/podman-info.1.html#example), that `podman info --format "{{ .host.security.rootless }}"` outputs `true` when in rootless mode and `false` otherwise. [Go Templates](https://pkg.go.dev/text/template#hdr-Actions) are quite powerful, so we can say: `docker info --format '{{ if .ClientInfo.Context }} {{ eq .ClientInfo.Context "rootless" }} {{ else }} {{ .host.security.rootless }} {{ end }}'`. This command outputs `true` when running in rootless mode and `false` otherwise on both Docker and Podman, although I have only tested it on Docker. I don't have Podman installed, so hopefully someone who does can straighten me out if I'm mistaken. Disclosure: I am the author of the rootless-docker GitHub Action.

alexvorndran: Here on my machine (Linux Mint 19.3, Docker version 20.10.16, build aa7e414) your suggestion returns `default` for rootless docker, as `rootless` seems to be part of the `Security Options` under `Server`.

Kurt-von-Laven: Thanks for letting me know. Bummer that the behavior isn't consistent across our machines even on the exact same build of Docker. I am on Ubuntu 22.04, but I wonder whether the difference has more to do with how we installed rootless Docker? It's possible to detect `rootless` mode under `SecurityOptions` too (e.g., `docker info --format '{{ range .SecurityOptions }}{{ if eq . "name=rootless" }}true{{ end }}{{ end }}'`), FWIW.

alexvorndran: IIRC I followed the [official instructions](https://docs.docker.com/engine/security/rootless) for Ubuntu. Maybe you chose to set the client context explicitly where I went for the `DOCKER_HOST` environment variable? Both options are listed in the instructions: https://docs.docker.com/engine/security/rootless/#client

Kurt-von-Laven: Yes, you are correct. I suppose one could check the `DOCKER_HOST` environment variable before running `docker info`.

themr0c:
Closing, as this pull request implements the feature: https://github.com/kubernetes/test-infra/pull/25841

PR #1506, opened 2020-06-12
Add three missing required dependencies in CONTRIBUTING.md

Pierre-Sassoulas:
There could still be missing dependencies, I'm not using a fresh environement
to test that. Also added the specific required git version see
https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.2.txt

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,12 +4,16 @@
 
 - The complete test suite depends on having at least the following installed
   (possibly not a complete list)
-  - git (A sufficiently newer version is required to run pre-push tests)
+  - git (Version 2.24.0 or above is required to run pre-merge-commit tests)
   - python2 (Required by a test which checks different python versions)
   - python3 (Required by a test which checks different python versions)
   - tox (or virtualenv)
   - ruby + gem
   - docker
+  - conda
+  - cargo (required by tests for rust dependencies)
+  - go (required by tests for go dependencies)
+  - swift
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,12 +4,16 @@
 
 - The complete test suite depends on having at least the following installed
   (possibly not a complete list)
-  - git (A sufficiently newer version is required to run pre-push tests)
+  - git (Version 1.8.2 or above is required to run pre-push tests)
```
asottile, line 7: there's also a version needed to run the `pre-merge-commit` tests

CONTRIBUTING.md
```diff
@@ -4,12 +4,16 @@
...
   - python2 (Required by a test which checks different python versions)
   - python3 (Required by a test which checks different python versions)
   - tox (or virtualenv)
   - ruby + gem
   - docker
+  - conda
+  - cargo (required by tests for rust dependencies)
+  - go (required by tests for go dependencies)
```
asottile, line 15: there's also a requirement for swift

CONTRIBUTING.md
Pierre-Sassoulas: OK, I changed this to 2.24.0.

CONTRIBUTING.md
Pierre-Sassoulas: Added in latest commit.

PR #1511, opened 2020-06-16
Using pathlib to handle Windows network mounts

RobT00:
Encountered issues when using pre-commit on mapped networks drives in Windows 10. Hopefully, this workaround will be useful

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import os
 import sys
+from pathlib import Path
 from typing import Any
 from typing import Optional
 from typing import Sequence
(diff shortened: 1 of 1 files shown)


asottile:
please do this without pathlib, thanks

asottile:
could you also demonstrate the specific problem you are encountering?

RobT00:
> could you also demonstrate the specific problem you are encountering?

I have detailed in #1530

RobT00:
> please do this without pathlib, thanks

Using `win32wnet.WNGetUniversalName` from [pywin32](https://pypi.org/project/pywin32/) also works, but I would guess it is undesirable to add this dependency?

PR #1539, opened 2020-07-21
Add --no-allow-missing-config option to init-templatedir

singergr:
The hook template created by `pre-commit init-templatedir` acts like it was installed by `pre-commit install --allow-missing-config`. That behavior makes sense for most people: if you're working with some repos that have pre-commit and some that don't, it's annoying to uninstall pre-commit from the repos that don't use it. However, for people trying to adopt a policy of using pre-commit in every repo, it's easier to catch repos that are missing the config if pre-commit fails hard.

To support this special case, add a `--no-allow-missing-config` option to the `init-templatedir` command. When that option is set, create a template hook that fails in repos where the pre-commit config is missing.

### DIFF
--- a/pre_commit/commands/init_templatedir.py
+++ b/pre_commit/commands/init_templatedir.py
@@ -15,10 +15,15 @@ def init_templatedir(
         store: Store,
         directory: str,
         hook_types: Sequence[str],
+        skip_on_missing_config: bool = True,
 ) -> int:
     install(
-        config_file, store, hook_types=hook_types,
-        overwrite=True, skip_on_missing_config=True, git_dir=directory,
+        config_file,
+        store,
+        hook_types=hook_types,
+        overwrite=True,
+        skip_on_missing_config=skip_on_missing_config,
+        git_dir=directory,
     )
     try:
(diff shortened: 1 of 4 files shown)


pre_commit/commands/init_templatedir.py
```diff
@@ -15,10 +15,21 @@ def init_templatedir(
         store: Store,
         directory: str,
         hook_types: Sequence[str],
+        skip_on_missing_config: bool = True,
 ) -> int:
+    if not skip_on_missing_config:
+        logger.warning(
+            'Missing pre-commit configs will NOT be allowed in newly cloned '
+            'repositories by default',
+        )
```
asottile, line 24: no need for a warning, they selected the option

asottile:
I'm fine adding the option, let's remove the warning though

singergr:
> I'm fine adding the option, let's remove the warning though

Thanks! It's too bad you figured out I'm actually a dog, though.

asottile:
> 
> 
> > I'm fine adding the option, let's remove the warning though
> 
> Thanks! It's too bad you figured out I'm actually a dog, though.

woof, woof woof!  WOOF woof!

Issue #1564, opened 2020-08-20
Idea: Improve hook speed by skipping before/after diff via readonly flag 


jhenkens:
https://github.com/pre-commit/pre-commit/blob/4f5cb99ff5c9e13edcd97461c09dcaf271197e5c/pre_commit/commands/run.py#L175

Hi all,

I am looking at implemented pre-commit to replace a harder-to-maintain custom bash script our company current uses. In my initial tests, pre-commit works great, except for the fact that it is slow. The majority of our hooks are readonly in nature, and run quite fast. However, on our repository, pre-commit is spending close to 1 second per hook performing a before/after diff.

I added some instrumentation, and in our test case with 10 executed hooks, the execution time of pre-commit is >7 seconds with the before/after diffs, and ~1 second with those diffs disabled.

In exploring issues related to perf, I had found https://github.com/pre-commit/pre-commit/issues/510, which seems to tackle a different problem - individual hooks that take a while, rather than the overhead of pre-commit itself. However, there was one comment that got me thinking, regarding a `readonly` attribute one could apply to a hook. I know - this metadata would only be as truthy as the author who wrote the hook, but if a hook was marked as readonly, `pre-commit` could theoretically skip the before/after hook diffs.

Any thoughts? Happy to contribute the PR if maintainers see this as valuable.

asottile:
this idea was contemplated at some point but was rejected as it would be difficult for two reasons:
- plugin-based hooks wouldn't ever be able to set this accurately
- if it was set innaccurately it would look like the framework's fault that it allowed through a commit
- in tests it really only made a difference on windows (and even then, not a significant difference)

one thing you could pursue is reducing the amount of git diffs by half (instead of before/after, carry over the diff from the previous hook run).  but even that wasn't pursued because it wasn't significant

jhenkens:
I'm still converting more hooks for our use case, so our runtime is only going to deteriorate further. I'd argue that, for our case, it is substantial. Our current (hard to maintain and grok) hooks execute in the 1-2 second range.

I am running all tests on a Mac (though much of our development team is on Windows, so if it is even slower there than that is a compounding problem).

> plugin-based hooks wouldn't ever be able to set this accurately
[...]

PR #1566, opened 2020-08-20
Save diff between hook executions

jhenkens:
This reduces the number of calls to `git diff` by about half, as discussed in https://github.com/pre-commit/pre-commit/issues/1564

Validated locally on a large repository.

```
Overall duration before: 7.93s
Overall duration after: 5.01s
```


pre_commit/commands/run.py
```diff
@@ -261,14 +275,17 @@ def _run_hooks(
         _all_filenames(args), config['files'], config['exclude'],
     )
     retval = 0
+    prior_diff: Optional[Tuple[int, bytes, Optional[bytes]]] = None
```
asottile, line 278: instead of optional, initialize this to the output of `git diff`?

pre_commit/commands/run.py
```diff
@@ -261,14 +275,17 @@ def _run_hooks(
...
+        current_retval, current_diff = _run_single_hook(
+            classifier, hook, skips, cols, prior_diff,
             verbose=args.verbose, use_color=args.color,
         )
+        retval |= current_retval
+        prior_diff = current_diff
         if retval and config['fail_fast']:
             break
-    if retval and args.show_diff_on_failure and git.has_diff():
+    if retval and args.show_diff_on_failure and _has_diff(prior_diff):
```
asottile, line 288: with the suggestion above, this can become `and prior_diff` -- though ~generally this isn't in the critical path for committing (since `--show-diff-on-failure` is not accessible there) so maybe this change isn't necessary

also if this is the last caller of `git.has_diff()` can you delete that code?

pre_commit/commands/run.py
```diff
@@ -248,6 +253,15 @@ def _all_filenames(args: argparse.Namespace) -> Collection[str]:
         return git.get_staged_files()
 
 
+def _has_diff(
+        diff: Optional[Tuple[int, bytes, Optional[bytes]]],
+) -> bool:
+    if diff is not None:
+        return diff[0] == 1
```
asottile, line 260: oh hmmm, I don't think we need to keep track of the returncode at all, that seems to be an accidental implementation detail of the current code

pre_commit/commands/run.py
```diff
@@ -134,9 +135,10 @@ def _run_single_hook(
         hook: Hook,
         skips: Set[str],
         cols: int,
+        diff_before: Optional[Tuple[int, bytes, Optional[bytes]]],
         verbose: bool,
         use_color: bool,
-) -> bool:
+) -> Tuple[bool, Optional[Tuple[int, bytes, Optional[bytes]]]]:
```
asottile, line 141: after the adjustment this can be `Tuple[bool, bytes]`

pre_commit/commands/run.py
jhenkens: Not the last user - https://github.com/pre-commit/pre-commit/blob/master/pre_commit/commands/try_repo.py#L28

PR #1598, opened 2020-09-12
add initial dotnet support

rkm:
Hello hello.

This PR adds [2nd class](https://github.com/pre-commit/pre-commit/blob/master/CONTRIBUTING.md#adding-support-for-a-new-hook-language) support for dotnet [tools](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create) by using `dotnet pack` and `dotnet tool install`.

I've also tested a "real world" hook:

<details>

```console
$ cd /tmp
$ rm -rf ./foo
$ dotnet new console --name foo
The template "Console Application" was created successfully.

Processing post-creation actions...
Running 'dotnet restore' on foo/foo.csproj...
  Determining projects to restore...
  Restored /tmp/foo/foo.csproj (in 190 ms).

Restore succeeded.

$ cd foo
$ git init
Initialized empty Git repository in /tmp/foo/.git/
$ sed -i 's/Console/    Console/g' Program.cs
$ pre-commit try-repo https://github.com/rkm/format --ref af89478 --files=foo.csproj
[INFO] Initializing environment for https://github.com/rkm/format.
===============================================================================
Using config:
===============================================================================
repos:
-   repo: https://github.com/rkm/format
    rev: af89478
    hooks:
    -   id: dotnet-format
===============================================================================
[INFO] Installing environment for https://github.com/rkm/format.
dotnet format............................................................Failed
- hook id: dotnet-format
- files were modified by this hook

  Formatting code files in workspace '/tmp/foo/foo.csproj'.
  Warnings were encountered while loading the workspace. Set the verbosity option to the 'diagnostic' level to log warnings.
  Formatted code file '/tmp/foo/Program.cs'.
  Format complete in 3530ms.
```
</details>

One of the issues I found was that nuget doesn't play nicely when some env vars are not set. In particular the Windows build was failing until I added the tox.ini changes below. If you don't want to carry that addition, then I can probably investigate a workaround using `get_env_patch`.

Looking forward to your review!

### DIFF
--- a/pre_commit/languages/dotnet.py
+++ b/pre_commit/languages/dotnet.py
@@ -1,0 +1,90 @@
+import contextlib
+import os.path
+from typing import Generator
+from typing import Sequence
+from typing import Tuple
+
+import pre_commit.constants as C
+from pre_commit.envcontext import envcontext
+from pre_commit.envcontext import PatchesT
+from pre_commit.envcontext import Var
+from pre_commit.hook import Hook
+from pre_commit.languages import helpers
+from pre_commit.prefix import Prefix
+from pre_commit.util import clean_path_on_failure
+from pre_commit.util import rmtree
+
+ENVIRONMENT_DIR = 'dotnetenv'
+BIN_DIR = 'bin'
+
+get_default_version = helpers.basic_get_default_version
+healthy = helpers.basic_healthy
+
+
+def get_env_patch(venv: str) -> PatchesT:
+    return (
+        ('PATH', (os.path.join(venv, BIN_DIR), os.pathsep, Var('PATH'))),
+    )
+
+
+@contextlib.contextmanager
+def in_env(prefix: Prefix) -> Generator[None, None, None]:
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, C.DEFAULT)
+    envdir = prefix.path(directory)
+    with envcontext(get_env_patch(envdir)):
+        yield
+
+
+def install_environment(
+        prefix: Prefix,
+        version: str,
+        additional_dependencies: Sequence[str],
+) -> None:
+    helpers.assert_version_default('dotnet', version)
+    helpers.assert_no_additional_deps('dotnet', additional_dependencies)
+
+    envdir = prefix.path(helpers.environment_dir(ENVIRONMENT_DIR, version))
+    with clean_path_on_failure(envdir):
+        build_dir = 'pre-commit-build'
+
+        # Build & pack nupkg file
+        helpers.run_setup_cmd(
+            prefix,
+            (
+                'dotnet', 'pack',
+                '--configuration', 'Release',
+                '--output', build_dir,
+            ),
+        )
+
+        # Determine tool from the packaged file <tool_name>.<version>.nupkg
+        build_outputs = os.listdir(os.path.join(prefix.prefix_dir, build_dir))
+        if len(build_outputs) != 1:
+            raise NotImplementedError(
+                f"Can't handle multiple build outputs. Got {build_outputs}",
+            )
+        tool_name = build_outputs[0].split('.')[0]
+
+        # Install to bin dir
+        helpers.run_setup_cmd(
+            prefix,
+            (
+                'dotnet', 'tool', 'install',
+                '--tool-path', os.path.join(envdir, BIN_DIR),
+                '--add-source', build_dir,
+                tool_name,
+            ),
+        )
+
+        # Cleanup build output
+        for d in ('bin', 'obj', build_dir):
+            rmtree(prefix.path(d))
+
+
+def run_hook(
+        hook: Hook,
+        file_args: Sequence[str],
+        color: bool,
+) -> Tuple[int, bytes]:
+    with in_env(hook.prefix):
+        return helpers.run_xargs(hook, hook.cmd, file_args, color=color)
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist = py36,py37,py38,pypy3,pre-commit
 
 [testenv]
 deps = -rrequirements-dev.txt
-passenv = HOME LOCALAPPDATA RUSTUP_HOME
+passenv = APPDATA HOME LOCALAPPDATA PROGRAMFILES RUSTUP_HOME
 commands =
     coverage erase
     coverage run -m pytest {posargs:tests}
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /.tox
 /dist
 /venv*
+.vscode/
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -8,6 +8,7 @@ from pre_commit.hook import Hook
 from pre_commit.languages import conda
 from pre_commit.languages import docker
 from pre_commit.languages import docker_image
+from pre_commit.languages import dotnet
 from pre_commit.languages import fail
 from pre_commit.languages import golang
 from pre_commit.languages import node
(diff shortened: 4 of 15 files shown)


pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,108 @@
...
+
+get_default_version = helpers.basic_get_default_version
+healthy = helpers.basic_healthy
+
+
+def get_env_patch(venv: str) -> PatchesT:
+
+    # https://github.com/dotnet/sdk/issues/9304
+    dotnet = shutil.which('dotnet')
+    assert dotnet, 'No dotnet executable found - please install one'
```
asottile, line 31: if I'm reading this correctly, we shouldn't need to do this as the host machine should have that set up properly already -- I think all that needs to happen here is a `passenv` in `tox.ini`?

(seems silly that they didn't just default it to the location where `argv[0]` is, but I didn't have time to read the whole dotnet issue!)

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,108 @@
...
+    with clean_path_on_failure(repodir):
+
+        build_dir = 'pre-commit-build'
+
+        # Build & pack nupkg file
+        helpers.run_setup_cmd(
+            prefix,
+            (
+                'dotnet', 'pack',
+                '--configuration', BUILD_CONFIG,
```
asottile, line 69: can just inline this constant

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,108 @@
...
+) -> None:
+
+    helpers.assert_version_default('dotnet', version)
+    helpers.assert_no_additional_deps('dotnet', additional_dependencies)
+
+    envdir = helpers.environment_dir(ENVIRONMENT_DIR, version)
+    repodir = prefix.path(envdir)
+
+    with clean_path_on_failure(repodir):
+
```
asottile, line 61: (I should really just write a formatter for this) -- generally blank lines at the top of blocks make it harder to read, I can go through and remove these if you want

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,108 @@
...
+                '--configuration', BUILD_CONFIG,
+                '--output', build_dir,
+            ),
+        )
+
+        # Determine the tool name
+        build_outputs = os.listdir(os.path.join(prefix.prefix_dir, build_dir))
+        assert (
+            len(build_outputs) == 1
+        ), f"Can't handle multiple build outputs. Got {build_outputs}"
```
asottile, line 78: let's change this into an if statement with a raise, that'll make this always work even if they (for some reason) set `-0`

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+        # Cleanup
+        for d in glob.glob(f'{prefix.prefix_dir}/*'):
+            if d.endswith(envdir):
+                continue
+            if os.path.isfile(d):
+                os.remove(d)
+            else:
+                rmtree(d)
```
asottile, line 99: could this be fixed by using a temporary directory to build and then just copy the products into the envdir?

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist = py36,py37,py38,pypy3,pre-commit
 
 [testenv]
 deps = -rrequirements-dev.txt
-passenv = HOME LOCALAPPDATA RUSTUP_HOME
+passenv = HOME LOCALAPPDATA RUSTUP_HOME APPDATA ProgramFiles
```
asottile, line 6: iirc these are case insensitive, let's sort them too!

asottile (commented):
looks great for a first pass, I'm excited about seeing this land!

pre_commit/languages/dotnet.py
rkm: You're right; I think this should be fine to remove. I added it initially since I have a non-standard install and ran into that issue. I've now got `DOTNET_ROOT` set in my env though and the hook still seems to work without this patch.

> (seems silly that they didn't just default it to the location where argv[0] is, but I didn't have time to read the whole dotnet issue!)

Welcome to the wonderful world of c# tooling 😊

pre_commit/languages/dotnet.py
rkm: Yeah I'm not sure if/how this is done for other languages -- it seems like the hook repo is already cloned before `install_environment` is called so there's no chance by that point to clone/build in a temp dir.

I had a minimal version before which removed just the `bin`, `obj`, and `pre-commit-build` dirs since those are the ones we know will exist. I can revert to that if you'd prefer? Or we could skip any cleanup completely. The main reason I added this was that the `format` repo I used in the example above is chunky, and all we need after install is the `envdir`.

asottile: I think `go` does a similar thing and build a temporary build directory and then copies the binaries out of it -- maybe that can be some inspiration?

rkm:
Think I've addressed all your comments now. I've spotted [this](https://stackoverflow.com/questions/52527004/how-to-fix-nu1212-for-dotnet-tool-install) issue though, so will check if we need to handle that here.

pre_commit/languages/dotnet.py
rkm: With `go` we seem to specify where the build artefacts will be placed. I'm not sure we can easily do here without doing bota a `dotnet build` cmd and then a `dotnet pack --no-build` and somehow finding the build output (which can be anywhere really depending on the project layout).

Could we do a `git clean` and exclude the final tool directory?

asottile: a `git clean` would unfortunately preclude doing 1st class in the future so we should avoid that (multiple `language_version`s end up in the same clone directory)

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,92 @@
...
+def install_environment(
+        prefix: Prefix,
+        version: str,
+        additional_dependencies: Sequence[str],
+) -> None:
+
+    helpers.assert_version_default('dotnet', version)
+    helpers.assert_no_additional_deps('dotnet', additional_dependencies)
+
+    repodir = prefix.path(helpers.environment_dir(ENVIRONMENT_DIR, version))
```
asottile, line 47: this is probably better named `envdir`

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,92 @@
...
+            ),
+        )
+
+        # Determine tool from the packaged file <tool_name>.<version>.nupkg
+        build_outputs = os.listdir(os.path.join(prefix.prefix_dir, build_dir))
+        if len(build_outputs) != 1:
+            raise NotImplementedError(
+                "Can't handle multiple build outputs. Got {build_outputs}",
+            )
+        tool_name = build_outputs[0].split('.')[0]
```
asottile, line 68: is this `os.path.splitext` or splitting for a different reason?

pre_commit/languages/dotnet.py
```diff
@@ -0,0 +1,92 @@
...
+            (
+                'dotnet', 'tool', 'install',
+                '--tool-path', os.path.join(repodir, BIN_DIR),
+                '--add-source', build_dir,
+                tool_name,
+            ),
+        )
+
+        # Cleanup build output
+        for d in ['bin', 'obj', build_dir]:
```
asottile, line 82: https://github.com/asottile/scratch/wiki/Prefer-tuples-to-lists

asottile (approved):
looks good! just a few little things and I think this is done (assuming you're happy with it)

rkm:
Heyo. I think I'm happy with this now if you are!

There will probably be some rough edges found once this starts being used in the wild. Happy for you to ping me in any issues though.

pre_commit/languages/dotnet.py
rkm: This parses the tool name e.g. `dotnet_hooks_csproj_repo` from `dotnet_hooks_csproj_repo.1.0.0.nupkg`. I'm not sure splitext would work in this case due to the version specifier also being present as part of the path.

asottile (approved):

asottile:
would you like to update [docs](https://github.com/pre-commit/pre-commit.com) for this? the `_new in_` should be 2.8.0

rkm:
Sure, will do.

PR #1601, opened 2020-09-16
Distinct exit codes for the various error cases

int3l:
This PR attempts to follow the Linux/Bash convention for exit codes. See more at [Exit Codes With Special Meanings](https://tldp.org/LDP/abs/html/exitcodes.html) (You might need to use the [Google web cache](https://www.google.com/search?q=%22Exit+Codes+With+Special+Meanings%22&oq=%22Exit+Codes+With+Special+Meanings%22), because the resource seems to be down right now).

Basically we will have the following:
- `1` for standard error
- `130` for interrupted process by user
- `3` for any unexpected error

Bash/GNU examples which return exit code `130`:
```bash
tail -f /dev/null
# press Ctrl + C
echo $?
# 130

sleep 42
# press Ctrl + C
echo $?
# 130
```

### DIFF
--- a/pre_commit/error_handler.py
+++ b/pre_commit/error_handler.py
@@ -12,7 +12,12 @@ from pre_commit.store import Store
 from pre_commit.util import force_bytes
 
 
-def _log_and_exit(msg: str, exc: BaseException, formatted: str) -> None:
+def _log_and_exit(
+    msg: str,
+    ret_code: int,
+    exc: BaseException,
+    formatted: str,
+) -> None:
     error_msg = f'{msg}: {type(exc).__name__}: '.encode() + force_bytes(exc)
     output.write_line_b(error_msg)
 
@@ -51,7 +56,7 @@ def _log_and_exit(msg: str, exc: BaseException, formatted: str) -> None:
         _log_line('```')
         _log_line(formatted.rstrip())
         _log_line('```')
-    raise SystemExit(1)
+    raise SystemExit(ret_code)
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


int3l:
Windows CI pipeline takes ages to complete 🎃 (25+ minutes)

pre_commit/error_handler.py
```diff
@@ -63,9 +68,9 @@ def error_handler() -> Generator[None, None, None]:
         yield
     except (Exception, KeyboardInterrupt) as e:
         if isinstance(e, FatalError):
-            msg = 'An error has occurred'
+            msg, ret_code = 'An error has occurred', 1
         elif isinstance(e, KeyboardInterrupt):
-            msg = 'Interrupted (^C)'
+            msg, ret_code = 'Interrupted (^C)', 130
```
graingert, line 73: on python 3.8 it will set the exit code correctly if you let it raise

asottile: that would produce a stacktrace so no please don't do this

pre_commit/error_handler.py
```diff
@@ -12,15 +12,11 @@
...
-    pass
-
-
 def _log_and_exit(
-    msg: str,
-    ret_code: int,
-    exc: BaseException,
-    formatted: str,
+        msg: str,
+        returncode: int,
```
int3l, line 17: Changed the name of this argument in order to consistent with the rest of the codebase, for example: https://github.com/pre-commit/pre-commit/search?q=%22returncode%22

pre_commit/error_handler.py
int3l: @graingert I completely understand your point (I also would like to have the appropriate return code for Windows too), but It would be a more complicated PR in that case. Maybe the Windows support should be a separate effort.

asottile (approved):

PR #1606, opened 2020-09-20
check cygwin mismatch earlier

### LINKED ISSUE #1605: Add Support for MSYS2 in Windows
Thanks @asottile and all the other contributors for the great tool! ❤️ 👍 

I am one of the maintainers for [Gaphor](https://github.com/gaphor/gaphor), and we make heavy use of pre-commit as part of our work flow. It is also built in GTK, and we have to use MSYS2 to build the app for Windows. I know it is a pretty niche environment to support, but it would be really great if we could make it happen.

### version information

```
pre-commit version: 2.7.1
sys.version:
    3.8.5 (default, Sep  7 2020, 18:26:56)  [GCC 10.2.0 64 bit (AMD64)]
sys.executable: c:/tools/msys64/home/dyeaw/gaphor/.venv/bin/python.exe
os.name: nt
sys.platform: win32
```

### error information

```
An unexpected error has occurred: FileNotFoundError: [WinError 3] The system cannot find the path specified: '/home/dyeaw/gaphor'
```

```
  File "C:/tools/msys64/home/dyeaw/gaphor/.venv/lib/python3.8/site-packages/pre_commit/error_handler.py", line 63, in error_handler
    yield
  File "C:/tools/msys64/home/dyeaw/gaphor/.venv/lib/python3.8/site-packages/pre_commit/main.py", line 343, in main
    _adjust_args_and_chdir(args)
  File "C:/tools/msys64/home/dyeaw/gaphor/.venv/lib/python3.8/site-packages/pre_commit/main.py", line 163, in _adjust_args_and_chdir
    os.chdir(toplevel)
FileNotFoundError: [WinError 3] The system cannot find the path specified: '/home/dyeaw/gaphor'
```

## Updates Needed
- [ ] Update pre-commit to make use of the `cygpath` command to convert the path to the standard Windows path format after it gets the top-level path from git.
- [ ] Remove cygwin warnings about incompatible Python+git versions if in the MSYS2 environment
- [ ] Make upstream updates to setuptools_scm to patch additional path conversions for similar things as 1 above.

My plan, as long is everyone is good with adding support for this environment, is to submit a PR for the first two, and then I'll start to work with the setuptools_scm team to see if I can fix the third one as well.

asottile:
resolves #1605

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -13,7 +13,7 @@ from identify.identify import ALL_TAGS
 
 import pre_commit.constants as C
 from pre_commit.color import add_color_option
-from pre_commit.error_handler import FatalError
+from pre_commit.errors import FatalError
 from pre_commit.languages.all import all_languages
 from pre_commit.logging_handler import logging_handler
 from pre_commit.util import parse_version
--- a/pre_commit/error_handler.py
+++ b/pre_commit/error_handler.py
@@ -7,12 +7,9 @@ from typing import Generator
 
 import pre_commit.constants as C
 from pre_commit import output
+from pre_commit.errors import FatalError
 from pre_commit.store import Store
 from pre_commit.util import force_bytes
-
-
-class FatalError(RuntimeError):
-    pass
 
 
 def _log_and_exit(msg: str, exc: BaseException, formatted: str) -> None:
--- a/pre_commit/errors.py
+++ b/pre_commit/errors.py
@@ -1,0 +1,2 @@
+class FatalError(RuntimeError):
+    pass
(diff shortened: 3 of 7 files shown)


danyeaw:
@asottile I tried this out, and it does move the cygwin mismatch check up earlier. I don't think it fixes the issue though, it isn't that I am missing the error because of the other issues, the problem is that I shouldn't be getting this warning at all. I am not running in cygwin, I am in MSYS2 and my git and Python aren't mismatched.

asottile:
they are mismatched, one is acting like windows and the other is acting like posix -- that's the entire point of the warning

PR #1607, opened 2020-09-20
Improve MSYS2 Compatibility

danyeaw:
This PR closes part of #1605 

It makes two updates:
1. Update pre-commit to make use of the cygpath command to convert the path to the standard Windows path format after it gets the top-level path from git.
2. Remove cygwin warnings about incompatible Python+git versions if in the MSYS2 environment


asottile:
yeah no sorry, this is just the tip of the iceberg -- one is acting posixlike and one is acting windows-like and sprinkling cygpath is a hack at best.  the warning is correct (one is acting like posix and the other is acting like windows) and that will not be supported

Issue #1454, opened 2020-05-12
Error creating pip environment when index requires a login


sodre:
I am having a problem creating a pre-commit virtual environment when running behind a proxying/private pypi repository that requires a login.
I have setup my pip.conf file to point to that index, and I am using `keyring` to manage my passwords, so it is not stored in plain-text.

To reproduce on my system:

```
$ pip install pre-commit
# no passwords are prompted
```

Create a very simple `.pre-commit-config.yaml` file, with "psf/black"
```
repos:
  - repo: https://github.com/psf/black
    rev: stable
    hooks:
      - id: black
```

Try to initialize the `pre-commit` virtual environments...
<details>

```
± pre-commit
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/home/pcarlos/.cache/pre-commit/repodmjniiuo/py_env-python3/bin/python', '-mpip', 'install', '.')
return code: 1
expected return code: 0
stdout:
    Looking in indexes: https://<NEXUS PRIVATE REPOSITORY>/repository/pypi-all/simple
    Processing /home/pcarlos/.cache/pre-commit/repodmjniiuo
      Installing build dependencies: started
      Installing build dependencies: finished with status 'error'
    User for <NEXUS PRIVATE REPOSITORY>: 
stderr:
      ERROR: Command errored out with exit status 2:
       command: /home/pcarlos/.cache/pre-commit/repodmjniiuo/py_env-python3/bin/python /home/pcarlos/.cache/pre-commit/repodmjniiuo/py_env-python3/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-oget82ct/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://<NEXUS PRIVATE REPOSITORY>/repository/pypi-all/simple -- 'setuptools>=41.0' setuptools-scm wheel
           cwd: None
      Complete output (48 lines):
      Looking in indexes: https://<NEXUS PRIVATE REPOSITORY>/repository/pypi-all/simple
      Collecting setuptools>=41.0
      User for <NEXUS PRIVATE REPOSITORY>: ERROR: Exception:
```
</details>

Traceback:
<details>

```python-traceback
      Traceback (most recent call last):
[...]

PR #1624, opened 2020-10-02
add '--isolated' to pip install to avoid authentication attempt to private PyPI

basnijholt:
Solves https://github.com/pre-commit/pre-commit/issues/1454

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -196,7 +196,14 @@ def install_environment(
     python = norm_version(version)
     if python is not None:
         venv_cmd.extend(('-p', python))
-    install_cmd = ('python', '-mpip', 'install', '--isolated', '.', *additional_dependencies)
+    install_cmd = (
+        'python',
+        '-mpip',
+        'install',
+        '--isolated',
+        '.',
+        *additional_dependencies,
+    )
 
     with clean_path_on_failure(envdir):
         cmd_output_b(*venv_cmd, cwd='/')
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -196,7 +196,7 @@ def install_environment(
     python = norm_version(version)
     if python is not None:
         venv_cmd.extend(('-p', python))
-    install_cmd = ('python', '-mpip', 'install', '.', *additional_dependencies)
+    install_cmd = ('python', '-mpip', 'install', '--isolated', '.', *additional_dependencies)
 
     with clean_path_on_failure(envdir):
         cmd_output_b(*venv_cmd, cwd='/')


astafan8:
i can confirm that this worked for me! just tested by installing pre-commit via pip from this branch. thank you @basnijholt !!

pre_commit/languages/python.py
```diff
@@ -196,7 +196,14 @@ def install_environment(
     python = norm_version(version)
     if python is not None:
         venv_cmd.extend(('-p', python))
-    install_cmd = ('python', '-mpip', 'install', '.', *additional_dependencies)
+    install_cmd = (
+        'python',
+        '-mpip',
+        'install',
+        '--isolated',
```
asottile, line 203: this will break people who depend on pip configuration variables, especially `PIP_INDEX_URL` (a common enterprise setup)

asottile (changes requested):

pre_commit/languages/python.py
astafan8: if not the change in this PR, what is recommended for enterprise setups that use pip.ini file to specify extra index url?

another alternative - perhaps pre-commit's config (don't know where) can allow a user to specify extra args for the pip call?

asottile: are you typing passwords every time you install? should you put an api key in your pip configuration?

astafan8: > are you typing passwords every time you install

no, I use `keyring` package to manage the passwords/tokens/etc.

> should you put an api key in your pip configuration?

i'm not a fan of having username/passwords in pip.ini file in plain text format. another solutipon would be to store them in ~/.netrc file but it's still plain-text-storage-on-disk. while keyring at least tries to hide things in some decent way.

basnijholt: @asottile, it doesn't seem like there is a satisfying solution to this problem. Therefore, I do not understand your premature closing of this PR.

asottile: `--isolated` is not a solution, therefore I closed the PR -- I'm not sure why you think that's premature?

astafan8: > I'm not sure why you think that's premature?

because it's a real problem :)

however, i think the salomon solution would be to let pre-commit pass some extra args to how it calls pip internally. that way folks like me will be able to pass --isolated, and some other folks will be able to NOT pass it in order to use `PIP_INDEX_URL` or whatever else there is. is this a decent approach?

asottile: > > I'm not sure why you think that's premature?
>
> because it's a real problem :)

well this PR is not a solution => therefore closed.

> that way folks like me will be able to pass --isolated

can't you already do that by setting `PIP_ISOLATED=1`? or `additional_dependencies: [--isolated]`?

astafan8: > well this PR is not a solution => therefore closed

fair.

> can't you already do that by setting PIP_ISOLATED=1? or additional_dependencies: [--isolated]?

`additional_dependencies: [--isolated]` would be great, but it didn't work (i'll try again to confirm that i did it right). and setting an environment variable only to run a single command does not seem sustainable for me.

asottile: if it didn't work in `additional_dependencies` then this patch won't work, they're exactly the same thing

PR #1643, opened 2020-10-14
ENH add --negate flag to pygrep

MarcoGorelli:
xref this comment on StackOverflow: https://stackoverflow.com/a/64344519/9988333

>  A PR may be considered if you can demonstrate a real use case for it

A use-case comes from `PyMC3`, where there is a pre-commit hook to check that all notebooks have a watermark. Currently, this is done by defining
```
- repo: local
  hooks:
    - id: watermark
      name: Check notebooks have watermark
      types: [jupyter]
      entry: python scripts/check_watermark.py
      language: python
```
in https://github.com/pymc-devs/pymc3/blob/master/.pre-commit-config.yaml, and then
```python
"""
Check that given Jupyter notebooks all contain a final watermark cell to facilite reproducibility.
This is intended to be used as a pre-commit hook, see `.pre-commit-config.yaml`.
You can run it manually with `pre-commit run watermark --all`.
"""

import argparse
from pathlib import Path
import re

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("filenames", nargs="*")
    args = parser.parse_args()
    for file_ in args.filenames:
        assert (
            re.search(
                r"%load_ext watermark.*%watermark -n -u -v -iv -w",
                Path(file_).read_text(),
                flags=re.DOTALL,
            )
            is not None
        ), (
            f"Watermark not found in {file_} - please see the PyMC3 Jupyter Notebook Style guide:\n"
            "https://github.com/pymc-devs/pymc3/wiki/PyMC's-Jupyter-Notebook-Style"
        )
```
in https://github.com/pymc-devs/pymc3/blob/master/scripts/check_watermark.py .

----

I think the output should be a file which does _not_ contain the pattern...I'll work on this tomorrow

### DIFF
--- a/pre_commit/languages/pygrep.py
+++ b/pre_commit/languages/pygrep.py
@@ -1,6 +1,7 @@
 import argparse
 import re
 import sys
+from typing import NamedTuple
 from typing import Optional
 from typing import Pattern
 from typing import Sequence
@@ -45,6 +46,46 @@ def _process_filename_at_once(pattern: Pattern[bytes], filename: str) -> int:
     return retv
 
 
+def _process_filename_by_line_negated(
+        pattern: Pattern[bytes],
+        filename: str,
+) -> int:
+    with open(filename, 'rb') as f:
+        for line in f:
+            if pattern.search(line):
+                return 0
+        else:
+            output.write_line(filename)
+            return 1
+
+
+def _process_filename_at_once_negated(
+        pattern: Pattern[bytes],
+        filename: str,
+) -> int:
+    with open(filename, 'rb') as f:
+        contents = f.read()
+    match = pattern.search(contents)
+    if match:
+        return 0
+    else:
+        output.write_line(filename)
+        return 1
+
+
+class Choice(NamedTuple):
+    multiline: bool
+    negate: bool
+
+
+FNS = {
+    Choice(multiline=True, negate=True): _process_filename_at_once_negated,
+    Choice(multiline=True, negate=False): _process_filename_at_once,
+    Choice(multiline=False, negate=True): _process_filename_by_line_negated,
+    Choice(multiline=False, negate=False): _process_filename_by_line,
+}
+
+
 def run_hook(
         hook: Hook,
         file_args: Sequence[str],
@@ -64,6 +105,7 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
     )
     parser.add_argument('-i', '--ignore-case', action='store_true')
     parser.add_argument('--multiline', action='store_true')
+    parser.add_argument('--negate', action='store_true')
     parser.add_argument('pattern', help='python regex pattern.')
     parser.add_argument('filenames', nargs='*')
     args = parser.parse_args(argv)
@@ -75,11 +117,9 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
     pattern = re.compile(args.pattern.encode(), flags)
 
     retv = 0
+    process_fn = FNS[Choice(multiline=args.multiline, negate=args.negate)]
     for filename in args.filenames:
-        if args.multiline:
-            retv |= _process_filename_at_once(pattern, filename)
-        else:
-            retv |= _process_filename_by_line(pattern, filename)
+        retv |= process_fn(pattern, filename)
     return retv
 
 
--- a/tests/languages/pygrep_test.py
+++ b/tests/languages/pygrep_test.py
@@ -8,6 +8,9 @@ def some_files(tmpdir):
     tmpdir.join('f1').write_binary(b'foo\nbar\n')
     tmpdir.join('f2').write_binary(b'[INFO] hi\n')
     tmpdir.join('f3').write_binary(b"with'quotes\n")
+    tmpdir.join('f4').write_binary(b'foo\npattern\nbar\n')
+    tmpdir.join('f5').write_binary(b'[INFO] hi\npattern\nbar')
+    tmpdir.join('f6').write_binary(b"pattern\nbarwith'foo\n")
     with tmpdir.as_cwd():
         yield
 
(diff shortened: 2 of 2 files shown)


pre_commit/languages/pygrep.py
```diff
@@ -80,6 +81,8 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
             retv |= _process_filename_at_once(pattern, filename)
         else:
             retv |= _process_filename_by_line(pattern, filename)
+    if args.negate:
+        retv = 1 - retv
```
asottile, line 85: this is buggy, when processing multiple files if any of them match it will return 0

tests/languages/pygrep_test.py
```diff
@@ -23,9 +24,21 @@ def some_files(tmpdir):
...
+    cap_out,
+    pattern,
+    expected_retcode,
+    expected_out,
+    negate,
+):
+    if not negate:
+        ret = pygrep.main((pattern, 'f1', 'f2', 'f3'))
+    else:
+        ret = pygrep.main((pattern, 'f1', 'f2', 'f3', '--negate'))
```
asottile, line 38: [don't put logic in tests](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html), write a separate test for this clearly demonstrating the use case

pre_commit/languages/pygrep.py
asottile: also `1 - retv` is much harder to understand than `retv = not retv`

MarcoGorelli: yes, you're right, thanks, sorry for having submitted this without having though it through

MarcoGorelli:
Marking as ready for review as I'd like to think this is enough to tell whether it's the right direction / whether the feature would be welcome. Thanks for the 'don't put logic in tests' link!

pre_commit/languages/pygrep.py
```diff
@@ -17,31 +17,47 @@
...
+                matched_lines = match[0].split(b'\n')
+                matched_lines[0] = contents.split(b'\n')[line_no]
+
+                output.write_line_b(b'\n'.join(matched_lines))
+        else:
+            if negate:
+                output.write_line(filename)
 
-            output.write_line_b(b'\n'.join(matched_lines))
     return retv
```
asottile, line 61: these functions now both are pretty hard to understand and have a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) -- I'd suggest refactoring them into 4 separate easier to understand functions

MarcoGorelli: Thanks for your review, and for another really good link!

Sure, I've made another attempt at this

pre_commit/languages/pygrep.py
```diff
@@ -45,6 +45,36 @@ def _process_filename_at_once(pattern: Pattern[bytes], filename: str) -> int:
...
+) -> int:
+    retv = 1
+    with open(filename, 'rb') as f:
+        contents = f.read()
+        match = pattern.search(contents)
+        if match:
+            retv = 0
+    if retv:
+        output.write_line(filename)
+    return retv
```
asottile, line 75: simpler:

remove `retv = 1`

do

```
if match:
    return 0
else:
    output.write_line(filename)
    return 1
```

pre_commit/languages/pygrep.py
```diff
@@ -45,6 +45,36 @@ def _process_filename_at_once(pattern: Pattern[bytes], filename: str) -> int:
...
+def _process_filename_by_line_negated(
+        pattern: Pattern[bytes],
+        filename: str,
+) -> int:
+    retv = 1
+    with open(filename, 'rb') as f:
+        for line in f:
+            if pattern.search(line):
+                retv = 0
+                break
```
asottile, line 57: this can be `return 0`

and then below you can
```python
else:
    output.write_line(filename)
    return 1
```

MarcoGorelli: Yes, much simpler, thanks!

Though here there wouldn't be an `else`, would there? It's necessary to check through all the lines before deciding to return 1

asottile: I mean [`for: ... else: ...`](https://youtu.be/8P7lXLXR_3c?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY)

pre_commit/languages/pygrep.py
```diff
@@ -45,6 +45,31 @@ def _process_filename_at_once(pattern: Pattern[bytes], filename: str) -> int:
...
+        pattern: Pattern[bytes],
+        filename: str,
+) -> int:
+    with open(filename, 'rb') as f:
+        contents = f.read()
+        match = pattern.search(contents)
+    if match:
+        return 0
+    output.write_line(filename)
+    return 1
```
asottile, line 70: please write `if: ... else: ...` -- it is much easier for humans to see at a glance that they are disparate branches

also if you're doing to dedent the `if match` you might as well dedent `match = ...` as well

pre_commit/languages/pygrep.py
MarcoGorelli: Ah, I see, thanks! I see you have a video about this too, will watch now https://youtu.be/8P7lXLXR_3c

pre_commit/languages/pygrep.py
```diff
@@ -75,6 +103,15 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
...
 
     retv = 0
+
+    if args.negate:
+        for filename in args.filenames:
+            if args.multiline:
+                retv |= _process_filename_at_once_negated(pattern, filename)
+            else:
+                retv |= _process_filename_by_line_negated(pattern, filename)
+        return retv
```
asottile, line 113: you can probably do:

```
# probably at module scope
class Choice(NamedTuple):
    multiline: bool
    negate: bool
FNS = {
    Choice(multiline=True, negate=True): _process_filename_at_once_negated,
    Choice(multiline=True, negate=False): _process_filename_at_once,
    Choice(multiline=False, negate=True): _process_filename_by_line_negated,
    Choice(multiline=False, negate=False): _process_filename_by_line,
}
# in code
process_fn = FNS[Choice(multiline=args.multiline, negate=args.negate)]

# then the for loop code, but without branches
```

just because there's a lot more duplication than before

MarcoGorelli: @asottile yes, that's much cleaner, thanks!

pre_commit/languages/pygrep.py
```diff
@@ -75,11 +117,10 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
     pattern = re.compile(args.pattern.encode(), flags)
 
     retv = 0
+
     for filename in args.filenames:
-        if args.multiline:
-            retv |= _process_filename_at_once(pattern, filename)
-        else:
-            retv |= _process_filename_by_line(pattern, filename)
+        process_fn = FNS[Choice(multiline=args.multiline, negate=args.negate)]
```
asottile, line 122: this should be outside the loop, otherwise you're doing this work every iteration instead of just once

tests/languages/pygrep_test.py
```diff
@@ -30,6 +33,40 @@ def test_main(cap_out, pattern, expected_retcode, expected_out):
...
+        ('pattern', 0, ''),
+        ('foo', 1, 'f5\n'),
+        ('pattern\nbar', 0, ''),
+    ),
+)
+def test_negate_by_file(cap_out, pattern, expected_retcode, expected_out):
+    ret = pygrep.main(('--multiline', pattern, 'f4', 'f5', 'f6', '--negate'))
+    out = cap_out.get()
+    assert ret == expected_retcode
+    assert out == expected_out
```
asottile, line 67: these tests might be easier to understand without parametrize -- 4 separate tests showing positive / negative for each case?  up to you

asottile (approved):

MarcoGorelli:
Thank you so much for the mentorship you provided me here, I really appreciate it, it's very kind of you!

PR #1668, opened 2020-10-28
don't use system for ruby/node if it is a shim exe

### LINKED ISSUE #1658: rbenv FileNotFoundError when installing markdownlint hook
I am trying to use https://github.com/markdownlint/markdownlint to lint my markdown files. I checked the issue tracker and [don't seem to see any related issues](https://github.com/pre-commit/pre-commit/issues?q=FileNotFoundError+rbenv). 

I have `.rbenv` installed on my system and am wondering if that could be interfering with `pre-commit`?

### .pre-commit-config.yaml
```
repos:
-   repo: https://github.com/markdownlint/markdownlint
    rev: master
    hooks:
    -   id: markdownlint
```

### version information

```
pre-commit version: 2.7.1
sys.version:
    3.9.0 (default, Oct  9 2020, 16:41:50) 
    [Clang 12.0.0 (clang-1200.0.32.2)]
sys.executable: /usr/local/Cellar/pre-commit/2.7.1_1/libexec/bin/python
os.name: posix
sys.platform: darwin
```

### error information

```
An unexpected error has occurred: FileNotFoundError: [Errno 2] No such file or directory: '/Users/adithyabalaji/.cache/pre-commit/repo09b5y8iv/rbenv-system/.install_state_v1staging'
```

```
  File "/usr/local/Cellar/pre-commit/2.7.1_1/libexec/lib/python3.9/site-packages/pre_commit/error_handler.py", line 63, in error_handler
    yield
  File "/usr/local/Cellar/pre-commit/2.7.1_1/libexec/lib/python3.9/site-packages/pre_commit/main.py", line 390, in main
    return run(args.config, store, args)
  File "/usr/local/Cellar/pre-commit/2.7.1_1/libexec/lib/python3.9/site-packages/pre_commit/commands/run.py", line 388, in run
    install_hook_envs(hooks, store)
  File "/usr/local/Cellar/pre-commit/2.7.1_1/libexec/lib/python3.9/site-packages/pre_commit/repository.py", line 206, in install_hook_envs
    _hook_install(hook)
  File "/usr/local/Cellar/pre-commit/2.7.1_1/libexec/lib/python3.9/site-packages/pre_commit/repository.py", line 92, in _hook_install
    _write_state(hook.prefix, venv, _state(hook.additional_dependencies))
  File "/usr/local/Cellar/pre-commit/2.7.1_1/libexec/lib/python3.9/site-packages/pre_commit/repository.py", line 48, in _write_state
    with open(staging, 'w') as state_file:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/adithyabalaji/.cache/pre-commit/repo09b5y8iv/rbenv-system/.install_state_v1staging'

```

asottile:
Resolves #1658

### DIFF
--- a/pre_commit/languages/helpers.py
+++ b/pre_commit/languages/helpers.py
@@ -1,6 +1,7 @@
 import multiprocessing
 import os
 import random
+import re
 from typing import Any
 from typing import List
 from typing import Optional
@@ -10,6 +11,7 @@ from typing import Tuple
 from typing import TYPE_CHECKING
 
 import pre_commit.constants as C
+from pre_commit import parse_shebang
 from pre_commit.hook import Hook
 from pre_commit.prefix import Prefix
 from pre_commit.util import cmd_output_b
@@ -19,6 +21,25 @@ if TYPE_CHECKING:
     from typing import NoReturn
 
 FIXED_RANDOM_SEED = 1542676187
+
+SHIMS_RE = re.compile(r'[/\\]shims[/\\]')
+
+
+def exe_exists(exe: str) -> bool:
+    found = parse_shebang.find_executable(exe)
+    if found is None:  # exe exists
+        return False
+
+    homedir = os.path.expanduser('~')
+    try:
+        common: Optional[str] = os.path.commonpath((found, homedir))
+    except ValueError:  # on windows, different drives raises ValueError
+        common = None
+
+    return (
+        not SHIMS_RE.search(found) and  # it is not in a /shims/ directory
+        common != homedir  # it is not in the home directory
+    )
 
 
 def run_setup_cmd(prefix: Prefix, cmd: Tuple[str, ...]) -> None:
--- a/pre_commit/languages/node.py
+++ b/pre_commit/languages/node.py
@@ -7,7 +7,6 @@ from typing import Sequence
 from typing import Tuple
 
 import pre_commit.constants as C
-from pre_commit import parse_shebang
 from pre_commit.envcontext import envcontext
 from pre_commit.envcontext import PatchesT
 from pre_commit.envcontext import UNSET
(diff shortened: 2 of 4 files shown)


avdv:
Hi.

This is a really disruptive change IMO, which - from the discription - should only have an impact for ruby and node shims.

But basically this change disallows any binaries installed inside the home directory. Is this intentional?

Asking because pre-commit failed installing a hook, since it could not find a binary which is available in my path, at `~/.nix-profile/bin` and symlinked to the real binary somewhere else. Lost an hour to find the reason for this...

asottile:
> This is a really disruptive change IMO, which - from the discription - should only have an impact for ruby and node shims.

this is correct, it only affects ruby and node as the code here is only used for ruby and node

> But basically this change disallows any binaries installed inside the home directory.

if you read the change, it does not do that, it must contain both `/shims/` and be in the home directory

it also doesn't ban them -- it just turns off the automatic `language_version: system` when a suitable binary is detected

> Lost an hour to find the reason for this...

avdv:
> if you read the change, it does not do that, it must contain both `/shims/` and be in the home directory

I think that's wrong. The `exe_exists` function only returns true if the path does not contain shim *and* the path is not inside the home dir.

Should this not rather be something like `return not (SHIMS_RE.search(found) and common != homedir)` -- which is basically what you said above?

asottile:
no it's meant to ban both homedir and shims executables

please describe your actual problem behaviour instead of solution hunting, I suspect you're lost and trying to find something which isn't causing your problem and wasting all of our time

avdv:
> no it's meant to ban both homedir and shims executables

But why? The problem started with something specific to rbenv shims. Now, all binaries inside the home dir (for node and ruby) are banned. Why?

> please describe your actual problem behaviour instead of solution hunting, I suspect you're lost and trying to find something ? which isn't causing your problem and wasting all of our time

Using downloaded binaries does not usually work on my system. So, having nodeenv fetch some arbirtary binary from the net (II suspect it does that) will fail. I was using pre-commit 2.7 before it everything worked smoothly. Now pre-commit failed because of this disruptive change.

asottile:
the reason is home-directory installed binaries are likely to target shifting sands, breaking environments after the fact -- the specific issue was about rbenv which makes it very easy to do this but the problem persists for other managers as well

you can easily restore the previous default with:

```yaml
default_language_version:
    node: system
```

asottile:
also, have some empathy "this disruptive change" is not appropriate language.  you're getting free software and you're not sponsoring so I don't have to take this attitude from you

PR #1677, opened 2020-11-02
Add types_or

### LINKED ISSUE #607: `OR` relationship with `types`
When we introduced this feature, I had some specific usecases in mind which required an `AND` relationship with `types` -- it comes to mind that some tools would like to use an `OR` relationship, but there isn't a great way to implement this without reverting to `files` (and thus losing the benefits of `types` altogether!).

This issue is to track _some_ sort of way of enabling an `OR` relationship.

This was originally inspired by https://github.com/prettier/prettier/issues/2745

MarcoGorelli:
closes #607


pre_commit/commands/run.py
```diff
@@ -83,20 +83,30 @@ def by_types(
...
             types: Collection[str],
+            types_or: Collection[str],
             exclude_types: Collection[str],
     ) -> List[str]:
-        types, exclude_types = frozenset(types), frozenset(exclude_types)
+        types, types_or, exclude_types = (
+            frozenset(types),
+            frozenset(types_or),
+            frozenset(exclude_types),
+        )
```
asottile, line 93: this should be three separate lines

pre_commit/commands/run.py
```diff
@@ -83,20 +83,30 @@ def by_types(
...
         ret = []
         for filename in names:
             tags = self._types_for_file(filename)
-            if tags >= types and not tags & exclude_types:
+            if tags >= types and not tags & exclude_types and tags & types_or:
```
asottile, line 97: teeny tiny thing, I'd order the conditions the same as the parameters

pre_commit/meta_hooks/check_useless_excludes.py
```diff
@@ -47,8 +47,12 @@ def check_useless_excludes(config_file: str) -> int:
             # the defaults applied during runtime
             hook = apply_defaults(hook, MANIFEST_HOOK_DICT)
             names = classifier.filenames
-            types, exclude_types = hook['types'], hook['exclude_types']
-            names = classifier.by_types(names, types, exclude_types)
+            types, types_or, exclude_types = (
+                hook['types'],
+                hook['types_or'],
+                hook['exclude_types'],
+            )
```
asottile, line 54: ""

testing/resources/types_or_repo/.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+-   id: python-cython-files
+    name: Python and Cython files
+    entry: bin/hook.sh
+    language: script
+    types: [file]
+    types_or: [python, cython]
```
asottile, line 6: these are usually easier to demo with a local repo

I'd probably not write a full integration test for this feature as well but it looks like there maybe aren't great examples using the matcher

testing/resources/types_or_repo/bin/hook.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+echo $@
```
asottile, line 2: ~technically this should be `echo "$@"`

asottile (commented):
good start!

I believe check_hooks_apply and check_useless_excludes need to be updated

MarcoGorelli:
> good start!

Thanks!

> I believe check_hooks_apply and check_useless_excludes need to be updated

Sorry, this isn't clear to me - I updated `check_useless_excludes`, and it's not obvious to me how `check_hooks_apply` needs to be updated, any hints?

asottile:
> > I believe check_hooks_apply and check_useless_excludes need to be updated
> 
> Sorry, this isn't clear to me - I updated `check_useless_excludes`, and it's not obvious to me how `check_hooks_apply` needs to be updated, any hints?

oh maybe I'm wrong, maybe it already works

testing/resources/types_or_repo/.pre-commit-hooks.yaml
MarcoGorelli: Thanks for your feedback, but I'm not sure I understand - are there any existing tests I should look at as examples? If not, I'll try to figure this out, no worries

asottile (approved):

PR #1678, opened 2020-11-02
fix dotnet build cleanup

rkm:
From testing, I found that the `bin` and `obj` directories may be elsewhere depending on the particular hook project layout. I'd also like to cleanup the build directory properly, since we only need the `dotnetenv-default` dir after we install the tool.

This patch should solve both cases.

### DIFF
--- a/pre_commit/languages/dotnet.py
+++ b/pre_commit/languages/dotnet.py
@@ -12,7 +12,6 @@ from pre_commit.hook import Hook
 from pre_commit.languages import helpers
 from pre_commit.prefix import Prefix
 from pre_commit.util import clean_path_on_failure
-from pre_commit.util import rmtree
 
 ENVIRONMENT_DIR = 'dotnetenv'
 BIN_DIR = 'bin'
@@ -76,9 +75,9 @@ def install_environment(
             ),
         )
 
-        # Cleanup build output
-        for d in ('bin', 'obj', build_dir):
-            rmtree(prefix.path(d))
+        # Clean the git dir, ignoring the environment dir
+        clean_cmd = ('git', 'clean', '-ffxd', '-e', f'{ENVIRONMENT_DIR}-*')
+        helpers.run_setup_cmd(prefix, clean_cmd)
 
 
 def run_hook(


pre_commit/languages/dotnet.py
```diff
@@ -76,9 +76,17 @@ def install_environment(
...
+        to_keep = (
+            helpers.environment_dir(ENVIRONMENT_DIR, version),
+            '.pre-commit-hooks.yaml',
+        )
+        for f in os.listdir(prefix.prefix_dir):
+            path = prefix.path(f)
+            if os.path.isfile(path):
+                os.remove(path)
+            elif f not in to_keep:
+                rmtree(path)
```
asottile, line 89: the problem with this is it will break when introducing `language_version` for `dotnet` as that allows multiple builds in a single clone

is there a way we can leverage `git clean` to do this for us?

rkm: Ah, good point -- I had hoped we could blitz the entire git directory after the tool install.

I'll look for a `git clean`er solution

asottile: there's also the potential that a `script` hook is defined in the same repo as well

pre_commit/languages/dotnet.py
```diff
@@ -76,9 +75,14 @@ def install_environment(
...
 
-        # Cleanup build output
-        for d in ('bin', 'obj', build_dir):
-            rmtree(prefix.path(d))
+        # Clean the git dir, ignoring the environment dir
+        helpers.run_setup_cmd(
+            prefix,
+            (
+                'git', 'clean', '-fxd',
+                '-e', helpers.environment_dir(ENVIRONMENT_DIR, version),
```
asottile, line 83: this should probably be `f'{ENVIRONMENT_DIR}-*'`

pre_commit/languages/dotnet.py
```diff
@@ -76,9 +75,14 @@ def install_environment(
...
         )
 
-        # Cleanup build output
-        for d in ('bin', 'obj', build_dir):
-            rmtree(prefix.path(d))
+        # Clean the git dir, ignoring the environment dir
+        helpers.run_setup_cmd(
+            prefix,
+            (
+                'git', 'clean', '-fxd',
```
asottile, line 82: I've actually got a video coming up soon that indicates this should be `-ffxd` :)

asottile (approved):

PR #1684, opened 2020-11-06
only perform migrate_config parsing if it is a list

asottile:


asottile:
diff is [best viewed with ?w=1](https://github.com/pre-commit/pre-commit/pull/1684/files?w=1)

PR #1685, opened 2020-11-06
Add a coursier_launch integration

JosephMoniz:
This is a follow up to the repository based coursier integration. The
standard coursier integration is to this integration what the docker
integration is to the docker_image integration.

This is meant to target either maven coordinates or coursier apps
directly without depending on a hook repository containing app
descriptors and hook configs making this integration highly suitable
for local hooks.

An example of hook local targets are bellow
```yaml
- repo: local
  hooks:
    - id: scalafmt
      name: scalafmt
      entry: scalafmt # coursier app target
      language: coursier_launch
      files: (\.scala|\.sbt|\.sc)$
    - id: scalafix
      name: scalafix
      entry: ch.epfl.scala:::scalafix-cli:latest.release # maven coordinates target
      language: coursier_launch
      files: (\.scala|\.sbt|\.sc)$
    - id: echo-java
      name: echo-java
      entry: echo-java
      language: coursier_launch
      args: ["--", "hello from java"] # forward arguments to the app
```

### DIFF
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -7,6 +7,7 @@ from typing import Tuple
 from pre_commit.hook import Hook
 from pre_commit.languages import conda
 from pre_commit.languages import coursier
+from pre_commit.languages import coursier_launch
 from pre_commit.languages import docker
 from pre_commit.languages import docker_image
 from pre_commit.languages import dotnet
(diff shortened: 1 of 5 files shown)


asottile:
would this be better implemented as `additional_dependencies` for `coursier`? that would be more like the current `language`s

asottile:
> 
> 
> would this be better implemented as `additional_dependencies` for `coursier`? that would be more like the current `language`s

especially as you can make a `language: system` hook today that's just `cs launch ...` -- the benefit to a `language would be that it can cache the installation

JosephMoniz:
From a JVM artifact perspective, `additional_dependencies` feels a bit confusing. It's unclear if the additional dependencies will be bundled in the JVM artifacts were fetching and building as executables or if we're fetching additional executable artifacts.

Also, now that you point out the caching thing, launch is doing its own artifact caching behind pre-commit's back. Maybe that's not the correct thing to do. Maybe this should simulate the behavior of launch by doing a local install of the maven/app artifact and then just running that on hook run. Thoughts?

asottile:
`rust` has the same confusion and so the `additional_dependencies` syntax adds a special `cli:` for differentiating bundled / non bundled dependencies

asottile:
(indicating: we could do the same for `coursier`)

JosephMoniz:
Ok cool, this is interesting.

Let's say we want to use `additional_dependencies` to install a cli artifact like `cli:scalafix` but we want to add a plugin dependency to `scalafix`, say the sort imports plugin. How do you feel about doing something like the following?
```yaml
additional_dependencies:
    - cli:scalafmt
    - cli:scalafix
    - scalafix<~com.nequissimus::sort-imports:0.5.5
```

asottile:
seems fine, as long as it's unambiguous

JosephMoniz:
The approach discussed in this thread seems way better than what this PR does. I'll go ahead and rework the existing coursier integration to implement this. Thanks for the nudge in the right direction man!

laughedelic:
Hi! @JosephMoniz thanks for integrating pre-commit with coursier! Have you had a chance to give a go to this new approach?

PR #1688, opened 2020-11-07
add test to guard against yaml_dump

asottile:


asottile:
tried to simplify here, got owned by yaml: https://twitter.com/codewithanthony/status/1324880502533181440

Issue #757, opened 2018-05-29
Support custom pip binary for python hooks


firba1:
Currently, we have a hook repository which relies on using [pip-custom-platform](https://github.com/asottile/pip-custom-platform) in order to be installed properly. However, since pre-commit always uses vanilla pip, we're having issues where wheel builds fail on hook install.

asottile:
Do you have a proposal for how this would work?  Currently there's a single invocation of pip:

https://github.com/pre-commit/pre-commit/blob/cf5f8406a1593aa31dcfa5fb2a3766a9ff3f8a96/pre_commit/languages/python.py#L144-L146

firba1:
Unfortunately, I don't really. The biggest issue is probably the fact that pip-custom-platform has to be bootstrapped into the virtualenv before it can be used for installation. I wonder if it might make sense to have a separate language profile specifically for using that? I'm spitballing here.

chriskuehl:
A little orthogonal to the general issue here (there could still be use-cases for using a different `pip` tool), but if we can get traction on https://github.com/pypa/pip/issues/5453, it could make it a lot easier to avoid needing `pip-custom-platform` with pre-commit.

chriskuehl:
@asottile if we could find a reasonable way to implement this, would you take a patch? Or is this something you think should be solved outside of pre-commit somehow?

This is causing us quite a bit of trouble as we don't have linux_x86_64 wheels anymore on our internal PyPI, requiring us to build packages on every install.

asottile:
Yeah I don't think the idea is completely unreasonable, but there's no proposal here yet :)

asottile:
(an aside: we're tackling this with separate subdirectories on a single pypi server, basically `export PIP_INDEX_URL="https://pypi.example.com/${DISTRIB_CODENAME}/simple"`)

asottile:
I don't think this'll ever be done so I'm going to close this

PR #1697, opened 2020-11-16
Add PYTHON_SETUP_DEPENDENCIES, which adds dependencies to install in each virtualenv

AllSeeingEyeTolledEweSew:
My use case is that I'm in an environment where I need to use a socks proxy for all traffic. I actually use [tails](https://tails.boum.org), which restricts all outgoing traffic through tor.

virtualenvs *almost* work in this environment: `pip` respects the `https_proxy` environment variable and has a `--proxy` argument; but `pip` does not have built-in socks support and requires the `pysocks` package. So creating a new virtualenv creates a chicken-and-egg problem: how do you install `pysocks` if `pip` requires `pysocks` to work?

I can't find a well-supported solution to this problem. The best answer I can find is to download a `pysocks` wheel with a web browser and `pip install` the wheel file.

This PR is a hacky way to apply this solution in `pre-commit`. Setting up a python environment now respects a `PYTHON_SETUP_DEPENDENCIES` environment variable, which is a `:`-separated list of things to `pip install`, before `pip install`'ing the normal things for the environment.

So for my use case, I download `pysocks` and do
```bash
export PYTHON_SETUP_DEPENDENCIES=/tmp/PySocks-1.7.1-py3-none-any.whl
```
then `pre-commit` works like normal.

I expect there are better solutions to this problem, but I thought a PR is a good way to get the ball rolling.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
     -   id: name-tests-test
     -   id: requirements-txt-fixer
     -   id: double-quote-string-fixer
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://gitlab.com/pycqa/flake8.git
     rev: 3.8.4
     hooks:
     -   id: flake8
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -197,10 +197,17 @@ def install_environment(
     if python is not None:
         venv_cmd.extend(('-p', python))
     install_cmd = ('python', '-mpip', 'install', '.', *additional_dependencies)
+    setup_dependencies_env = os.environ.get('PYTHON_SETUP_DEPENDENCIES')
+    if setup_dependencies_env:
+        setup_dependencies = setup_dependencies_env.split(':')
+    else:
+        setup_dependencies = []
 
     with clean_path_on_failure(envdir):
         cmd_output_b(*venv_cmd, cwd='/')
         with in_env(prefix, version):
+            for dep in setup_dependencies:
+                cmd_output_b('python', '-mpip', 'install', dep)
             helpers.run_setup_cmd(prefix, install_cmd)
 
 


AllSeeingEyeTolledEweSew:
Also: the change for the `flake8` repo url is due to the fact that for some reason, gitlab doesn't like cloning a "project page" url via tor; it wants the real git url.

```bash
amnesia@amnesia:~$ git clone https://gitlab.com/pycqa/flake8
Cloning into 'flake8'...
fatal: unable to access 'https://gitlab.com/pycqa/flake8/': The requested URL returned error: 403
amnesia@amnesia:~$ git clone https://gitlab.com/pycqa/flake8.git
Cloning into 'flake8'...
remote: Enumerating objects: 308, done.
remote: Counting objects: 100% (308/308), done.
remote: Compressing objects: 100% (168/168), done.
remote: Total 9637 (delta 161), reused 256 (delta 119), pack-reused 9329
Receiving objects: 100% (9637/9637), 2.55 MiB | 452.00 KiB/s, done.
Resolving deltas: 100% (6116/6116), done.
```

asottile:
this isn't going to be acceptable, sorry -- you should create a custom `virtualenv` seeder and do it that way

asottile:
see the discussion in #757

AllSeeingEyeTolledEweSew:
At first glance, I don't see a way to write a virtualenv plugin that could override the default seeder. Looks like it always chooses `"app-data"` unless CLI args say otherwise.

Looks like pre-commit only invokes virtualenv in a single place without arguments: https://github.com/pre-commit/pre-commit/blob/59acc7e48a58de3771d7552c64104add7e3dd1bf/pre_commit/languages/python.py#L195-L198

Does this mean you'd accept a PR to pass args to virtualenv? Do you have any guidance on how I should do that?

asottile:
you can configure `virtualenv` using `VIRTUALENV_` environment variables

PR #1707, opened 2020-11-21
added warning if globs are used instead of regex

### LINKED ISSUE #1702: Add warning for `files` / `exclude` containing `/*`
it's a common mistake to believe that those fields are globs, they're not -- they're regular expressions. (and matching `/*` is nonsense, it matches `''`, `'/'` as well as `'//////////////////////////////////////////////////////'`

paulhfischer:
resolves #1702

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -112,6 +112,18 @@ LOCAL = 'local'
 META = 'meta'
 
 
+class OptionalSensibleRegex(cfgv.OptionalNoDefault):
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if '/*' in dct.get(self.key, ''):
+            logger.warning(
+                f'The {self.key!r} field in hook {dct.get("id")!r} is a '
+                f"regex, not a glob -- matching '/*' probably isn't what you "
+                f'want here',
+            )
+
+
 class MigrateShaToRev:
     key = 'rev'
 
@@ -227,6 +239,8 @@ CONFIG_HOOK_DICT = cfgv.Map(
         for item in MANIFEST_HOOK_DICT.items
         if item.key != 'id'
     ),
+    OptionalSensibleRegex('files', cfgv.check_string),
+    OptionalSensibleRegex('exclude', cfgv.check_string),
 )
 CONFIG_REPO_DICT = cfgv.Map(
     'Repository', 'repo',
--- a/tests/clientlib_test.py
+++ b/tests/clientlib_test.py
@@ -166,6 +166,23 @@ def test_validate_warn_on_unknown_keys_at_top_level(tmpdir, caplog):
     ]
 
 
+def test_validate_optional_sensible_regex(caplog):
+    config_obj = {
+        'id': 'flake8',
+        'files': 'dir/*.py',
+    }
+    cfgv.validate(config_obj, CONFIG_HOOK_DICT)
+
+    assert caplog.record_tuples == [
+        (
+            'pre_commit',
+            logging.WARNING,
+            "The 'files' field in hook 'flake8' is a regex, not a glob -- "
+            "matching '/*' probably isn't what you want here",
+        ),
+    ]
+
+
 @pytest.mark.parametrize('fn', (validate_config_main, validate_manifest_main))
 def test_mains_not_ok(tmpdir, fn):
     not_yaml = tmpdir.join('f.notyaml')


asottile:
it probably makes more sense in pre_commit/clientlib.py? you can probably add a specific validator for this into the `cfgv` schema for the hook objects

paulhfischer:
thanks for the hint, i added a [validator](https://github.com/asottile/cfgv/#writing-your-own-validator) (+ test) to`pre_commit/clientlib.py` that is applied on config hooks (`CONFIG_HOOK_DICT`). let me know if you think this is fine or if i should change something.

pre_commit/clientlib.py
```diff
@@ -112,6 +112,25 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
 META = 'meta'
 
 
+class OptionalSensibleRegex:
+    def __init__(self, key: str, check_fn: Any) -> None:
+        self.key = key
+        self.check_fn = check_fn
```
asottile, line 118:
```suggestion
class OptionalSensibleRegex(NamedTuple):
    key: str
```

I don't think `check_fn` is needed here

pre_commit/clientlib.py
```diff
@@ -112,6 +112,25 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+
+    def check(self, dct: Dict[str, Any]) -> None:
+        cfgv.Optional.check(self, dct)
+
+        if '/*' in dct.get(self.key, ''):
+            logger.warning(
+                f'The {self.key!r} field in {dct.get("id")!r} is regex, not '
+                'globs -- matching \'/*\' probably isn\'t what you want here',
```
asottile, line 126:
```suggestion
                f'The {self.key!r} field in hook {dct.get("id")!r} is a regex, not a '
                f"glob -- matching '/*' probably isn't what you want here",
```

pre_commit/clientlib.py
```diff
@@ -217,6 +236,8 @@ def warn_unknown_keys_repo(
     'Hook', 'id',
 
     cfgv.Required('id', cfgv.check_string),
+    OptionalSensibleRegex('files', check_string_regex),
+    OptionalSensibleRegex('exclude', check_string_regex),
```
asottile, line 240: these'll need changing with the changes above

pre_commit/clientlib.py
```diff
@@ -225,7 +246,7 @@ def warn_unknown_keys_repo(
     *(
         cfgv.OptionalNoDefault(item.key, item.check_fn)
         for item in MANIFEST_HOOK_DICT.items
-        if item.key != 'id'
+        if item.key not in ['id', 'files', 'exclude']
```
asottile, line 249: these should still get applied -- there's no harm in checking a key more than once

just a tip though for python, you'd be better to write `if item.key not in {'id', 'files', 'exclude'}` (as a set) than as a list

tests/clientlib_test.py
```diff
@@ -166,6 +166,28 @@ def test_validate_warn_on_unknown_keys_at_top_level(tmpdir, caplog):
...
+        '        files: dir/*.py\n',
+    )
+    ret_val = validate_config_main((f.strpath,))
+    assert not ret_val
+    assert caplog.record_tuples == [
+        (
+            'pre_commit',
+            logging.WARNING,
+            'The \'files\' field in \'flake8\' is regex, not globs -- '
+            'matching \'/*\' probably isn\'t what you want here',
```
asottile, line 186:
```suggestion
            "The 'files' field in hook 'flake8' is a regex, not a glob -- "
            "matching '/*' probably isn't what you want here",
```

asottile (commented):
looks great!  just a few minor improvements

tests/clientlib_test.py
```diff
@@ -166,6 +166,28 @@ def test_validate_warn_on_unknown_keys_at_top_level(tmpdir, caplog):
...
+    f = tmpdir.join('cfg.yaml')
+    f.write(
+        'repos:\n'
+        '-   repo: https://gitlab.com/pycqa/flake8\n'
+        '    rev: 3.7.7\n'
+        '    hooks:\n'
+        '    -   id: flake8\n'
+        '        files: dir/*.py\n',
+    )
+    ret_val = validate_config_main((f.strpath,))
```
asottile, line 179: oh one other thing -- you can call the validate function directly `cfgv.validate({...}, CONFIG_HOOK_DICT)` without needing to write a file out

paulhfischer: thanks for the note! just out of interest: is there any particular reason why e.g. `test_validate_warn_on_unknown_keys_at_{top|repo}_level` are writing to a temp file?

pre_commit/clientlib.py
```diff
@@ -112,6 +112,25 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+
+        if '/*' in dct.get(self.key, ''):
+            logger.warning(
+                f'The {self.key!r} field in {dct.get("id")!r} is regex, not '
+                'globs -- matching \'/*\' probably isn\'t what you want here',
+            )
+
+    apply_default = cfgv.OptionalNoDefault.apply_default
+
+    remove_default = cfgv.OptionalNoDefault.remove_default
```
paulhfischer, line 131: also had to change this part as the class now inherits from `namedtuple`

pre_commit/clientlib.py
paulhfischer: i kept the `__init__` function as it caused problems with `apply_default`/`remove_default`:

```console
TypeError: _dct_noop() missing 1 required positional argument: 'dct'
```

asottile (approved):

tests/clientlib_test.py
asottile: most likely they were older tests / I didn't notice them when reviewing the patches

pre_commit/clientlib.py
asottile: ah, ok, I ended up adjusting this slightly! no problem!

asottile:
@paulhfischer do you want to implement this for top-level `files` / `exclude` as well?  I notice it's missing there :)

paulhfischer:
ah that's a good idea! i'll add it :)

PR #1709, opened 2020-11-22
fixed message if repo couldn't be updated due to missing hook(s)

### LINKED ISSUE #1708: running `pre-commit autoupdate` fails because tip of HEAD is missing hook
Hello 👋 
I'm setting up `pre-commit` on a project and came across an issue when adding hook `destroyed-symlinks`. The error message suggested running `pre-commit autoupdate`. I ran that and saw that it cannot update because the tip of HEAD is missing that hook. I'm not sure what that means so posting here.

```console
$ echo '    -   id: destroyed-symlinks' >> .pre-commit-config.yaml
$ git add -p !$
git add -p .pre-commit-config.yaml
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index bfde4717..949f3ffc 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,3 +21,4 @@ repos:
     -   id: check-vcs-permalinks
     -   id: check-xml
     -   id: debug-statements
+    -   id: destroyed-symlinks
(1/1) Stage this hunk [y,n,q,a,d,e,?]? y

$ git commit -m 'new hook destroyed-symlinks'
[ERROR] `destroyed-symlinks` is not present in repository https://github.com/pre-commit/pre-commit-hooks.  Typo? Perhaps it is introduced in a newer version?  Often `pre-commit autoupdate` fixes this.
$ git status
On branch pre-commit
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   .pre-commit-config.yaml

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        tests/__init__.py

$ pre-commit autoupdate
Updating https://github.com/pre-commit/pre-commit-hooks ... [INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
Cannot update because the tip of HEAD is missing these hooks:
destroyed-symlinks
$ git checkout .
Updated 0 paths from the index
$ pre-commit autoupdate
Updating https://github.com/pre-commit/pre-commit-hooks ... Cannot update because the tip of HEAD is missing these hooks:
destroyed-symlinks
$ pre-commit --version
pre-commit 2.9.0
```

paulhfischer:
should resolve #1708.

### DIFF
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -79,8 +79,8 @@ def _check_hooks_still_exist_at_rev(
     hooks_missing = hooks - {hook['id'] for hook in manifest}
     if hooks_missing:
         raise RepositoryCannotBeUpdatedError(
-            f'Cannot update because the tip of HEAD is missing these hooks:\n'
-            f'{", ".join(sorted(hooks_missing))}',
+            f'Cannot update because the update target is missing these '
+            f'hooks:\n{", ".join(sorted(hooks_missing))}',
         )
 
 


pre_commit/commands/autoupdate.py
```diff
@@ -79,8 +79,10 @@ def _check_hooks_still_exist_at_rev(
     hooks_missing = hooks - {hook['id'] for hook in manifest}
     if hooks_missing:
         raise RepositoryCannotBeUpdatedError(
-            f'Cannot update because the tip of HEAD is missing these hooks:\n'
-            f'{", ".join(sorted(hooks_missing))}',
+            f'Cannot update because the latest tag is missing these hooks:\n'
```
asottile, line 82: hmmm but this can also show up when using `--bleeding-edge` so this is subtly wrong too 🤔 

how about "tip of HEAD" -> "update target" and not adding the new sentence

paulhfischer: ah, you're correct -- didn't think about that

asottile (approved):

PR #1715, opened 2020-11-25
added warning if mutable rev is used

### LINKED ISSUE #974: issue a warning if a mutable ref is used (`HEAD` / `stable` / `master` / etc.)
Perhaps nudging towards https://pre-commit.com/#using-the-latest-version-for-a-repository or `pre-commit autoupdate`

paulhfischer:
resolves #974

- only added a warning instead of raising an error because this might break multiple projects otherwise
- not sure if other mutable revs should also cause a warning, couldn't think of a way to detect them in general …
- the check could also be included in the `MigrateShaToRev`-class, but i think it's better to keep them separated
- test only checks for `HEAD`, this is sufficient imo, but i can also change it so it tests for all mutable revs defined in `{'HEAD', 'stable', 'master'}`

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -1,6 +1,7 @@
 import argparse
 import functools
 import logging
+import re
 import shlex
 import sys
 from typing import Any
@@ -110,6 +111,25 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
 
 LOCAL = 'local'
 META = 'meta'
+
+
+# should inherit from cfgv.Conditional if sha support is dropped
+class WarnMutableRev(cfgv.ConditionalOptional):
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if self.key in dct:
+            rev = dct[self.key]
+
+            if '.' not in rev and not re.match(r'^[a-fA-F0-9]+$', rev):
+                logger.warning(
+                    f'The {self.key!r} field of repo {dct["repo"]!r} '
+                    f'appears to be a mutable reference '
+                    f'(moving tag / branch).  Mutable references are never '
+                    f'updated after first install and are not supported.  '
+                    f'See https://pre-commit.com/#using-the-latest-version-for-a-repository '  # noqa: E501
+                    f'for more details.',
+                )
 
 
 class OptionalSensibleRegex(cfgv.OptionalNoDefault):
@@ -261,6 +281,14 @@ CONFIG_REPO_DICT = cfgv.Map(
     ),
 
     MigrateShaToRev(),
+    WarnMutableRev(
+        'rev',
+        cfgv.check_string,
+        '',
+        'repo',
+        cfgv.NotIn(LOCAL, META),
+        True,
+    ),
     cfgv.WarnAdditionalKeys(('repo', 'rev', 'hooks'), warn_unknown_keys_repo),
 )
 DEFAULT_LANGUAGE_VERSION = cfgv.Map(
--- a/tests/clientlib_test.py
+++ b/tests/clientlib_test.py
@@ -180,6 +180,70 @@ def test_ci_key_must_be_map():
         cfgv.validate({'ci': 'invalid', 'repos': []}, CONFIG_SCHEMA)
 
 
+@pytest.mark.parametrize(
+    'rev',
+    (
+        'v0.12.4',
+        'b27f281',
+        'b27f281eb9398fc8504415d7fbdabf119ea8c5e1',
+        '19.10b0',
+        '4.3.21-2',
+    ),
+)
+def test_warn_mutable_rev_ok(caplog, rev):
+    config_obj = {
+        'repo': 'https://gitlab.com/pycqa/flake8',
+        'rev': rev,
+        'hooks': [{'id': 'flake8'}],
+    }
+    cfgv.validate(config_obj, CONFIG_REPO_DICT)
+
+    assert caplog.record_tuples == []
+
+
+@pytest.mark.parametrize(
+    'rev',
+    (
+        '',
+        'HEAD',
+        'stable',
+        'master',
+        'some_branch_name',
+    ),
+)
+def test_warn_mutable_rev_invalid(caplog, rev):
+    config_obj = {
+        'repo': 'https://gitlab.com/pycqa/flake8',
+        'rev': rev,
+        'hooks': [{'id': 'flake8'}],
+    }
+    cfgv.validate(config_obj, CONFIG_REPO_DICT)
+
+    assert caplog.record_tuples == [
+        (
+            'pre_commit',
+            logging.WARNING,
+            "The 'rev' field of repo 'https://gitlab.com/pycqa/flake8' "
+            'appears to be a mutable reference (moving tag / branch).  '
+            'Mutable references are never updated after first install and are '
+            'not supported.  '
+            'See https://pre-commit.com/#using-the-latest-version-for-a-repository '  # noqa: E501
+            'for more details.',
+        ),
+    ]
+
+
+def test_warn_mutable_rev_conditional():
+    config_obj = {
+        'repo': 'meta',
+        'rev': '3.7.7',
+        'hooks': [{'id': 'flake8'}],
+    }
+
+    with pytest.raises(cfgv.ValidationError):
+        cfgv.validate(config_obj, CONFIG_REPO_DICT)
+
+
 def test_validate_optional_sensible_regex(caplog):
     config_obj = {
         'id': 'flake8',


pre_commit/clientlib.py
```diff
@@ -112,6 +112,24 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
 META = 'meta'
 
 
+class CheckMutableRev:
```
asottile, line 115: can you use the pattern I adjusted your code to in the other PR?  I believe this should be a `Conditional` similar to the one in `MigrateShaToRev` (which will eventually be deleted)

pre_commit/clientlib.py
```diff
@@ -112,6 +112,24 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+    key = 'rev'
+
+    def check(self, dct: Dict[str, Any]) -> None:
+        if any(
+            mutable in dct.get(self.key, '')
+            for mutable in {'HEAD', 'stable', 'master'}
```
asottile, line 121: I'm not sure this will catch all the cases, notably it's missing `''`, `'main'` at the very least

I wonder if there's an easier way to catch these -- maybe match anything that doesn't look like a version?

pre_commit/clientlib.py
```diff
@@ -112,6 +112,24 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+        ):
+            logger.warning(
+                f"The {self.key!r} field of repo {dct.get('repo')!r} "
+                f"shouldn't be mutable",
```
asottile, line 125: the error message should link to the documentation

pre_commit/clientlib.py
```diff
@@ -112,6 +112,24 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+class CheckMutableRev:
+    key = 'rev'
+
+    def check(self, dct: Dict[str, Any]) -> None:
+        if any(
+            mutable in dct.get(self.key, '')
+            for mutable in {'HEAD', 'stable', 'master'}
+        ):
+            logger.warning(
+                f"The {self.key!r} field of repo {dct.get('repo')!r} "
```
asottile, line 124: repo is required by this point

pre_commit/clientlib.py
```diff
@@ -112,6 +112,24 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+                f"shouldn't be mutable",
+            )
+
+    apply_default = cfgv.Required.apply_default
+
+    remove_default = cfgv.Required.remove_default
```
asottile, line 130: `rev` isn't `Required`, for example `repo: local` / `repo: meta` don't use it

paulhfischer: didn't think of that, changed it to `Conditional`

pre_commit/clientlib.py
paulhfischer: i didn't change it to inherit from `cfgv.Conditional` but took a similar approach as `MigrateShaToRev`, as an inheritance would require passing in the arguments `key`, `check_fn`, `condition_key` and `condition_value` which should be hard coded in `_cond` in my opinion

pre_commit/clientlib.py
paulhfischer: changed it to `dct['repo']` if that's what you referred to here

pre_commit/clientlib.py
paulhfischer: alternative idea (not added yet):

```python
rev = dct.get(self.key, "")

if (
    "." not in rev and (  # not a version
        len(rev) != 40 or  # full sha
        len(rev) != 7 or  # short sha
        len(rev) != 8 or  # gitlab lists 8 char long sha in the commit-list
    ) or
    not re.match(r"^[a-zA-Z0-9]+$", rev)  # contains other chars (e.g. "_", "-", …)
)
# print warning
```

asottile: Yeah that might be better -- I wouldn't validate length at all though -- you could have a 15 character "short sha" for example so that would lead to a false positive

It might be worth trying to implement some sort of heuristic and then searching github to find if things adhere to it?

pre_commit/clientlib.py
```diff
@@ -112,6 +112,34 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+        ):
+            logger.warning(
+                f'The {self.key!r} field of repo {dct["repo"]!r} '
+                f"shouldn't be mutable. See https://pre-commit.com/#using-the-"
+                f'latest-version-for-a-repository for further details.',
+            )
+
+    apply_default = cfgv.Conditional.apply_default
+
+    remove_default = cfgv.Conditional.remove_default
```
asottile, line 140: I'd still rather see this inherit from `Conditional` as I plan to delete `MigrateShaToRev` in pre-commit 3.x

pre_commit/clientlib.py
paulhfischer: so i took a look at current usages of `rev` in `.pre-commit-config.yaml`-files (via github api [`https://api.github.com/search/code?q=filename:%2Epre-commit-config%2Eyaml&page={page}`]) and extracted all revs field. here's what i got so far:

- we can detect shas in general with `r"^[a-z0-9]+$"`, as they all only contain **lowercase** letters and numbers. this will result in false positives for branches, tho.
- version tags are all messed up. most of them follow the pattern `r"^v?([0-9]+\.)+[0-9]$"`, but:
  - some (e.g. ["20.8b1"](https://github.com/psf/black/releases/tag/20.8b1)) contain letters
  - some (e.g. ["4.3.21-2"](https://github.com/PyCQA/isort/releases/tag/4.3.21-2)) contain dashes
  thus, the only reasonable way to detect tags - in my opinion - is, to whitelist all strings that contain dots. (e.g contains `r"[^\.]\.[^\.]"`)
- negative examples i found are:
  - `"stable"`
  - `"master"`
  - `""`

thus, this is my suggestion (tested it against the data i got from the api calls):
```python
bad_revs = {
    "stable",
    "master",
    "HEAD",
}
rev = dct.get(self.key, "")

if "." not in rev and (
    not re.match(r"^[a-z0-9]+$", rev)
    or any(bad_rev in rev for bad_rev in bad_revs)
    or not rev
):
    print(f"Warning: {rev}")
```

pre_commit/clientlib.py
paulhfischer: i had to change the inheritance to `ConditionalOptional`, due to the support of `sha` instead of `rev`:

```console
FAILED tests/commands/migrate_config_test.py::test_migrate_config_sha_to_rev - pre_commit.clientlib.InvalidConfigError:
```

also added a comment that the inheritance can (and should) be changed to `Conditional` if support for `sha` (e.g. deletion of `MigrateShaToRev`) is dropped.

tests/clientlib_test.py
```diff
@@ -166,6 +166,54 @@ def test_validate_warn_on_unknown_keys_at_top_level(tmpdir, caplog):
...
+
+    expected = [
+        (
+            'pre_commit',
+            logging.WARNING,
+            "The 'rev' field of repo 'https://gitlab.com/pycqa/flake8' "
+            "shouldn't be mutable. See https://pre-commit.com/#using-the-"
+            'latest-version-for-a-repository for further details.',
+        ),
+    ] if error else []
```
asottile, line 199: please don't put logic in tests, this should be two separate tests, one demonstrating positive cases and one demonstrating negatives -- as it is right now you've introduced a boolean trap via parametrize (a common antipattern)

pre_commit/clientlib.py
```diff
@@ -112,6 +113,32 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+class CheckBadRev(cfgv.ConditionalOptional):
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if self.key in dct.keys():
+            bad_revs = {
+                'stable',
+                'master',
+                'HEAD',
+            }
```
asottile, line 126: this should be a module constant if we keep it, I don't think we should though

pre_commit/clientlib.py
```diff
@@ -112,6 +113,32 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+            rev = dct[self.key]
+
+            if '.' not in rev and (
+                not re.match(r'^[a-z0-9]+$', rev) or
+                any(bad_rev in rev for bad_rev in bad_revs) or
+                not rev
```
asottile, line 132: the empty string should warn too, it's (accidentally) an alias for `HEAD`

pre_commit/clientlib.py
```diff
@@ -112,6 +113,32 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+            bad_revs = {
+                'stable',
+                'master',
+                'HEAD',
+            }
+            rev = dct[self.key]
+
+            if '.' not in rev and (
+                not re.match(r'^[a-z0-9]+$', rev) or
+                any(bad_rev in rev for bad_rev in bad_revs) or
```
asottile, line 131: I think if you delete this line it'll still function the same -- also these should do exact matching, `'HEAD'` in `rev` is nonsense

pre_commit/clientlib.py
```diff
@@ -112,6 +113,32 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
 META = 'meta'
 
 
+# should inherit from cfgv.Conditional if sha support is dropped
+class CheckBadRev(cfgv.ConditionalOptional):
```
asottile, line 117: `WarnMutableRev` is probably a better name

pre_commit/clientlib.py
```diff
@@ -112,6 +113,32 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if self.key in dct.keys():
```
asottile, line 121: `.keys()` should almost never be called -- there's only one valid use case that I've ever seen for it (set operations (`|` / `&`) against keys) and every other use can be replaced by removing `.keys()`, you want `if self.key in dct`

pre_commit/clientlib.py
```diff
@@ -112,6 +113,32 @@ def validate_manifest_main(argv: Optional[Sequence[str]] = None) -> int:
...
+            bad_revs = {
+                'stable',
+                'master',
+                'HEAD',
+            }
+            rev = dct[self.key]
+
+            if '.' not in rev and (
+                not re.match(r'^[a-z0-9]+$', rev) or
```
asottile, line 130: upper case shas are valid git revisions, this should be `a-f0-9A-F0-9`

pre_commit/clientlib.py
paulhfischer: isn't `not rev` creating a warning for empty strings?

pre_commit/clientlib.py
paulhfischer: good point! i forgot that letters are also restricted to `a-f` …

pre_commit/clientlib.py
paulhfischer: i also added a test case for an empty `rev` string

asottile: yeah I think you can remove the `or not rev` -- `'.'` is not contained in `''` and `''` does not match your pattern

tests/clientlib_test.py
```diff
@@ -166,6 +166,72 @@ def test_validate_warn_on_unknown_keys_at_top_level(tmpdir, caplog):
     ]
 
 
+@pytest.mark.parametrize(
+    ('rev'),
+    (
+        ('v0.12.4'),
```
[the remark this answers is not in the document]
paulhfischer: oh good to know! thanks for the note, i changed it to match the signature.

asottile (approved):

PR #1726, opened 2020-12-06
yamllint fixes

CodeBleu:
I use yamllint in vim and noticed that the 'sample-config' generates lint errors.   I know you have your own check-yaml hook configured and am not sure if there is a reason not to use the yamllint hook.


asottile:
because yamllint's opinions are wrong

Issue #1738, opened 2020-12-20
Installing additional_dependencies with `language: conda` fails due to disregarding channels


siboehm:
# Example:
flake8-action/.pre-commit-hooks.yaml
```yaml
-   id: flake8-conda
    name: flake8-conda
    description: This hook runs flake8.
    entry: flake8
    language: conda
    types: [python]
    require_serial: true
```
flake8-action/environment.yaml
```yaml
channels:
- conda-forge
dependencies:
- flake8=3.8.4
```

test_repo/.pre-commit-config.yaml
```yaml
repos:
  - repo: ../flake8-action
    rev: <COMMIT-SHA>
    hooks:
     - id: flake8-conda
       additional_dependencies: [flake8-docstrings, flake8-rst-docstrings]
```

Now running `pre-commit` in `test_repo` fails with: 
```
[INFO] Initializing environment for ../flake8-action.
[INFO] Initializing environment for ../flake8-action:flake8-docstrings,flake8-rst-docstrings.
[INFO] Installing environment for ../flake8-action.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/home/simon/miniconda3/bin/python', '/home/simon/miniconda3/condabin/conda', 'install', '-p', '/home/simon/.cache/pre-commit/repomsutw5a0/conda-default', 'flake8-docstrings', 'flake8-rst-docstrings')
return code: 1
expected return code: 0
stdout:
    Collecting package metadata (current_repodata.json): ...working... done
    Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
    Collecting package metadata (repodata.json): ...working... done
    Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
    
stderr:
    
    PackagesNotFoundError: The following packages are not available from current channels:
    
      - flake8-docstrings
      - flake8-rst-docstrings
    
    Current channels:
    
      - https://repo.anaconda.com/pkgs/main/linux-64
      - https://repo.anaconda.com/pkgs/main/noarch
      - https://repo.anaconda.com/pkgs/r/linux-64
      - https://repo.anaconda.com/pkgs/r/noarch
    
    To search for alternate channels that may provide the conda package you're
    looking for, navigate to
    
        https://anaconda.org
    
    and use the search bar at the top of the page.
    
    
    
Check the log at /home/simon/.cache/pre-commit/pre-commit.log
```
[...]

PR #1739, opened 2020-12-21
Use temp env file to install add_deps for conda

siboehm:
Regards #1738 

It's possible to install `conda-forge` dependencies though using `additional_dependencies: [-c, conda-forge, ...]`, but it's not documented, triggers two lengthy solves and looks hacky (`-c` is not a conda package). 

I've changed the code to create a temporary env.yml instead and use that to install the environment.
I've talked about it with @xhochy (who authored the original code) and we agreed that this approach is cleaner, plus it's faster too.

### DIFF
--- a/pre_commit/languages/conda.py
+++ b/pre_commit/languages/conda.py
@@ -1,5 +1,6 @@
 import contextlib
 import os
+from tempfile import NamedTemporaryFile
 from typing import Generator
 from typing import Sequence
 from typing import Tuple
@@ -14,6 +15,8 @@ from pre_commit.languages import helpers
 from pre_commit.prefix import Prefix
 from pre_commit.util import clean_path_on_failure
 from pre_commit.util import cmd_output_b
+from pre_commit.util import yaml_dump
+from pre_commit.util import yaml_load
 
 ENVIRONMENT_DIR = 'conda'
 get_default_version = helpers.basic_get_default_version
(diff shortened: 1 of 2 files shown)


pre_commit/languages/conda.py
```diff
@@ -59,16 +62,25 @@ def install_environment(
...
+                yaml_dump(env_yaml, stream=tmp_env_file)
+
             cmd_output_b(
-                'conda', 'install', '-p', env_dir, *additional_dependencies,
-                cwd=prefix.prefix_dir,
+                'conda', 'env', 'create', '-p', env_dir, '--file',
+                tmp_env_file.name, cwd=prefix.prefix_dir,
             )
+        finally:
+            os.remove(tmp_env_file.name)
```
xhochy, line 83: You will need to handle the case that the `NamedTemporaryFile` creation didn't work and the file / variable doesn't exist.

asottile:
no sorry, please use the thing that already works

xhochy:
@asottile Any chance to convince you otherwise? The `['-c', 'conda-forge']` additional dependency is quite a hack that only works because of the current implementation. Also saving a single solve decreases the hook setup time quite a lot as an additional benefit (this also reduces a bit of my temptation to make a PR that let's one use `mamba` instead of `conda` as the backing solver).

asottile:
it's not a hack, that's the intended way to add options to the install command (it's how, for example, you can add `--index-url` or other options to `pip install` without having to grow a ton of language-dependent settings or code fragile dependence on the configuration format of a particular package manager)

the extra solve isn't really a big deal, it only installs once and then gets cached -- it also sounds like a bug in conda (all of the things are listed, it should just solve them together)

PR #1743, opened 2020-12-21
Expose remote branch ref as an environment variable

surafelabebe:
The remote branch reference was exposed as an environment variable to be consumed by a pre-push hook, which makes git push decisions to remote branches with a certain name pattern.


pre_commit/commands/run.py
```diff
@@ -371,6 +371,11 @@ def run(
         environ['PRE_COMMIT_FROM_REF'] = args.from_ref
         environ['PRE_COMMIT_TO_REF'] = args.to_ref
 
+    # Exposes remote branch ref as an environment variable
+    # for pre-push hooks consumption
+    if args.remote_branch:
+        environ['PRE_COMMIT_REMOTE_REF'] = args.remote_branch
```
asottile, line 377: this should be combined with the block below -- the naming here is inconsistent as well

asottile (commented):
hmmm I wonder if it's easy to extend the existing tests to cover this change

pre_commit/commands/run.py
surafelabebe: I updated the code as recommended.

asottile (approved):

asottile:
seems I can't merge this because you made the pull request from a company account (this also seems to be why pre-commit.ci couldn't autofix your pull request):

```console
$ git push clumio-oss HEAD -f
ERROR: Permission to clumio-oss/pre-commit.git denied to asottile.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

I'm going to close this and then reopen another pull request and maintain your authorship there

PR #1750, opened 2021-01-01
extended warning if globs are used instead of regex to top level

paulhfischer:
extends #1702 (see [this](https://github.com/pre-commit/pre-commit/pull/1707#issuecomment-753351851) comment).

i used the same validator. to distinguish between a hook field and a top-level field i checked if `id` is present in the dictionary. i think this should be fine, but if you have a better idea just let me know :)

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -132,7 +132,7 @@ class WarnMutableRev(cfgv.ConditionalOptional):
                 )
 
 
-class OptionalSensibleRegex(cfgv.OptionalNoDefault):
+class OptionalSensibleRegexAtHook(cfgv.OptionalNoDefault):
     def check(self, dct: Dict[str, Any]) -> None:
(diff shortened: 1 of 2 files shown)


pre_commit/clientlib.py
```diff
@@ -137,11 +137,17 @@ def check(self, dct: Dict[str, Any]) -> None:
...
+                logger.warning(
+                    f'The {self.key!r} field in hook {dct.get("id")!r} is a '
+                    f"regex, not a glob -- matching '/*' probably isn't what "
+                    f'you want here',
+                )
+            else:
+                logger.warning(
+                    f'The top-level {self.key!r} field is a regex, not a glob '
+                    f"-- matching '/*' probably isn't what you want here",
+                )
```
asottile, line 150: I would probably have two separate classes -- this feels like over-DRYing

asottile (approved):

PR #1763, opened 2021-01-21
update pyyaml dependency to 5.4

shawalli:
This PR updates the minimum PyYAML version to 5.4, which remediates [CVE-2020-14343 (https://access.redhat.com/security/cve/cve-2020-14343), as noted in the [5.4 Changelog](https://github.com/yaml/pyyaml/blob/master/CHANGES#L14).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ install_requires =
     cfgv>=2.0.0
     identify>=1.0.0
     nodeenv>=0.11.1
-    pyyaml>=5.1
+    pyyaml>=5.4
     toml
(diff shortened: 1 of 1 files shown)


asottile:
we're not vulnerable to CVE-2020 so updating this dependency is unnecessary

shawalli:
Unfortunately though, the use of the CVE-included PyYAML causes code-scanning tools to ring alarm bells. Even if a project isn't using PyYAML, if it is using `pre-commit`, it can get included in CVE warnings.

asottile:
no, that's not how it works

PR #1779, opened 2021-01-29
install_uninstall: prefer shebang with only major version

chibby0ne:
Having a major.minor version of python in the shebangs, has the unwanted
side effect of requiring an uninstallation and then reinstallation
all the pre-commit hooks, once the system updates the python
installation to a newer minor version, since it will not find the
version which the shebangs specifies.

To make matters worse the uninstallation of the hooks has to be manual
since the script will not be able to run for the aforementioned reason.

This change makes the python symlink with just the major version as the
preferred choice over the major.minor version specified shebang.

Signed-off-by: Antonio Gutierrez <chibby0ne@gmail.com>

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -58,8 +58,8 @@ def shebang() -> str:
         py, _ = os.path.splitext(SYS_EXE)
     else:
         exe_choices = [
-            f'python{sys.version_info[0]}.{sys.version_info[1]}',
             f'python{sys.version_info[0]}',
+            f'python{sys.version_info[0]}.{sys.version_info[1]}',
         ]
         # avoid searching for bare `python` as it's likely to be python 2
         if SYS_EXE != 'python':


asottile:
the ordering here is intentional -- check the commit history and the associated patches

chibby0ne:
@asottile Thanks for the quick reply.

Ok, in that case, what's the best approach for having the installed commits hooks work when a newer version of python is released?

asottile:
re-run `pre-commit install` -- upgrading a system python (and ditching the old one) is a rather infrequent task

chibby0ne:
@asottile I can't believe I didn't think of that! That works :+1:, but with the error message at commit time:
```
/usr/bin/env: 'python3.8': No such file or directory
```
It wasn't entirely obvious that it had to do with the pre-commit hook tbh.

Also python is now moving from a 18 to 12 month cadence, so this sort of thing will happen 30% more often. [PEP 602 -- Annual Release Cycle for Python](https://www.python.org/dev/peps/pep-0602)

PR #1799, opened 2021-02-15
Support for R / renv as a language

lorenzwalthert:
This PR aims to add 2nd level support for R to pre-commit via the virtual environment manager [renv](https://rstudio.github.io/renv/), motivated by isolated dependency management and (hopefully) compatibility with pre-commit.ci for 2nd level languages as outlined in https://github.com/lorenzwalthert/precommit/issues/215. The implementation follows the recently contributed support for conda (#1232) for source code and tests as well as the implementation of the python virtualenv language.

Further considerations:

* ~Initial tests only run on Linux. I have to add other platforms as we go if this PR is generally approved.~
* ~I'd appreciate feedback to the questions in the source code comments, in particular on whether or not to remove `in_env()` as it seems not needed or whether it should be kept for consistency with other languages.~ Also, let me know if you want certain comments that explain stuff to be removed. Was not sure how much.
* If this gets through, we probably should add documentation for pre-commit.com.
* We might want to support file types for  R (https://pre-commit.com/#filtering-files-with-types).

Before this PR, I was not familiar with the pre-commit code base nor a very avid python user, hence sorry if there are obvious flaws in this PR.

### DIFF
--- a/pre_commit/languages/all.py
+++ b/pre_commit/languages/all.py
@@ -16,6 +16,7 @@ from pre_commit.languages import node
 from pre_commit.languages import perl
 from pre_commit.languages import pygrep
 from pre_commit.languages import python
+from pre_commit.languages import r
 from pre_commit.languages import ruby
 from pre_commit.languages import rust
 from pre_commit.languages import script
@@ -52,6 +53,7 @@ languages = {
     'perl': Language(name='perl', ENVIRONMENT_DIR=perl.ENVIRONMENT_DIR, get_default_version=perl.get_default_version, healthy=perl.healthy, install_environment=perl.install_environment, run_hook=perl.run_hook),  # noqa: E501
     'pygrep': Language(name='pygrep', ENVIRONMENT_DIR=pygrep.ENVIRONMENT_DIR, get_default_version=pygrep.get_default_version, healthy=pygrep.healthy, install_environment=pygrep.install_environment, run_hook=pygrep.run_hook),  # noqa: E501
     'python': Language(name='python', ENVIRONMENT_DIR=python.ENVIRONMENT_DIR, get_default_version=python.get_default_version, healthy=python.healthy, install_environment=python.install_environment, run_hook=python.run_hook),  # noqa: E501
+    'r': Language(name='r', ENVIRONMENT_DIR=r.ENVIRONMENT_DIR, get_default_version=r.get_default_version, healthy=r.healthy, install_environment=r.install_environment, run_hook=r.run_hook),  # noqa: E501
     'ruby': Language(name='ruby', ENVIRONMENT_DIR=ruby.ENVIRONMENT_DIR, get_default_version=ruby.get_default_version, healthy=ruby.healthy, install_environment=ruby.install_environment, run_hook=ruby.run_hook),  # noqa: E501
     'rust': Language(name='rust', ENVIRONMENT_DIR=rust.ENVIRONMENT_DIR, get_default_version=rust.get_default_version, healthy=rust.healthy, install_environment=rust.install_environment, run_hook=rust.run_hook),  # noqa: E501
     'script': Language(name='script', ENVIRONMENT_DIR=script.ENVIRONMENT_DIR, get_default_version=script.get_default_version, healthy=script.healthy, install_environment=script.install_environment, run_hook=script.run_hook),  # noqa: E501
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -279,6 +279,54 @@ def test_node_hook_with_npm_userconfig_set(tempdir_factory, store, tmpdir):
         test_run_a_node_hook(tempdir_factory, store)
 
 
+def test_r_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'r_hooks_repo',
+        'hello-world', [os.devnull],
+        b'Hello, World, from R!\n',
+    )
+
+
+def test_r_inline_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'r_hooks_repo',
+        'hello-world-inline', ['some-file'],
+        b'Hi-there, some-file, from R!\n',
+    )
+
+
+def test_r_with_additional_dependencies_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'r_hooks_repo',
+        'additional-deps', [os.devnull],
+        b'OK\n',
+        config_kwargs={
+            'hooks': [{
+                'id': 'additional-deps',
+                'additional_dependencies': ['cachem@1.0.4'],
+            }],
+        },
+    )
+
+
+def test_r_local_with_additional_dependencies_hook(store):
+    config = {
+        'repo': 'local',
+        'hooks': [{
+            'id': 'local-r',
+            'name': 'local-r',
+            'entry': 'Rscript -e',
+            'language': 'r',
+            'args': ['if (packageVersion("R6") == "2.1.3") cat("OK\n")'],
+            'additional_dependencies': ['R6@2.1.3'],
+        }],
+    }
+    hook = _get_hook(config, store, 'local-r')
+    ret, out = _hook_run(hook, (), color=False)
+    assert ret == 0
+    assert _norm_out(out) == b'OK\n'
+
+
 def test_run_a_ruby_hook(tempdir_factory, store):
     _test_hook_repo(
         tempdir_factory, store, 'ruby_hooks_repo',
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,6 +26,10 @@ jobs:
         Write-Host "##vso[task.prependpath]C:\Strawberry\perl\site\bin"
         Write-Host "##vso[task.prependpath]C:\Strawberry\c\bin"
       displayName: Add strawberry perl to PATH
+    - task: PowerShell@2
+      inputs:
+        filePath: "testing/get-r.ps1"
+      displayName: install R
 - template: job--python-tox.yml@asottile
   parameters:
     toxenvs: [py37]
@@ -42,6 +46,8 @@ jobs:
         testing/get-swift.sh
         echo '##vso[task.prependpath]/tmp/swift/usr/bin'
       displayName: install swift
+    - bash: testing/get-r.sh
+      displayName: install R
 - template: job--python-tox.yml@asottile
   parameters:
     toxenvs: [pypy3, py36, py37, py38, py39]
@@ -56,3 +62,5 @@ jobs:
         testing/get-swift.sh
         echo '##vso[task.prependpath]/tmp/swift/usr/bin'
       displayName: install swift
+    - bash: testing/get-r.sh
+      displayName: install R
--- a/pre_commit/languages/r.py
+++ b/pre_commit/languages/r.py
@@ -1,0 +1,141 @@
+import contextlib
+import os
+import shlex
+import shutil
+from typing import Generator
+from typing import Sequence
+from typing import Tuple
+
+from pre_commit.envcontext import envcontext
+from pre_commit.envcontext import PatchesT
+from pre_commit.hook import Hook
+from pre_commit.languages import helpers
+from pre_commit.prefix import Prefix
+from pre_commit.util import clean_path_on_failure
+from pre_commit.util import cmd_output_b
+
+ENVIRONMENT_DIR = 'renv'
+get_default_version = helpers.basic_get_default_version
+healthy = helpers.basic_healthy
+
+
+def get_env_patch(venv: str) -> PatchesT:
+    return (
+        ('R_PROFILE_USER', os.path.join(venv, 'activate.R')),
+    )
+
+
+@contextlib.contextmanager
+def in_env(
+        prefix: Prefix,
+        language_version: str,
+) -> Generator[None, None, None]:
+    envdir = _get_env_dir(prefix, language_version)
+    with envcontext(get_env_patch(envdir)):
+        yield
+
+
+def _get_env_dir(prefix: Prefix, version: str) -> str:
+    return prefix.path(helpers.environment_dir(ENVIRONMENT_DIR, version))
+
+
+def _prefix_if_file_entry(
+    entry: Sequence[str],
+    prefix: Prefix,
+) -> Sequence[str]:
+    if entry[1] == '-e':
+        return entry[1:]
+    else:
+        return (prefix.path(entry[1]),)
+
+
+def _entry_validate(entry: Sequence[str]) -> None:
+    """
+    Allowed entries:
+    # Rscript -e expr
+    # Rscript path/to/file
+    """
+    if entry[0] != 'Rscript':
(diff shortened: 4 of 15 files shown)


testing/get-renv.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+sudo apt install r-base
+# create empty folder for user library.
+# necessary for non-root users who have
+# never installed an R package before.
+# Alternatively, we require the renv
+# package to be installed already, then we can
+# omit that.
+Rscript -e 'dir.create(Sys.getenv("R_LIBS_USER"), recursive = TRUE)'
```
lorenzwalthert, line 9: This won't work if a non-root user has never installed any R package. Maybe catch the installation error and fail informatively?

pre_commit/languages/renv.py
```diff
@@ -0,0 +1,97 @@
...
+        file_args: Sequence[str],
+        color: bool,
+) -> Tuple[int, bytes]:
+    with in_env(hook.prefix, hook.language_version):
+        # activates renv through .Rprofile in
+        # hook.prefix.prefix_dir
+        # expose entry to the user for maximal control
+        return helpers.run_xargs(
+            hook, hook.cmd, file_args,
+            color=color, cwd=hook.prefix.prefix_dir,
```
asottile, line 96: the current expectation for hook authors is that hooks will be executed from the root of the repository, so `cwd=...` won't be workable here

I think that necessarily changes how the environment setup is working now -- the ~normalish way of setting this up is by pointing some set of environment variables at the ENVIRONMENT_DIR (and subdirs)

note also that things must be installed entirely inside the ENVIRONMENT_DIR or they potentially won't be persisted by caching

lorenzwalthert: > the current expectation for hook authors is that hooks will be executed from the root of the repository, so cwd=... won't be workable here.

of course 🤦. 

>  the ~normalish way of setting this up is by pointing some set of environment variables at the ENVIRONMENT_DIR (and subdirs).

That's what I saw. Problem is (and that's a peculiarity of R) the most operations happen in an R session, not the terminal session. `renv` can only be activated properly from inside R. For that reason, my current suggestion is to place a `renv::activate()` call in a file, then point the `R_PROFILE_USER` environment variable to that file so the code will run at [R startup](https://rstats.wtf/r-startup.html). The idea is now to allow two versions, inline code expressions and file in `entry:`:
```bash
Rscript -e expr # like Rscript -e "1+1" 
Rscript path/to/file 
```
An inject the options `--no-save`, `--no-restore`, `--no-site-file` and `--no-environ` after `Rscript` (to basically have `--vanilla` start, but run this one `renv::activate()` statement) and expand the path similar to `language: script` since we have to run the hooks in the working directory of the repo. The files and the hook args are then passed after the above, i.e. 
```r
Rscript --no-save --no-restore --no-site-file --no-environ /abs/path/ [files] [hook args]
Rscript --no-save --no-restore --no-site-file --no-environ -e expr [files] [hook args]
```


> note also that things must be installed entirely inside the ENVIRONMENT_DIR or they potentially won't be persisted by caching.

This should be solved now. Just placing the whole env in a subdirectory now.

testing/resources/renv_hooks_repo/DESCRIPTION
```diff
@@ -0,0 +1,18 @@
+Package: gli.clu
```
lorenzwalthert, line 1: The `DESCRIPTION` file is basically the equivalent of `setup.py`, `renv.lock` is the equivalent of the conda `environment.yml`. Not sure if one or both should be listed in `LOCAL_RESOURCES` as this was introduced here:
https://github.com/pre-commit/pre-commit/pull/1232/files#diff-a52ef0cf627ce783529c259e63d4ea82b77520936a2455cbf4f90ad2fb36d27dR176

tests/repository_test.py
```diff
@@ -279,14 +280,123 @@ def test_node_hook_with_npm_userconfig_set(tempdir_factory, store, tmpdir):
         test_run_a_node_hook(tempdir_factory, store)
 
 
+def _test_renv_parsing(
```
lorenzwalthert, line 283: let me know if these are too many tests (because it slows down testing). I think we can get 100% coverage with fewer tests too (but then not cover as many combinations).

pre_commit/languages/all.py
```diff
@@ -16,6 +16,7 @@
 from pre_commit.languages import perl
 from pre_commit.languages import pygrep
 from pre_commit.languages import python
+from pre_commit.languages import renv
```
asottile, line 19: I think this should maybe just be called `r`?

pre_commit/languages/renv.py
```diff
@@ -0,0 +1,138 @@
...
+
+def run_hook(
+        hook: Hook,
+        file_args: Sequence[str],
+        color: bool,
+) -> Tuple[int, bytes]:
+    with in_env(hook.prefix, hook.language_version):
+
+        return helpers.run_xargs(
+            hook, cmd_from_hook(hook), file_args, color=color,
```
asottile, line 91: I usually prefer to read a file from the bottom upwards -- also helper functions that live only in this module should be underscore-prefixed to make it more clear what is/isn't part of the api (the api specification is in the `languages/all.py` file)

pre_commit/languages/renv.py
```diff
@@ -0,0 +1,138 @@
...
+
+def entry_validate(entry: Sequence[str]) -> None:
+    """
+    # allowed entries:
+    # Rscript -e expr
+    # Rscript path/to/file
+    # TODO what if expr is x;2? not captured by shlex.split()
+    """
+    if entry[0] != 'Rscript':
+        raise SyntaxError('entry must start with `Rscript`.')
```
asottile, line 115: SyntaxError is for python syntax being incorrect -- these should be `ValueError` probably

pre_commit/languages/renv.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if entry[1] == '-e':
+        if len(entry) > 3:
+            raise SyntaxError('You can supply at most one expression.')
+    elif len(entry) > 2:
+        raise SyntaxError(
+            ' '.join([
+                'The only valid syntax is `Rscript -e {expr}`',
+                'or `Rscript path/to/hook/script`',
+            ]),
```
asottile, line 125: this can be written as an implicitly joined string:

```
raise TypeError(
    'first line here '
    'second line here'
)
```

pre_commit/languages/renv.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+
+    None
```
asottile, line 128: this line does nothing and can be removed, note that python implicitly returns `None` for "void" functions and doesn't support implicit-return-without-`return`-keyword

pre_commit/languages/renv.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def prefix_if_file_entry(
+    entry: Sequence[str],
+    prefix: Prefix,
+) -> Sequence[str]:
+    if entry[1] == '-e':
+        return(entry[1:])
```
asottile, line 136: note that `return` isn't a function, so no need to put parens here -- also usually put a space after `return` -- I eventually want to fix `pycodestyle` to point this out directly, but there's only so much I can work on 😆

testing/resources/renv_hooks_repo/.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,48 @@
+# parsing file
+-   id: parse-file-no-opts-no-args
+    name: Say hi
+    entry: Rscript parse-file-no-opts-no-args.R
+    language: renv
+    files: \.[Rr]$
```
asottile, line 6: we should probably add `r` as a language to pre-commit/identify if it's not there already

testing/resources/renv_hooks_repo/DESCRIPTION
asottile: ah `LOCAL_RESOURCES` is to allow an `additional_dependencies`-only `repo: local` hook -- it seems like additional_dependencies is supported for R so we should probably make an empty one?

tests/repository_test.py
```diff
@@ -19,6 +19,7 @@
 from pre_commit.languages import helpers
 from pre_commit.languages import node
 from pre_commit.languages import python
+from pre_commit.languages import renv
```
asottile, line 22: the tests specifically for the renv helper functions should live in tests/languages/renv_test.py (or `r_test.py` after renaming)

tests/repository_test.py
```diff
@@ -279,14 +280,123 @@ def test_node_hook_with_npm_userconfig_set(tempdir_factory, store, tmpdir):
...
+        )
+
+
+def test_renv_parsing_expr_args_in_entry2(tempdir_factory, store):
+    with pytest.raises(SyntaxError, match='at most one'):
+        renv.entry_validate(['Rscript', '-e', 'expr1', '--another-arg'])
+
+
+def test_renv_parsing_expr_non_Rscirpt(tempdir_factory, store):
+
```
asottile, line 357: (throughout the review, nbd I can fix this myself later -- I don't put blank lines at the beginning or the ending of blocks)

tests/repository_test.py
```diff
@@ -279,14 +280,123 @@ def test_node_hook_with_npm_userconfig_set(tempdir_factory, store, tmpdir):
...
+    with pytest.raises(SyntaxError, match='entry must start with'):
```
asottile, line 358: I really don't like the `match=` arg and prefer to test the exact error message, you can follow this pattern:

```python
with pytest.raises(TypeHere) as excinfo:
    ...
msg, = excinfo.value.args
assert msg == ...
```

tests/repository_test.py
```diff
@@ -279,14 +280,123 @@ def test_node_hook_with_npm_userconfig_set(tempdir_factory, store, tmpdir):
...
+
 def test_run_a_ruby_hook(tempdir_factory, store):
     _test_hook_repo(
         tempdir_factory, store, 'ruby_hooks_repo',
         'ruby_hook', [os.devnull], b'Hello world from a ruby hook\n',
     )
 
 
-@xfailif_windows  # pragma: win32 no cover
+@ xfailif_windows  # pragma: win32 no cover
```
asottile, line 399: decorators shouldn't have spaces after them -- not sure what happened here

asottile (commented):
overall looks great, really happy you took this on and I'm excited to add this support!

pre_commit/languages/all.py
lorenzwalthert: Thought about that also. The thing is that `language: conda` can both host R and Python hooks and `renv` can [manage python dependencies](https://rstudio.github.io/renv/articles/python.html), so I thought I go for a more specific name. On the other hand,  `language: python` uses `venv` (and is still called `python`, not `venv`). Since working with virtual environments is still not as common in R compared to Python and not every R user knows `renv`, using `r` as a name might also make it easier for people to find what they are looking for. So I am happy to switch to `r`, please let me know what you prefer.

pre_commit/languages/renv.py
lorenzwalthert: Ok, I can adapt the order. I can also prefix the helper functions not part of the API. But why is `in_env` and `get_env_patch` not prefixed in other modules although not listed in `languages/all.py`?

testing/resources/renv_hooks_repo/.pre-commit-hooks.yaml
lorenzwalthert: Apparently it is there already: 
```python
identify.tags_from_path('testing/resources/renv_hooks_repo/hello-world.R')
#> {'r', 'non-executable', 'text', 'file'}
```
So I can use this in that unit test. However, there are other R file types we want to support in some way in `identify` probably: `.Rprofile`, `.Rmd`, `.Rmarkdown` and `Rscript` shebang. I think I'd like to think about that later, and check with https://github.com/lorenzwalthert/precommit which ones could be useful.

testing/resources/renv_hooks_repo/DESCRIPTION
lorenzwalthert: Ok, thanks for the explanation. I added a template for `renv.lock` in cf88089e717c766afb456b5f4737bd73dbb90c24. However, there is an R version hard-coded in it which will get outdated. As far as I can tell from a quick experiment, the docs and issues in https://github.com/rstudio/renv, there is no indication that the R version recorded in the lock file will be used at `renv::restore()`, which we use in `renv.install_environment()`. Even removing seems fine (I left it there anyways). But that might still be a bit brittle.

lorenzwalthert: Also, should we add a unit test for a local hook with additional dependencies to check if things work as expected? I don't think that's currently covered.

tests/repository_test.py
lorenzwalthert: Also wondered, but one of the pre-commit hook changed it, maybe autopep? I even rebased my PR master and tried to get rid of this. Can revert but then not sure if the pre-commit.ci passes.

pre_commit/languages/all.py
asottile: yeah I think `r` makes the most sense, `conda` is kinda the special case right now and I think if people google "pre-commit r" they'll be much more likely to hit this than if it were named `renv`

we can always add other names if new ways of installing isolated r environments happen in the future and/or shuffle them around (there used to be a `python_venv` language which was merged into `python` as `virtualenv` makes `venv`-style environments anyway nowadays)

pre_commit/languages/renv.py
asottile: hmmmm yeah good point, those should probably be underscore-prefixed everywhere -- probably just legacy and copy paste 😅

testing/resources/renv_hooks_repo/.pre-commit-hooks.yaml
asottile: cool cool -- the place to do that is in https://github.com/pre-commit/identify there's an `extensions.py` file where you can add those!

tests/repository_test.py
asottile: ah my guess is there was a syntax error earlier in the file and then autopep8 freaked -- I've been meaning to fix that in autopep8 but it's tricky (basically sometimes it will try and rewrite a file even if it has a syntax error 😱)

testing/resources/renv_hooks_repo/DESCRIPTION
asottile: yeah it would probably be good to have one unit test demoing a `local` hook via `additional_dependencies`

asottile (approved):

asottile:
added a stub for this in the docs -- https://github.com/pre-commit/pre-commit.com/pull/473

lorenzwalthert:
Thanks. I was still thinking about this, maybe we should change some more things here later, in particular:

* use the boostrap activation script created with `renv::init()` so the R package `{renv}` won't be installed in the global library if it's not there already. 
* appending the content of the newly created `activate.R` to `renv/activate.R` and ask hook authors to commit not just the `renv.lock` but also the [other recommended files](https://rstudio.github.io/renv/articles/collaborating.html).

PR #1800, opened 2021-02-16
Add support for post-merge hooks

psacawa:
- Add: post-merge hook support
- Add post-merge tests

See commit message.

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -76,6 +76,7 @@ def _ns(
         remote_url: Optional[str] = None,
         commit_msg_filename: Optional[str] = None,
         checkout_type: Optional[str] = None,
+        is_squash_merge: Optional[str] = None,
 ) -> argparse.Namespace:
     return argparse.Namespace(
         color=color,
@@ -88,6 +89,7 @@ def _ns(
         commit_msg_filename=commit_msg_filename,
         all_files=all_files,
         checkout_type=checkout_type,
+        is_squash_merge=is_squash_merge,
         files=(),
         hook=None,
         verbose=False,
@@ -158,6 +160,7 @@ _EXPECTED_ARG_LENGTH_BY_HOOK = {
     'post-commit': 0,
     'pre-commit': 0,
     'pre-merge-commit': 0,
+    'post-merge': 1,
     'pre-push': 2,
 }
 
@@ -199,6 +202,8 @@ def _run_ns(
             hook_type, color,
             from_ref=args[0], to_ref=args[1], checkout_type=args[2],
         )
+    elif hook_type == 'post-merge':
+        return _ns(hook_type, color, is_squash_merge=args[0])
     else:
         raise AssertionError(f'unexpected hook type: {hook_type}')
 
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -67,8 +67,8 @@ class AppendReplaceDefault(argparse.Action):
 def _add_hook_type_option(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         '-t', '--hook-type', choices=(
-            'pre-commit', 'pre-merge-commit', 'pre-push',
-            'prepare-commit-msg', 'commit-msg', 'post-commit', 'post-checkout',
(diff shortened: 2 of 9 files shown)


pre_commit/commands/hook_impl.py
```diff
@@ -76,6 +76,7 @@ def _ns(
         remote_url: Optional[str] = None,
         commit_msg_filename: Optional[str] = None,
         checkout_type: Optional[str] = None,
+        is_squash_commit: Optional[str] = None,
```
asottile, line 79: I see this passed along, is this exposed to hooks in any way?  I think the current examples for other hooks expose these as environment variables -- probably something like `PRE_COMMIT_IS_SQUASH_COMMIT`?

psacawa: An environmental variable is added, as is a `precommit run` flag. This breaks some tests

pre_commit/main.py
```diff
@@ -136,6 +136,15 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
             'file from the index, flag=0).'
         ),
     )
+    parser.add_argument(
+        '--is-squash-merge',
+        action='store_const',
+        const='1',
```
asottile, line 142: I think this should probably be `'--squash-merge', default='', action='store_const', const='1'`

though it might make sense to make this a boolean?  and then we'd parse the incoming value from git in the `hook-impl` command (maybe `== '1'`? does git specify what values it'll send here?)

tests/commands/hook_impl_test.py
```diff
@@ -97,6 +97,7 @@ def call(*_, **__):
         ('pre-push', ['branch_name', 'remote_name']),
         ('commit-msg', ['.git/COMMIT_EDITMSG']),
         ('post-commit', []),
+        ('post-merge', [True]),
```
asottile, line 100: this doesn't match the arguments that git will send (just to make this realistic, these should all be strings)

tests/commands/install_uninstall_test.py
```diff
@@ -759,6 +759,45 @@ def test_post_commit_integration(tempdir_factory, store):
...
+        open('master', 'a').close()
+        cmd_output('git', 'add', '.')
+        git_commit()
+
+        cmd_output('git', 'checkout', '-b', 'branch', 'HEAD^')
+        open('branch', 'a').close()
+        cmd_output('git', 'add', '.')
+        git_commit()
+
+        cmd_output('git', 'switch', 'master')
```
asottile, line 794: I know `git switch` is the nice new way of doing things, but there's realistically some packagers that use old git 2.x that don't have this yet -- could you use `git checkout -b` instead?

tests/commands/install_uninstall_test.py
```diff
@@ -759,6 +759,45 @@ def test_post_commit_integration(tempdir_factory, store):
...
+        install(C.CONFIG_FILE, store, hook_types=['post-merge'])
+        retc, stdout, stderr = cmd_output('git', 'merge', 'branch')
```
asottile, line 796: this needs to use the special helper that the others use, otherwise it will write to the programmer's home directory under test: `cmd_output_mocked_pre_commit_home`

asottile (commented):
otherwise looks good -- thanks for the work on this!

recepsirin (approved):

pre_commit/main.py
psacawa: Git sends `'0'` and  `'1'` for `post-merge` hooks. This is true for `post-checkout` as well, with the `--checkout-type` argument. The existing `checkout-type` flag does not supply a default, though plausibly it should default to `'1'`. I will use your suggestion, except with default `'0'`

tests/commands/install_uninstall_test.py
psacawa: Used git checkout, as I think this is what you meant.

pre_commit/main.py
asottile: ah, I ended up changing this to be more like the `--checkout-type` argument -- that way you can simulate both `=1` and `=0` -- as it was before I fixed it it was always setting the `PRE_COMMIT_IS_...` variable which was not quite right (since `'0'` is truthy)

asottile (approved):

PR #1812, opened 2021-02-23
Use exclusion syntax for git diff filtering

matthewhughes934:
Since git 1.8.5 [1], '--diff-filter' can take lowercase args to mean
"show everything but these classes". Use this to exclude deleted files
rather than specify every other class.

[1] https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/RelNotes/1.8.5.txt#n186

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -127,8 +127,7 @@ def get_staged_files(cwd: Optional[str] = None) -> List[str]:
     return zsplit(
         cmd_output(
             'git', 'diff', '--staged', '--name-only', '--no-ext-diff', '-z',
-            # Everything except for D
-            '--diff-filter=ACMRTUXB',
+            '--diff-filter=d',
             cwd=cwd,
         )[1],
     )


pre_commit/git.py
```diff
@@ -128,7 +128,7 @@ def get_staged_files(cwd: Optional[str] = None) -> List[str]:
         cmd_output(
             'git', 'diff', '--staged', '--name-only', '--no-ext-diff', '-z',
             # Everything except for D
```
asottile, line 130: you'll want to update and/or remove the comment

matthewhughes934: > you'll want to update and/or remove the comment

Removed f7cad33

asottile:
while the code does seem improved, this decreases supported git versions for ~basically no improvement.  we don't explicitly check for git versions because the performance of doing so adds an avoidable ~5-10ms to the critical path so reducing the versions here for a cosmetic change doesn't seem worth it

thanks nonetheless for the patch -- appreciate it!

PR #1818, opened 2021-02-25
Add support for Go 1.16

### LINKED ISSUE #1815: Add support for Go 1.16
Go 1.16 now rejects empty `go.mod` files (c.f. https://github.com/golang/go/commit/d4986e0e1d6e03e1b92e905ca5e01b4c223fbeb3). This means that `golang` hooks cannot be installed anymore:

```
$ pre-commit install-hooks
[INFO] Initializing environment for local.
[INFO] Installing environment for local.
An unexpected error has occurred: CalledProcessError: command: ('/usr/local/bin/go', 'get', './...')
return code: 1
expected return code: 0
stdout: (none)
stderr:
    go: no module declaration in go.mod. To specify the module path:
    	go mod edit -module=example.com/mod
    
Check the log at /Users/rdraoui/.cache/pre-commit/pre-commit.log
```

I could get it to work by changing `pre_commit/resources/empty_template_go.mod` from an empty file to:

```
module example/mod
```

But I'm not all that familiar with how modules work in different Go versions, so maybe making that change has some other implications that I don't know about. This [Go Wiki entry](https://github.com/golang/go/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file) seems to suggest that empty `go.mod` files are legal, but maybe it's outdated... (c.f: https://github.com/pre-commit/pre-commit/pull/1428#pullrequestreview-404547009)

The fix is compatible with Go 1.15.

I tried validating the fix by running the test suite (`pytest tests -k golang`), but I can't get it to pass with Go 1.15 and current `master` branch either.

rafikdraoui:
Go 1.16 changes the way modules are handled. It now expects Go projects to have non-empty `go.mod` files.

This change is compatible with Go 1.15.

Fixes #1815

---

The golang tests pass under all combinations of `Go [1.15|1.16]` + `GO111MODULE=[""|"on"|"auto"]`.

I could also install and run the following hook successfully with both Go 1.15 and Go 1.16:

```
- repo: local
    hooks:
      - id: shfmt
        name: Format shell files with shfmt
        language: golang
        additional_dependencies: ["mvdan.cc/sh/v3/cmd/shfmt@v3.2.1"]
        types: [shell]
        entry: shfmt -w
```

These changes work for my use cases, and they seem sensible to me, but I'm not all that familiar with Go modules, so perhaps this has some implications I don't know about.

### DIFF
--- a/pre_commit/resources/empty_template_go.mod
+++ b/pre_commit/resources/empty_template_go.mod
@@ -1,0 +1,1 @@
+module pre-commit-dummy-empty-module
--- a/testing/resources/golang_hooks_repo/go.mod
+++ b/testing/resources/golang_hooks_repo/go.mod
@@ -1,0 +1,1 @@
+module golang-hello-world


asottile (approved):
validated that this is the only reasonable fix -- even setting the remote properly does not convince go that it should build

PR #1831, opened 2021-03-08
Fix r hooks when hook repo is a package 

lorenzwalthert:
In #1799, we assume `env_dir` instead of `prefix_dir` as the package source and install it, which creates a package with the right name (because of `DESCRIPTION`), but obviously does not expose any API and is hence useless. Hooks that rely on the installed package hence fail. ~Also removing an [(apparently) randomly introduced `\`](https://github.com/pre-commit/pre-commit/compare/master...lorenzwalthert:renv-fix#diff-813c025754186e7d7af82275ff89a79dd4bce3f2e037190c30c2f5a1aa1c8a9dL97) that does not cause any harm.~

### DIFF
--- a/pre_commit/languages/r.py
+++ b/pre_commit/languages/r.py
@@ -88,13 +88,11 @@ def install_environment(
     env_dir = _get_env_dir(prefix, version)
     with clean_path_on_failure(env_dir):
         os.makedirs(env_dir, exist_ok=True)
-        path_desc_source = prefix.path('DESCRIPTION')
-        if os.path.exists(path_desc_source):
-            shutil.copy(path_desc_source, env_dir)
         shutil.copy(prefix.path('renv.lock'), env_dir)
         cmd_output_b(
             'Rscript', '--vanilla', '-e',
-            """\
+            f"""\
+            prefix_dir <- {prefix.prefix_dir!r}
             missing_pkgs <- setdiff(
                 "renv", unname(installed.packages()[, "Package"])
             )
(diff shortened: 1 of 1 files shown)


pre_commit/languages/r.py
```diff
@@ -88,13 +88,11 @@ def install_environment(
...
     with clean_path_on_failure(env_dir):
         os.makedirs(env_dir, exist_ok=True)
-        path_desc_source = prefix.path('DESCRIPTION')
-        if os.path.exists(path_desc_source):
-            shutil.copy(path_desc_source, env_dir)
         shutil.copy(prefix.path('renv.lock'), env_dir)
         cmd_output_b(
             'Rscript', '--vanilla', '-e',
-            """\
+            "prefix_dir <- '" + prefix.prefix_dir + "'" +
```
asottile, line 94: this should use an f-string to do formatting, also triple-quoted literals can contain quote characters

the backslash was intentional, actually -- it makes the line numbers of the string match up with the string (and avoids a blank line at the top of your string)

lorenzwalthert: Thanks, makes all sense. Was too lazy to check if `python < 3.6` is required to make f strings work... should be fixed now. Also, because the `prefix_dir` is passed into string expression (`Rscript -e {expr}`), another escaping is required for Windows, as it has backward slashes in the path. Maybe there is a better way to do this but I could not find one...

pre_commit/languages/r.py
```diff
@@ -88,13 +88,13 @@ def install_environment(
...
-        if os.path.exists(path_desc_source):
-            shutil.copy(path_desc_source, env_dir)
         shutil.copy(prefix.path('renv.lock'), env_dir)
         cmd_output_b(
             'Rscript', '--vanilla', '-e',
-            """\
+            f"""\
+            prefix_dir <- "{
+                prefix.prefix_dir.encode('unicode_escape').decode()
+            }"
```
asottile, line 97: I think  you can use `{prefix.prefix_dir!r}` to do this as well -- that'll give you the quotes and then you won't have to do an encoding dance

asottile (commented):
otherwise looks good -- just a small thing

pre_commit/languages/r.py
lorenzwalthert: Nice. I really start to like Python string manipulations 😀. All checks pass now.

asottile (approved):

PR #1832, opened 2021-03-09
chore: drop universal wheel

henryiii:
Pre-commit does not support Python 2 anymore (`python_requires = >=3.6.1`), so it does not need universal wheels activated, which produce `py3.py2` in the wheel name instead of `py3`.

You can see the "Universal" naming here: https://pypi.org/project/pre-commit/#files

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,9 +50,6 @@ exclude =
     tests*
     testing*
 
-[bdist_wheel]
-universal = True
-
 [coverage:run]
 plugins = covdefaults
 omit = pre_commit/resources/*


henryiii:
No it does not. See https://github.com/pypa/setuptools/issues/1974 when [Setuptools 45](https://setuptools.readthedocs.io/en/latest/history.html#v45-1-0) dropped universal=True from setuptools itself https://github.com/pypa/setuptools/pull/1978. Setuptools can tell if a project is pure Python; the universal=1 added `py2.py3` tag.

If I see something explicitly wrong, I generally immediately make a PR. I looked for a dupe PR, but not a dupe Issue.

henryiii:
I do not see an Issue discussing this either.

henryiii:
You will also notice that after setuptools 45 dropped Python 2 support all mention of the universal setting was excised from the documentation. If you don't have an Extension, it's pure Python, if you do, it's not. Universal was there just to activate Python 2 + Python 3 support, since it couldn't tell that without help (it predates `python_requires`).

layday:
The wheel documentation describes 'universal' as 'Python 2/3 compatible, pure Python' (https://github.com/pypa/wheel/blob/2e90eef2e52dab03cff1f71bcb8931ce141e320d/docs/quickstart.rst), the news entry from 2012 introducing 'universal' says, 'universal=1 will apply the py2.py3-none-any tag for pure python wheels' (https://github.com/pypa/wheel/blob/a6fa4839ba1c74a37cb43eefb959c0766594ec63/docs/news.rst) and the only thing the 'universal' tests actually test is that only one wheel is produced with the 'py2.py3' compatibility tag (https://github.com/pypa/wheel/blob/2e90eef2e52dab03cff1f71bcb8931ce141e320d/tests/test_tagopt.py#L86).

I also don't see how this apparent hostility is at all warranted against a contribution made in good faith.

asottile:
no hostility, just the crusade to remove this flag is [not correct](https://github.com/pypa/wheel/commit/4a1b292b73a20e71f3939decbbb771507da79ce7#diff-8d068e8797e88947c320f79e856c3e16a72b730124a8f9d7031e2c4680dfa534R110-R111) and I closed this change because it is not welcome here

Issue #1835, opened 2021-03-09
R hooks give misleading warning when {renv} package versions differ


lorenzwalthert:
The implementation in #1799 relies on a globally installed version of the `{renv}` R package (with 0 dependencies). When the version of this installed package is different from what is provided in `renv.lock` (which is managed by the hook author), a warning like the following is displayed when hooks are ran:

```
renv 0.13.0 was loaded from project library, but this project is configured to use renv 0.12.5.
Use `renv::record("renv@0.13.0")` to record renv 0.13.0 in the lockfile.
Use `renv::restore(packages = "renv")` to install renv 0.12.5 into the project library. 
```
This is very misleading since these commands had to be ran in `env_dir` of the hook, not in the directory from which the hook is invoked and if the user were to run the above command, it would in the worst case change the virtual environment of the project from which the hooks are invoked - and in no case fix the warning.

The solution to this is IMO, as mentioned in https://github.com/pre-commit/pre-commit/pull/1799#issuecomment-792697793, to require the hook authors to provide the activation script `renv/activate.R` in the hook repo in addition to `renv.lock`, as [recommended for collaboration per the docs of `{renv}`](https://rstudio.github.io/renv/articles/collaborating.html#collaborating-with-renv-1) (`.Rprofile` can be omitted since all it contains is a call to `renv/activate.R`). This will boostrap the right version of `{renv}` installation into the virtual environment used for the hook. This approach fixes a few other problematic aspects of the current implementation in addition to the above:

* `{renv}` would no longer be installed into the global package library. We could go from a 99% insolation to 100% isolation from the global library.
* `{renv}` is not required in the global package library, i.e. the hooks still work if it was for some reason uninstalled or not available after a major R version update (where the user starts off with an empty library).
* https://github.com/pre-commit/pre-commit/pull/1799#discussion_r576345988 would not be an issue anymore, i.e. on linux, non-root users would no longer be required to have installed any package to make the hooks works.

[...]

PR #1841, opened 2021-03-16
Avoid warnings with R hooks when renv version don't match

lorenzwalthert:
Attempts to solve #1835 as initially outlined in https://github.com/pre-commit/pre-commit/issues/1835#issue-826659015. To further avoid warnings for differing R versions, we suppress warning messages on environment loading. To not write infrastructure like `.Rprofile`, we use `renv::load()` over `renv::activate()`. For installing additional dependencies, we also added skipping of the startup process. I think after #1878, I might have fixed the biggest bugs I created -.-

### DIFF
--- a/pre_commit/languages/r.py
+++ b/pre_commit/languages/r.py
@@ -15,6 +15,7 @@ from pre_commit.util import clean_path_on_failure
 from pre_commit.util import cmd_output_b
 
 ENVIRONMENT_DIR = 'renv'
+RSCRIPT_OPTS = ('--no-save', '--no-restore', '--no-site-file', '--no-environ')
 get_default_version = helpers.basic_get_default_version
 healthy = helpers.basic_healthy
 
@@ -69,12 +70,11 @@ def _entry_validate(entry: Sequence[str]) -> None:
 
 
 def _cmd_from_hook(hook: Hook) -> Tuple[str, ...]:
-    opts = ('--no-save', '--no-restore', '--no-site-file', '--no-environ')
     entry = shlex.split(hook.entry)
     _entry_validate(entry)
 
     return (
-        *entry[:1], *opts,
+        *entry[:1], *RSCRIPT_OPTS,
         *_prefix_if_file_entry(entry, hook.prefix),
         *hook.args,
     )
@@ -89,22 +89,23 @@ def install_environment(
     with clean_path_on_failure(env_dir):
         os.makedirs(env_dir, exist_ok=True)
         shutil.copy(prefix.path('renv.lock'), env_dir)
+        shutil.copytree(prefix.path('renv'), os.path.join(env_dir, 'renv'))
         cmd_output_b(
             'Rscript', '--vanilla', '-e',
             f"""\
             prefix_dir <- {prefix.prefix_dir!r}
-            missing_pkgs <- setdiff(
-                "renv", unname(installed.packages()[, "Package"])
-            )
             options(
                 repos = c(CRAN = "https://cran.rstudio.com"),
                 renv.consent = TRUE
             )
-            install.packages(missing_pkgs)
-            renv::activate()
+            source("renv/activate.R")
             renv::restore()
             activate_statement <- paste0(
-              'renv::activate("', file.path(getwd()), '"); '
+              'suppressWarnings({{',
+              'old <- setwd("', getwd(), '"); ',
+              'source("renv/activate.R"); ',
+              'setwd(old); ',
+              'renv::load("', getwd(), '");}})'
             )
             writeLines(activate_statement, 'activate.R')
             is_package <- tryCatch({{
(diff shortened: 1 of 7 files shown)


lorenzwalthert:
@asottile can you have a look at this and tell me if it makes sense, also in the light of our discussion in #1835? I used this locally for a few weeks and it worked fine. Maybe we need to figure out licensing for the {renv} files.

asottile:
yeah seems fine!  hmm yeah licensing I didn't think about that -- perhaps we can make a `./licenses` directory and add `license_files = licenses/renv_LICENCE` or whatever (assuming it is permissively licensed, if it isn't this might be a dead end)

lorenzwalthert:
[{renv} is MIT licensed](https://github.com/rstudio/renv/blob/master/DESCRIPTION#L13), as pre-commit, so maybe we don’t need to include additional licenses?  We can also ask in https://github.com/rstudio/renv how they suggest to handle this, because essentially every project that uses {renv} has this file in version control, and I haven’t seen any special copyright attribution in any project tust uses {renv} (although our case might not be 100% the same).

asottile:
I would rather lean on the side of caution here -- it's not difficult to abide by the license terms -- we just need a copy of their license in our source tree and our distributables

we could even put it in the resources directory so it gets copied for the empty template into `renv` just to be on the safe side

sorry for the slow response, things have been crazy recently

lorenzwalthert:
> sorry for the slow response, things have been crazy recently
No problem.

> we just need a copy of their license in our source tree and our distributables

So licensing for R packages work usually like this: Put the type of licence in the `DESCRIPTION` file as a field. For {renv}, it's [`MIT + file LICENCE`](https://github.com/rstudio/renv/blob/master/DESCRIPTION#L13). Then, there is a separate file called `LICENCE`, which can have the actual licence or just the copyright holders. {renv} has only the [latter](https://github.com/rstudio/renv/blob/master/LICENSE), so then I guess we should include a MIT Licence and add the copyright and year from `LICENCE`.

asottile:
hmmmmm given the contents of that file, I'm not sure they've licensed it for redistribution -- we would probably want to clarify with them what terms that it is allowed to be copied and redistributed.

I don't see an actual license file in their repo beyond that 2-liner (which probably has no legal precedence).  I think it would be wrong to take that and stuff it into a MIT license (I think that would be relicensing without approval) -- we should probably start an issue thread on their side unless there's an actual copyable license file!

lorenzwalthert:
Sure, I can open an issue in their repo so we can figure things out and be on the Safe side. I know the authors and I think they don’t have fundamental objections to what we are doing here.

kevinushey:
> I don't see an actual license file in their repo beyond that 2-liner (which probably has no legal precedence). I think it would be wrong to take that and stuff it into a MIT license (I think that would be relicensing without approval) -- we should probably start an issue thread on their side unless there's an actual copyable license file!

FWIW this is just the canonical way that R packages (at least, those destined for [CRAN](https://cran.r-project.org/web/packages/)) declare that they are MIT-licensed (which I agree is a bit strange).

But perhaps I can change this so the, shall we say, CRAN-compatible-declaration-of-MIT-license version of the LICENSE file is used on CRAN, but the "real" version is used on GitHub.

asottile:
> 
> 
> > I don't see an actual license file in their repo beyond that 2-liner (which probably has no legal precedence). I think it would be wrong to take that and stuff it into a MIT license (I think that would be relicensing without approval) -- we should probably start an issue thread on their side unless there's an actual copyable license file!
> 
> FWIW this is just the canonical way that R packages (at least, those destined for [CRAN](https://cran.r-project.org/web/packages/)) declare that they are MIT-licensed (which I agree is a bit strange).
> 
> But perhaps I can change this so the, shall we say, CRAN-compatible-declaration-of-MIT-license version of the LICENSE file is used on CRAN, but the "real" version is used on GitHub.

yeah I think it would be sufficient for my uses to have a `LICENSE.CRAN` and a `LICENSE` or whatever naming you're comfortable with (this would also make it so github auto-detects your license too)

lorenzwalthert:
> But perhaps I can change this so the, shall we say, CRAN-compatible-declaration-of-MIT-license version of the LICENSE file is used on CRAN, but the "real" version is used on GitHub.

I think at some point, `LICENCE` used to contain the actual license text. If that is still CRAN compatible, it would maybe also be an option.

kevinushey:
> I think at some point, LICENSE used to contain the actual license text. If that is still CRAN compatible, would it maybe also be an option?

Unfortunately this is explicitly not accepted by CRAN. 😞

kevinushey:
I've made a change now so that:

- The canonical source code at https://github.com/rstudio/renv includes the standard MIT license file;
- When preparing a tarball for CRAN (via `R CMD build`), a tarball with the CRAN-compatible declaration of the MIT license is used instead.

Hopefully this is a fair compromise.

asottile:
@kevinushey thanks a ton!  that's super helpful!

lorenzwalthert:
Great. @asottile how do we progress from here? From the {renv} authors: https://github.com/rstudio/renv/issues/738

> These files are (at least currently) also MIT-licensed, and so I think they're fine to use in any project that is licensed in an MIT-compatible way. That is, please feel free to use them with pre-commit.

> The intent is definitely to allow users to freely use those files in their projects without needing to worry about copyright / licensing occurs, so we'll evaluate if unlicense might be a good choice.

asottile:
for now let's take their `LICENSE` file and also place that into renv directories as a resource

lorenzwalthert:
Please see c3d186d74d370932643a48ab27cad8aa714f5d1f how I added the licenses:

* pre_commit/resources/empty_template_LICENSE.renv (will be copied into the users `renv/` folder for local hooks, where the `activate.R` resides)
* testing/resources/r_hooks_repo/renv/LICENSE (where an `activate.R` file resides for testing)

Is that what you had in mind?

asottile (approved):

PR #1847, opened 2021-03-18
Remove nodeenv dependency

mxr:
Not ready for merge, still need to do the following..

- [ ] Get a better grasp of the file structure (when to use `prefix`, where I should be writing files, what to extract vs not, what permissions I need to add to which binaries)
- [ ] Clean up extra files
- [ ] Test this more especially on Windows
- [ ] More stuff based on CR


asottile:
feel free to reopen if you get a chance to get this closer to finish -- but for now probably going to continue with nodeenv at least

mxr:
Looking at this again, does it still make sense for this to be a drop-in replacement? I'm not sure if it's best to do (1) drop-in replacement (2) a separate `language: node2` (name tbd) (3) a new parameter like `language_installer` to be used with `language: node` or (4) something else

asottile:
I'm actually working on a version whether the same result happens so I wouldn't worry about it either way

mxr:
I'm not following what you mean by "whether the same result happens" - are you removing the `nodeenv` dep in a separate branch?

asottile:
yeah I'm reworking how languages work (slowly)

PR #1885, opened 2021-04-21
fix: fix Docker path mounting

### LINKED ISSUE #1387: Path is not mounted correctly when running Docker hooks from Docker
**Situation**:

- In our CI we want to run `pre-commit` inside Docker.
- Some of our hooks are `docker_image`

**Problem**
This line mostly https://github.com/pre-commit/pre-commit/blob/528c7afd18dafa6e47ce73add2c8e1550d105674/pre_commit/languages/docker.py#L94

Currently `pre-commit` mounts the current directory to `/src` and uses current directory name as mount base.
However this does not work when `pre-commit` is run inside the container on some mounted path already, because mount points are relative to the host, not to the container.

 Example: 
```
/opt/my_code      <- host, mounts /opt/my_code:/project
/project              <- in Docker running pre-commit, pre-commit is doing mount /project:/src
/src                   <-  (in Dockerized hook)
```

Currently pre-commit will try to mount it as `-v /project:/src,rw,Z`. Expected - to mount it as `-v /opt/my_code:/src`

**Possible solution**:

When I replaced `os.getcwd()` from the code above to `translate_path(os.getcwd())` where `translate_path` is taken from https://gist.github.com/dpfoose/f96d4e4b76c2e01265619d545b77987a, it worked perfectly. It does add extra `docker` pip-dependency though.

**See also**: https://forums.docker.com/t/mounting-a-volume-not-working-with-running-docker-in-docker/25775/2

okainov:
Currently pre-commit mounts the current directory to /src and uses current directory name as mount base.
However this does not work when pre-commit is run inside the container on some mounted path already, because mount points are relative to the host, not to the container.

See https://gist.github.com/dpfoose/f96d4e4b76c2e01265619d545b77987a
Fixes #1387

I just took the snipped we have used in our team dirty-patching already installed pre-commit and tried to fit it into existing code a bit more nicely. Haven't really tried resulted code though, but I believe it should be the same.


pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+def get_binds() -> Dict[str, str]:
+    """
+    :return: A dictionary with paths in the container as keys and paths
+             on the host as values
+    """
+    container = get_current_container()
+    return {
+        bind.split(':')[1]: bind.split(':')[0]
+        for bind in container['HostConfig']['Binds']
+    }
```
asottile, line 340: none of this should live in `util.py` -- it's specific to `docker`

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+        output = subprocess.check_output(('docker', 'inspect', hostname))
+    except CalledProcessError:
+        raise ValueError('Not running in Docker container')
+
+    return json.loads(output)[0]
+
+
+def get_binds() -> Dict[str, str]:
+    """
+    :return: A dictionary with paths in the container as keys and paths
```
asottile, line 333: most of these docstrings are not useful -- if it;s not obvious from the function name you should fix the function name

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+            return f'{binds[result.group(1)]}/{result.group(2)}'
+    raise ValueError(
+        f'Path {path} not present in a bind mount. ' +
+        'Volume mount will fail when running this in Docker.',
+    )
+
+
+def get_current_container() -> Dict[str, Any]:
+    """
+    Will raise ValueError if there is no container with the same hostname as
```
asottile, line 314: docstrings should never start with whitespace

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+    :param path: A string representing a path within the container
+    :return: A string representing a path on the host (or the original
+             path if the path is not in a bound volume)
+    """
+    if not in_docker():
+        return path
+    binds = get_binds()
+    if path in binds.keys():
+        return binds[path]
+    exps = ['(%s)/(.*)' % key for key in binds.keys()]
```
asottile, line 301: please don't use % formatting, it's twice-deprecatede (`.format()`, f-strings)

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+    for exp in exps:
+        result = re.search(exp, path)
+        if result:
+            return f'{binds[result.group(1)]}/{result.group(2)}'
```
asottile, line 305: `result[1]` and `result[2]`

pre_commit/util.py
asottile, line 301: [never use `.keys()` -- iterating over a dict works as-is](https://github.com/asottile/scratch/wiki/PythonHintsForNoobs#iterating-a-dict-gives-you-its-keys)

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+             path if the path is not in a bound volume)
+    """
+    if not in_docker():
+        return path
+    binds = get_binds()
+    if path in binds.keys():
+        return binds[path]
+    exps = ['(%s)/(.*)' % key for key in binds.keys()]
+    for exp in exps:
+        result = re.search(exp, path)
```
asottile, line 303: this is really really inefficient -- you should use `os.path` functions instead of regex searching binds

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
 def parse_version(s: str) -> Tuple[int, ...]:
     """poor man's version comparison"""
     return tuple(int(p) for p in s.split('.'))
+
+
+def in_docker() -> bool:
+    """
+    Check if running in Docker
+    :return: Whether or not this is running in Docker container
+    """
```
asottile, line 280: this docstring is incredibly useless!  what else would an `in_docker` function do?

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+    try:
+        with open('/proc/1/cgroup') as cgroup_file:
+            return 'docker' in cgroup_file.read()
+    except FileNotFoundError:
+        return False
+
+
+def translate_path(path: str) -> str:
+    """
+    Method to get the right path considering it can be mounted in Docker
```
asottile, line 290: this isn't a method, methods are on classes.  this is a function

pre_commit/util.py
```diff
@@ -268,3 +271,70 @@ def handle_remove_readonly(
...
+        return binds[path]
+    exps = ['(%s)/(.*)' % key for key in binds.keys()]
+    for exp in exps:
+        result = re.search(exp, path)
+        if result:
+            return f'{binds[result.group(1)]}/{result.group(2)}'
+    raise ValueError(
+        f'Path {path} not present in a bind mount. ' +
+        'Volume mount will fail when running this in Docker.',
+    )
```
asottile, line 309: I don't believe this should error, the status quo does not error

PR #1888, opened 2021-04-21
fix: fix path mounting when running in Docker

### LINKED ISSUE #1387: Path is not mounted correctly when running Docker hooks from Docker

okainov:
Currently pre-commit mounts the current directory to /src and uses
current directory name as mount base.
However this does not work when pre-commit is run inside the container
on some mounted path already, because mount points are relative to the
host, not to the container.

Fixes #1387

Second try. Sorry, no way I can not copy-paste `in_docker` one, it looks too straightforward and quite pythonic already IMO. But the rest was indeed not really optimal, rewrote completely.

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -1,5 +1,7 @@
 import hashlib
+import json
 import os
+import socket
 from typing import Sequence
 from typing import Tuple
 
@@ -8,11 +10,40 @@ from pre_commit.hook import Hook
 from pre_commit.languages import helpers
 from pre_commit.prefix import Prefix
 from pre_commit.util import clean_path_on_failure
+from pre_commit.util import cmd_output_b
 
 ENVIRONMENT_DIR = 'docker'
 PRE_COMMIT_LABEL = 'PRE_COMMIT'
 get_default_version = helpers.basic_get_default_version
 healthy = helpers.basic_healthy
+
+
+def _is_in_docker() -> bool:
+    try:
+        with open('/proc/1/cgroup', 'rb') as f:
+            return b'docker' in f.read()
+    except FileNotFoundError:
+        return False
+
+
+def _get_docker_path(path: str) -> str:
+    if not _is_in_docker():
+        return path
+    hostname = socket.gethostname()
+
+    _, out, _ = cmd_output_b('docker', 'inspect', hostname)
+
+    container, = json.loads(out)
+    for mount in container['Mounts']:
+        src_path = mount['Source']
+        to_path = mount['Destination']
+        if os.path.commonpath((path, to_path)) == to_path:
+            # So there is something in common,
+            # and we can proceed remapping it
+            return path.replace(to_path, src_path)
+    # we're in Docker, but the path is not mounted, cannot really do anything,
+    # so fall back to original path
+    return path
 
 
 def md5(s: str) -> str:  # pragma: win32 no cover
@@ -73,7 +104,7 @@ def docker_cmd() -> Tuple[str, ...]:  # pragma: win32 no cover
         # https://docs.docker.com/engine/reference/commandline/run/#mount-volumes-from-container-volumes-from
         # The `Z` option tells Docker to label the content with a private
         # unshared label. Only the current container can use a private volume.
-        '-v', f'{os.getcwd()}:/src:rw,Z',
+        '-v', f'{_get_docker_path(os.getcwd())}:/src:rw,Z',
         '--workdir', '/src',
     )
 
--- a/tests/languages/docker_test.py
+++ b/tests/languages/docker_test.py
@@ -1,4 +1,11 @@
+import builtins
+import json
+import ntpath
+import os.path
+import posixpath
 from unittest import mock
+
+import pytest
 
 from pre_commit.languages import docker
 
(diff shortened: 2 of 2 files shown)


okainov:
```
Name                             Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------
pre_commit/languages/docker.py      65      0     10      1    99%   40->36
----------------------------------------------------------------------------
TOTAL                             7089      0    714      1    99%

100 files skipped due to complete coverage.
Coverage failure: total of 100 is less than fail-under=100

```
I don't really understand why CI is not happy. All lines are test-covered. what is `40->36`?

asottile:
> 
> 
> ```
> Name                             Stmts   Miss Branch BrPart  Cover   Missing
> ----------------------------------------------------------------------------
> pre_commit/languages/docker.py      65      0     10      1    99%   40->36
> ----------------------------------------------------------------------------
> TOTAL                             7089      0    714      1    99%
> 
> 100 files skipped due to complete coverage.
> Coverage failure: total of 100 is less than fail-under=100
> ```
> 
> I don't really understand why CI is not happy. All lines are test-covered. what is `40->36`?

that's branch coverage -- it means that once it enters the loop you've only tested the two positive cases inside and not the (implicit) `else: pass`

okainov:
@asottile fixed, please review and merge if it's fine

pre_commit/languages/docker.py
```diff
@@ -15,6 +18,32 @@
 healthy = helpers.basic_healthy
 
 
+def is_in_docker() -> bool:
+    try:
+        with open('/proc/1/cgroup') as f:
+            return 'docker' in f.read()
```
asottile, line 24: it would be slightly better to read this as binary `, 'rb'` and then do `b'docker'` in ...

pre_commit/languages/docker.py
```diff
@@ -15,6 +18,32 @@
...
+    except FileNotFoundError:
+        return False
+
+
+def get_docker_path(path: str) -> str:
```
asottile, line 29: probably underscore prefix this and the above function since they are private to this module

pre_commit/languages/docker.py
```diff
@@ -15,6 +18,32 @@
...
+    if not is_in_docker():
+        return path
+    hostname = socket.gethostname()
+    docker_output = json.loads(
+        subprocess.check_output(['docker', 'inspect', hostname]),
```
asottile, line 34: this should use `cmd_output_b` which normalizes a bunch of stuff

pre_commit/languages/docker.py
```diff
@@ -15,6 +18,32 @@
...
+    )
+    for mount_path in docker_output[0]['HostConfig']['Binds']:
+        src_path, to_path = mount_path.split(':')
```
asottile, line 37: is it possible that these have more than 2 colons?  I seem to remember binds can have a third segment (such as for readonly / etc. options)

pre_commit/languages/docker.py
```diff
@@ -15,6 +18,32 @@
...
+        if to_path == path:
+            return src_path
+        elif path.startswith(to_path):
+            return path.replace(to_path, src_path)
```
asottile, line 41: this is slightly error prone, you could have `/foo/bar` and `/foo/barbutsomethingelse` and it would do the wrong thing --  I think you have to use the special path functions to figure this out (I believe there's a prefix function somewhere in `os.path`)

tests/languages/docker_test.py
```diff
@@ -12,3 +16,75 @@ def invalid_attribute():
         getgid=invalid_attribute,
     ):
         assert docker.get_docker_user() == ()
+
+
+class TestInDocker:
```
asottile, line 21: we don't use test classes in this project

tests/languages/docker_test.py
```diff
@@ -12,3 +16,75 @@ def invalid_attribute():
...
+    @mock.patch('builtins.open', new_callable=mock_open)
+    def test_in_docker_no_file(self, mock_file):
+        mock_file.side_effect = FileNotFoundError
+
+        assert docker.is_in_docker() is False
+        mock_file.assert_called()
+
+    @mock.patch('builtins.open', new_callable=mock_open, read_data='tdockert')
```
asottile, line 29: please prefer the context manager, the decorator form is signature-mutating and much harder to understand the scope and debug.  you'll notice that there is no use of the mock decorator in this project

asottile: also please use `mock.patch.object` -- `mock.patch` has an undesirable side-effect of triggering an import which can easily lead to test pollution and other strange bugs

tests/languages/docker_test.py
```diff
@@ -12,3 +16,75 @@ def invalid_attribute():
...
+            '/opt/my_code:/project',
+        ]
+        output_string = json.dumps([{'HostConfig': {'Binds': binds_list}}])
+        with mock.patch.object(
+            subprocess, 'check_output',
+            return_value=output_string,
+        ):
+            assert docker.get_docker_path('/project') == '/opt/my_code'
+
+    @mock.patch.object(docker, 'is_in_docker', return_value=True)
```
asottile, line 65: this should be a fixture that gets reused in each of the tests

tests/languages/docker_test.py
```diff
@@ -12,3 +16,75 @@ def invalid_attribute():
...
+    def test_in_docker_binds_path_complex(self, _):
+        binds_list = [
+            '/opt/my_code:/project',
+        ]
+        output_string = json.dumps([{'HostConfig': {'Binds': binds_list}}])
+        with mock.patch.object(
+            subprocess, 'check_output',
+            return_value=output_string,
+        ):
+            assert docker.get_docker_path('/project/test/something') == \
```
asottile, line 75: don't use backslashes

tests/languages/docker_test.py
```diff
@@ -12,3 +16,75 @@ def invalid_attribute():
...
+
+
+class TestDockerPath:
+    @mock.patch.object(docker, 'is_in_docker', return_value=False)
+    def test_not_in_docker_returns_same(self, _):
+        assert docker.get_docker_path('abc') == 'abc'
+
+    @mock.patch.object(docker, 'is_in_docker', return_value=True)
+    def test_in_docker_no_binds_same_path(self, _):
+        binds_list: List[str] = []
```
asottile, line 45: if you inline this you shouldn't need the type signature (?)

pre_commit/languages/docker.py
okainov: done

pre_commit/languages/docker.py
okainov: done

pre_commit/languages/docker.py
okainov: done

pre_commit/languages/docker.py
okainov: done

tests/languages/docker_test.py
okainov: Well, I do want to group tests for a given function together. Test classes are the best suited for this. Having bunch of files like `test_modulename_functionname` is a bit too much IMO

tests/languages/docker_test.py
okainov: I got warning from CI about it

pre_commit/languages/docker.py
okainov: done

pre_commit/languages/docker.py
okainov: done

tests/languages/docker_test.py
okainov: done

tests/languages/docker_test.py
okainov: done

tests/languages/docker_test.py
okainov: not sure alternative is in any way better/easier to read, but whatever

pre_commit/languages/docker.py
okainov: done

pre_commit/languages/docker.py
```diff
@@ -8,13 +10,43 @@
...
+
+
+def _get_docker_path(path: str) -> str:
+    if not _is_in_docker():
+        return path
+    hostname = socket.gethostname()
+
+    _, out, _ = cmd_output_b(
+        ' '.join(['docker', 'inspect', hostname]),
+    )
```
asottile, line 36: this should be `cmd_output_b('docker', 'inspect', 'hostname')`

pre_commit/languages/docker.py
```diff
@@ -8,13 +10,43 @@
...
+
+    docker_output = json.loads(out)
+    for mount_path in docker_output[0]['HostConfig']['Binds']:
+        src_path, to_path, *_ = mount_path.split(':')
+        if os.path.commonpath([path, to_path]) != os.path.sep:
```
asottile, line 41: (1) [prefer tuples to lists](https://github.com/asottile/scratch/wiki/Prefer-tuples-to-lists) (2) this isn't quite correct -- you'll need to use `== to_path` I believe:

```pycon
>>> os.path.commonpath(('/a/b/c', '/a/d/b'))
'/a'
```

tests/languages/docker_test.py
```diff
@@ -1,14 +1,134 @@
+import builtins
+import json
+from typing import List
 from unittest import mock
+from unittest.mock import mock_open
```
asottile, line 5: the `mock` module is already imported -- you can use `mock.mock_open` directly

tests/languages/docker_test.py
asottile: we don't use test classes in this project

tests/languages/docker_test.py
asottile: I mean if you don't have this as a variable you shouldn't need the annotation

tests/languages/docker_test.py
```diff
@@ -1,14 +1,134 @@
...
+            getgid=invalid_attribute,
     ):
         assert docker.get_docker_user() == ()
+
+
+class TestInDocker:
+
+    @pytest.fixture
+    def mock_file_fixture(self):
+        return lambda read_data: \
```
asottile, line 28: no backslashes, again

you probably want this to  be a yield fixture:

```
@pytest.fixture
def mocked_read():
    with mock.patch.object(builtins, 'open', new_callable=mock_open) as mck:
        yield mck
```

and then in your test, utilize `mocked_read.read_data = ...`

tests/languages/docker_test.py
```diff
@@ -1,14 +1,134 @@
...
+            assert docker._get_docker_path('abc') == 'abc'
+
+    def test_in_docker_binds_path_equal(self, in_docker):
+        binds_list = [
+            '/opt/my_code:/project',
+        ]
+        output_string = json.dumps([{'HostConfig': {'Binds': binds_list}}])
+        with mock.patch.object(
+                docker, 'cmd_output_b',
+                return_value=(0, output_string, b''),
```
asottile, line 81: this `output_string` is of the wrong type -- it should be bytes

pre_commit/languages/docker.py
okainov: done. Function signature is a bit misleading then =(

tests/languages/docker_test.py
okainov: I want to have consistent tests so they look alike

tests/languages/docker_test.py
okainov: done

tests/languages/docker_test.py
okainov: Please be more constructive. I would like to have tests grouped together by their test target. Test classes is the built-in way to do so and in comparison with bare functions it improves both readability of the code an readability of reports. They don't introduce any extra tremendous boilerplate, so what's the problem?

tests/languages/docker_test.py
okainov: >you probably want this to be a yield fixture:

Not really, `yield` one doesn't really work, it doesn't allow to set `read_data` afterwards. Seems like it can be set only in constructor, so I keep it like it is.

tests/languages/docker_test.py
okainov: doned

pre_commit/languages/docker.py
okainov: this leads to different behavior on Windows and Linux, so I've updated the source for mounts, added Win-specific test, done.

pre_commit/languages/docker.py
asottile: what is misleading about it?  it collects strings into `cmd` therefore `*cmd`

okainov: Because "cmd" is usually "a cmd", single command line. If it accepts many things, "args" or something similar would have been clearer IMO. But nevermind.

okainov:
@asottile thanks for merging. Any plans when to expect new release?

asottile:
no, it's also kinda rude to ask -- the release will happen when it happens

PR #1911, opened 2021-05-15
Use more inclusive language

asottile:
Committed via https://github.com/asottile/all-repos


MarcoGorelli:
Nice! In a similar vein, perhaps consider adding a code of conduct?

PR #1614, opened 2020-09-23
update rbenv / ruby-build

asottile:
resolves #1612

PR #1913, opened 2021-05-15
Update rbenv / ruby-build

jalessio:
Similar to PR https://github.com/pre-commit/pre-commit/pull/1614, updates `rbenv` and `ruby-build` to their latest releases which  makes more recent versions of Ruby available to pre-commit.

### rbenv

Doesn't look like `rbenv` has [any recent tags](https://github.com/rbenv/rbenv/tags) _(latest is from March 25, 2019)_, so this is current HEAD of master branch:
https://github.com/rbenv/rbenv/commit/585ed84283f3308380b843391ee7b12706ecff87

### ruby-build 20210510

https://github.com/rbenv/ruby-build/releases/tag/v20210510


jalessio:
@asottile per your mention about "reproducible builds" for the `tar.gz` files it could be accomplished in a Github-specific way by updating `make-archives` grab the URLs like this instead of doing a `git clone`:

https://github.com/rbenv/rbenv/archive/585ed84283f3308380b843391ee7b12706ecff87.tar.gz

Looks like it also supports short shas:
https://github.com/rbenv/ruby-build/archive/e9fa4bf.tar.gz

It also looks like calling that repeatedly produces a tar file with the same `sha1sum` so in a way that is "reproducible".

This would work for the given set of [REPOS](https://github.com/pre-commit/pre-commit/blob/147b047487f80872644dbf2db6b66c06f57db784/testing/make-archives#L16-L24) since they're all hosted on Github, but wouldn't work if other git hosting providers are added to that list.

asottile:
I'll rebase this on #1915 once that merges!

Issue #1897, opened 2021-05-02
How to change font color of `Passed` for hooks that have passed on pre-commit in iTerm2?


avatar-lavventura:
On pre-commit `...Passed` and `...Skipped` colors are as follows (Passed color is white where it is not readable on top of green background) if users are using dark theme:

[![enter image description here][1]][1]

Here https://pre-commit.com/#usage, it's much more readable:

[![enter image description here][2]][2]

How could I change the font and background color of `Passed` and `Skipped` that is located at the end of each check?

Related issue https://github.com/pre-commit/pre-commit/issues/969 has only solutions as disabling colors with `--color=never`

--------

- I am using [`iTerm2`][3] as terminal emulator.
- My shell is `zsh`.
-  Theme is [`dracula`](https://draculatheme.com/iterm).

My color configuration under iTerm2:

[![enter image description here][4]][4]

-----

Afterwards, when I change foreground theme to something else than white, font color of `...Passed` is changed but it effects my overall font configuration.

[![enter image description here][5]][5]

  [1]: https://i.stack.imgur.com/uQPvL.png
  [2]: https://i.stack.imgur.com/VEfns.png
  [3]: https://iterm2.com
  [4]: https://i.stack.imgur.com/4BryRl.png
  [5]: https://i.stack.imgur.com/EzUjz.png

asottile:
the foreground color is going to be the same as your default text color -- I'd recommend you make the green color darker ?

avatar-lavventura:
I am sorry for keep asking questions related to this.

I changed bright green color to dark-red as solution.

kdheepak:
If anyone else is interested in configuring the colors, you can hack it by changing the global variables on the cli entry. Just as an example, you can have `Passed` in italics, `Skipped` in strikethrough and `Failed` in bold and red, illustrated in the screenshot below:

<img width="1680" alt="Screen Shot 2021-07-02 at 2 11 21 PM" src="https://user-images.githubusercontent.com/1813121/124324102-a6099300-db3f-11eb-918a-e93f03d497ea.png">

I'm pretty sure **_the developers will not recommend doing this_**, so if you choose to proceed you are on your own. I'm  sharing in case someone else is looking for something like this:

1) Find where `pre-commit` is installed: `which pre-commit` on UNIX or `where pre-commit` on Windows:

You should see a file that looks something like this:

```python
!#/usr/bin/env python
# -*- coding: utf-8 -*-
import re
import sys
[...]

PR #1962, opened 2021-07-01
Make color configurable

kdheepak:
This is related to issues like https://github.com/pre-commit/pre-commit/issues/1897. Currently one needs to change the terminal's color to make text readable. This is not ideal and it would be nice if the colors were configurable using environment variables (or alternatively command line arguments). This PR allows for using environment variables to change the colors.

### DIFF
--- a/pre_commit/color.py
+++ b/pre_commit/color.py
@@ -54,10 +54,10 @@ if sys.platform == 'win32':  # pragma: no cover (windows)
 else:  # pragma: win32 no cover
     terminal_supports_color = True
 
-RED = '\033[41m'
-GREEN = '\033[42m'
-YELLOW = '\033[43;30m'
-TURQUOISE = '\033[46;30m'
+RED = os.environ.get("PRE_COMMIT_RED", '\033[30;41m')
+GREEN = os.environ.get("PRE_COMMIT_GREEN", '\033[30;42m')
+YELLOW = os.environ.get("PRE_COMMIT_YELLOW", '\033[43;30m')
+TURQUOISE = os.environ.get("PRE_COMMIT_TURQUOISE", '\033[46;30m')
 SUBTLE = '\033[2m'
 NORMAL = '\033[m'
 


asottile:
no thank you, sorry

kdheepak:
Can you elaborate on why this is not something you'd like to do? In most popular dark themes, the text for "passed"  is just not readable.

Here's what it looks like before and after configuring it:

Screen Shot 2021-07-01 at 3 42 14 PM

Screen Shot 2021-07-01 at 3 43 47 PM

kdheepak:
Or would you be at least open to considering a PR with the foreground text color changed to black? That makes it easier to read on light and dark themes.

Screen Shot 2021-07-01 at 3 46 30 PM

asottile:
I've already elaborated in the closed issue you linked and the duplicates from there

kdheepak:
Thanks for the reply. I've read your comments on the other issues. With all due respect, it is my opinion that you should reconsider your position on this. 

Firstly, I think the defaults are not ideal. Using black on green or red makes it much more readable compared to white on green or red (imo). As you can see from screenshots, there are very popular themes that I'm using and here precommit's defaults are basically not readable. 

And I think maybe we are in agreement about the above point, since you suggest changing the theme's base colors to make precommit look readable. 

I don't think changing the theme's base color is a practical answer though. For example, I have other tools configured to match the theme colors in the terminal. Changing what "green" or "red" or foreground color in the base theme might make text in other tools not readable. 

Also, I have other software using 256 colors, and currently it is not possible for me to make things to look consistent if this is not configurable. It would also be nice to use 256 colors in precommit if a user wanted to do so. I didn't see any points in your comments in other issues about why you are against a change for making the colors more configurable. Could you elaborate your thoughts on this?

Lastly, the defaults here for green and red are good for implying "green" and "red" for "pass" and "fail" respectively, but as someone with red-green color blindness this is visually really hard to distinguish. Red green colorblindness is quite common too and I can't imagine others are not feeling similarly given this project's large appeal :)

asottile:
I don't need a lecture, I'm not changing the branding, you're free to configure your terminal however you want.  good day.

PR #1965, opened 2021-07-02
Add pre-commit hook for tox-ini-fmt

wwuck:
Format tox.ini automatically with a pre-commit hook


asottile:
tip for open source in general: don't touch anyone's infrastructure unless you are asked to

Issue #1972, opened 2021-07-11
wrong path in case of workdirs 


daschuer:
When working with work dirs the conflict check is using the wrong path. This results in a weird error message during a commit with the base worktree in conflict. 
```
[INFO] Checking merge-conflict files only.
An unexpected error has occurred: CalledProcessError: command: ('/usr/lib/git-core/git', 'diff', '--name-only', '--no-ext-diff', '-z', '-m', 'e8fb6f96dfe86b454ceb6913a3e02f88a9103565', 'HEAD', 'MERGE_HEAD')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: mehrdeutiges Argument 'MERGE_HEAD': unbekannter Commit oder Pfad existiert nicht
    im Arbeitsverzeichnis
    Benutzen Sie '--', um Pfade und Commits zu trennen, ähnlich wie:
    'git <Befehl> [<Commit>...] -- [<Datei>...]'
    
Check the log at /home/sperry/.cache/pre-commit/pre-commit.log
```

daschuer:
It is a possible regression from https://github.com/pre-commit/pre-commit/pull/832 fixing https://github.com/pre-commit/pre-commit/issues/831

It looks like we need tow get_git_dir() functions one for the git dir for the regular usage  and  one for the common git dir used during install.

asottile:
you're missing: a reproduction of your error and a test in the pr demonstrating the change

daschuer:
The issue can reproduced like that: 
* create a second worktee from the default worktree that contains the common .git folder 
* create a conflicting merge in the origin worktee
* don't resolve the conflicts 
* change to the other worktree folder 
* change a file 
* commit 
* the error message posted above appears.

I have no clue how to create a test for that. Can you help out here? Is there an other test that can be taken as a template?

asottile:
there's a handful of relevant similar tests in https://github.com/pre-commit/pre-commit/blob/835f9c65e93cdaa517fb87a4cd42edb4f9b084a3/tests/git_test.py#L27 and https://github.com/pre-commit/pre-commit/blob/835f9c65e93cdaa517fb87a4cd42edb4f9b084a3/tests/staged_files_only_test.py

I believe the latter on is relevant to the error you posted above

daschuer:
The test is in place.

daschuer:
Since https://github.com/pre-commit/pre-commit/pull/1973 has been closed, is this issue solved?

[...]

PR #809, opened 2018-08-06
Support `pre-commit install` inside a worktree

asottile:

PR #1973, opened 2021-07-11
Split get_git_dir() into get_git_dir() and get_common_git_dir()

daschuer:
This fixes the conflicted state check when using work trees. #1972

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -40,7 +40,7 @@ def _hook_paths(
         hook_type: str,
         git_dir: Optional[str] = None,
 ) -> Tuple[str, str]:
-    git_dir = git_dir if git_dir is not None else git.get_git_dir()
+    git_dir = git_dir if git_dir is not None else git.get_git_common_dir()
     pth = os.path.join(git_dir, 'hooks', hook_type)
     return pth, f'{pth}.legacy'
 
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -55,13 +55,15 @@ def get_root() -> str:
         root = os.path.abspath(
             cmd_output('git', 'rev-parse', '--show-cdup')[1].strip(),
         )
-        git_dir = os.path.abspath(get_git_common_dir())
+        inside_git_dir = cmd_output(
+            'git', 'rev-parse', '--is-inside-git-dir',
+        )[1].strip()
     except CalledProcessError:
         raise FatalError(
             'git failed. Is it installed, and are you in a Git repository '
             'directory?',
         )
-    if os.path.samefile(root, git_dir):
+    if inside_git_dir != 'false':
         raise FatalError(
             'git toplevel unexpectedly empty! make sure you are not '
             'inside the `.git` directory of your repository.',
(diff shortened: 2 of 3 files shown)


asottile:
would it maybe make more sense to change the `is_in_merge_conflict` to utilize `git rev-parse -q --verify MERGE_HEAD` ? and specifically test that in the test?

daschuer:
The root cause of this issue was that get_git_dir() returned the wrong directory in the worktree case.  
The test checks that this is fixed now. 
The issue with the merge conflict was only a symptom because is_in_merge_conflict() calls  get_git_dir('.') 
So I don't think it is necessary to add another test.

asottile:
I'm not asking for another test -- I think that `get_git_dir` should maybe not be used to detect merge conflicts and that a more reliable approach would not require the rest of the refactor

daschuer:
I have double checked the usage and discovered a minor issue with get_root() not failing in a .git sub dir as desired.  
A test is not possible, because the case causes a fatal error or do you know a way to assert a fatal error call?  

In general I think the refactoring is reasonable, because get_git_dir() now uses '--git-dir' and get_git_common_dir() uses  '---git-common-dir'. The alternative would be much more confusing if we stick with get_git_dir for '---git-common-dir' and new name like get_git_individual_dir or such for '---git-dir'

I have rebased this PR to rename get_common_git_dir() to get_git_common_dir() to match the git parameter. 

What do you think now?

asottile:
`--git-common-dir` is too new for many git versions so changing to use that unconditionally isn't acceptable

can you see if *just* fixing the `in_merge_conflict` function solves your issue?  take my suggestion above please

to assert a failure you can use `pytest.raises` -- there are quite a few examples already

daschuer:
The fall back path still exists for  git < 2.5.

daschuer:
`get_conflicted_files` is also broken.  All in all we have four usages of git_get_dir() two are broken one is broken for another reason and one is correct.

daschuer:
This is the breaking commit https://github.com/pre-commit/pre-commit/commit/3f784877695da4b86e76086136dfc58149779f32
https://github.com/pre-commit/pre-commit/pull/809
Before get_git_dir() was using --git-dir as expected and `is_in_merge_conflict` was working.
This PR restores the original function an provides another function to maintain the capability of worktree installation.

What do you think? Merge?

asottile:
> What do you think? Merge?

I think the tests are failing -- I also really think there's probably a way to rewrite the merge-conflict parts to not use `git_dir` at all and then the refactor wouldn't be needed

daschuer:
Which tests are failing? I am a bit lost with the Azure results. 

If we want to keep the old get_git_dir function to return the git common dir, we need a new function which returns the real git dir.
How should this be called?

I did not get the point why this solution is problematic. What issues do you expect? 

This PR fixes a regression since #809 by reverting the changed behaviour of get_git_dir() and introducing a new one with the most obvious name.

asottile:
if you click through azure pipelines and click on the red things you eventually get to tox output -- they are all failing these tests:

```
FAILED tests/git_test.py::test_get_root_bare_worktree - pre_commit.errors.Fat...
FAILED tests/git_test.py::test_get_root_worktree_in_git - pre_commit.errors.F...
```

daschuer:
Both text are fixed now. I had basically reverted https://github.com/pre-commit/pre-commit/pull/1778, Now I use --is-inside-git-dir to make the old and the new test working.

daschuer:
Azure is still red. Is there something missing?

PR #2001, opened 2021-08-05
Permit installing to core.hooksPath if within repo

noahp:
### Summary

There's [several
threads](https://github.com/pre-commit/pre-commit/issues?q=is%3Aissue+core+hooksPath+is%3Aclosed)
around using pre-commit with global git hooks directories, but I didn't
see anything around setting `core.hooksPath` to a directory within the
repo, as the `husky` tool does.

This patch adds support for installing/uninstalling pre-commit when
`core.hooksPath` references a path within the git repo.

 ### Testing

- add a couple of simple tests
- a repo with core.hooksPath set within the repo, install/uninstall does
  the right thing
- a repo with core.hooksPath set to `../hooks`, install/uninstall fails
- a repo with core.hooksPath unset, install/uninstall works as before

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -3,6 +3,7 @@ import logging
 import os.path
 import shutil
 import sys
+from pathlib import Path
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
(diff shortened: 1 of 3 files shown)


asottile:
the conclusion of both of those threads is "no thank you"

PR #2025, opened 2021-08-27
Support custom core.hooksPath

uryyakir:
I get that `pre-commit` should treat repos with a set `core.hooksPath` with caution, but it could simply follow that custom directory and install the hooks there, right?
IMO people with a custom hooks directory should have access to `pre-commit`, instead of having to reset their path...

I implemented a `--follow-custom-hooks-path` flag for `pre-commit install` that follows the custom hooks path, whilst maintaining the original behavior of `pre-commit install` (force installing, renaming previously defined hook to `pre-commit.legacy` etc).
I also slightly adjusted the error message when running `pre-commit install` with a set `core.hooksPath`.

(This is my first open-source contribution, hopefully it's not too bad :)

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -146,7 +146,8 @@ def install(
         if not follow_hooks_path:
             logger.error(
                 'Cowardly refusing to install hooks with `core.hooksPath` set.'
-                '\nhint: `git config --unset-all core.hooksPath`',
+                '\nhint: `git config --unset-all core.hooksPath`, '
+                'or use the `--follow-custom-hooks-path` flag',
             )
             return 1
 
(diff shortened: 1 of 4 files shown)


asottile:
sorry, I've already decided I do not want to support this -- please see previous discussions: https://github.com/pre-commit/pre-commit/issues?q=is%3Aissue+core.hooksPath+is%3Aclosed

uryyakir:
darn, should've looked at that before I jumped to coding...
What about enforcing that the set path is indeed inside the repository?
The thing is - a lot of people (including myself) create a custom hooks directory INSIDE the repo (but outside the `.git/` folder) so hooks actually get committed and pushed to GitHub.
@asottile

asottile:
no, it is not safe to encourage that

Issue #2043, opened 2021-09-08
add warning for regular expression with `[/\]`


asottile:
pre-commit normalizes slashes to forward slashes since 2.5.0 (and further back to 1.19.0 for per-hook configuration)

there's already a [similar warning for `/*`](https://github.com/pre-commit/pre-commit/blob/2ef29b7f95384f899ede86a370bf2a1c766095b4/pre_commit/clientlib.py#L140-L145) so this should be a fairly beginner-friendly changeset

dogukanteber:
Can you open up the issue a little more, please? I do not understand the problem exactly.

asottile:
right now, one may write something like:

```yaml
    files: ^somedir[/\].*\.py$
    # or
    exclude: ^somedir[\/]otherdir[\/].*\.py$
```

expecting that they need to match both posix paths `somedir/foo/bar.py` and windows paths `somedir\foo\bar.py` -- however this is unnecessary because pre-commit normalizes all slashes to forward slashes

there's a similar regular expression warning for doing something like `files: whatever/*.py` where someone assumes that it is a glob and not a regex

asottile:
via #2053

PR #2053, opened 2021-09-17
Add warning for regular expression with [\/] (#2043)

radek-sprta:
Add warning about unnecessary slash regex as per #2043

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -143,6 +143,18 @@ class OptionalSensibleRegexAtHook(cfgv.OptionalNoDefault):
                 f"regex, not a glob -- matching '/*' probably isn't what you "
                 f'want here',
             )
+        if r'[\/]' in dct.get(self.key, ''):
+            logger.warning(
+                fr'pre-commit normalizes slashes in the {self.key!r} field '
+                fr'in hook {dct.get("id")!r} to forward slashes, so you '
+                fr'can use / instead of [\/]',
+            )
+        if r'[/\\]' in dct.get(self.key, ''):
+            logger.warning(
+                fr'pre-commit normalizes slashes in the {self.key!r} field '
+                fr'in hook {dct.get("id")!r} to forward slashes, so you '
+                fr'can use / instead of [/\\]',
+            )
 
 
 class OptionalSensibleRegexAtTop(cfgv.OptionalNoDefault):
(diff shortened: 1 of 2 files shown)


pre_commit/clientlib.py
```diff
@@ -156,6 +156,30 @@ def check(self, dct: Dict[str, Any]) -> None:
...
+class OptionalSensibleSlashesAtHook(cfgv.OptionalNoDefault):
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if r'[\/]' in dct.get(self.key, ''):
+            logger.warning(
+                f'Pre-commit normalizes the slashes in {self.key!r} field in '
+                f"hook {dct.get('id')!r} to forward slashes, so you can use "
+                f'/ instead of [\\/]',
+            )
```
asottile, line 168: I would just add these to the existing SensibleRegex checks

also this should check for both `r'[\/]'` and `r'[/\]'` since those are both valid spellings of "either slash"

pre_commit/clientlib.py
```diff
@@ -156,6 +156,30 @@ def check(self, dct: Dict[str, Any]) -> None:
             )
 
 
+class OptionalSensibleSlashesAtHook(cfgv.OptionalNoDefault):
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if r'[\/]' in dct.get(self.key, ''):
+            logger.warning(
+                f'Pre-commit normalizes the slashes in {self.key!r} field in '
```
asottile, line 165: pre-commit is not capitalized

pre_commit/clientlib.py
```diff
@@ -156,6 +156,30 @@ def check(self, dct: Dict[str, Any]) -> None:
...
+
+class OptionalSensibleSlashesAtTop(cfgv.OptionalNoDefault):
+    def check(self, dct: Dict[str, Any]) -> None:
+        super().check(dct)
+
+        if r'[\/]' in dct.get(self.key, ''):
+            logger.warning(
+                f'Pre-commit normalizes the slashes in the top-level '
+                f'{self.key!r} field to forward slashes, so you can use / '
+                f'instead of [\\/]',
```
asottile, line 179: you can use `fr'...'` for a raw f-string by the way

radek-sprta: Fixed.

pre_commit/clientlib.py
radek-sprta: I capitalized it as a sentence, even though it wasn't one. Fixed.

pre_commit/clientlib.py
radek-sprta: I've merged the check into the existing functions. However, `r[/\]` is not a valid regex, `\]` is an escape sequence to enable things like `[...\]]`.

asottile: `[/\\]` is though, with the same issue

radek-sprta: Ok, I'll add it. Although it seems unlikely to me, that someone would use '[/\\]' instead of just '[\/]'.

radek-sprta: I've kept it in three ifs. If you feel looping through the regexes in for loop would be more readable, let me know.

asottile (approved):

sanjioh:
@asottile @radek-sprta Hi! Thanks for this feature. I just wanted to point out that there's one more permutation that's currently not addressed and you migth want to consider, `[\\/]`. I have it in one of my `.pre-commit-config.yaml` files and it didn't get caught by `pre-commit` 2.16. Cheers!

asottile:
@sanjioh can you make a new issue please?

Issue #2072, opened 2021-10-02
"run --files" should normalize paths to repository root before matching with "files" or "exclude"


wavexx:
### describe your issue

Let's assume you want to exclude a directory from a hook. Using the: ``exclude: ^dir/`` rule seems like the obvious choice, but it can fail to work if calling pre-commit interactively:

```
cd subdir
pre-commit run --files ../dir/file.cpp
```

This is because the path is not canonicalized in respect to the repository root before matching. I consider this behavior to be broken.

- If I'm able to use ``pre-commit run`` anywhere within a git repository, I expect no differences in behavior according the CWD.
- Without canonicalization we cannot build robust rules that match the path prefix in any way.

### pre-commit --version

pre-commit 2.15.0

### .pre-commit-config.yaml

```yaml
-
```

### ~/.cache/pre-commit/pre-commit.log (if present)

-

asottile:
cannot reproduce -- they are normalized first -- please fill out the issue template next time

PR #2073, opened 2021-10-02
Use os.path.realpath() for path canonicalization

wavexx:
os.getcwd() returns the real path of the current working directory. In
a directory which has been symlinked this does not necessarily match the
same prefix of the current directory.

When os.getcwd() is cd-ed to relativize path names a few lines below
this results in paths which are relative to symlinked tree, not to the
real tree. This makes path rule matches fail.

Use os.path.realpath() when making paths absolute, so that relative
paths refer to the same tree as os.getcwd()

This was the underlying issue of #2072

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -159,11 +159,11 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
 def _adjust_args_and_chdir(args: argparse.Namespace) -> None:
     # `--config` was specified relative to the non-root working directory
     if os.path.exists(args.config):
-        args.config = os.path.abspath(args.config)
+        args.config = os.path.realpath(args.config)
     if args.command in {'run', 'try-repo'}:
-        args.files = [os.path.abspath(filename) for filename in args.files]
+        args.files = [os.path.realpath(filename) for filename in args.files]
     if args.command == 'try-repo' and os.path.exists(args.repo):
-        args.repo = os.path.abspath(args.repo)
+        args.repo = os.path.realpath(args.repo)
 
     toplevel = git.get_root()
     os.chdir(toplevel)


asottile:
I can't accept the patch without a reproduction and test

wavexx:
I have a toy repository where I reproduced this. If this plus a repro script is enough, I can provide this right now. Otherwise, I can't dedicate more time on the issue right now - I'll have to wait.

asottile:
I can't see your screen, I don't know why you didn't show that in your issue -- you're not being helpful by avoiding the template which prompts you to do that.

wavexx:
The configuration in my repository was not something I could share without editing (and not really relevant as it needs extra setup). The log is empty.

This is the test case I conjured while trying to reproduce the issue:

[test.tar.gz](https://github.com/pre-commit/pre-commit/files/7271930/test.tar.gz)

see run.sh.

asottile:
I took a look at your reproduction and I don't think it's valid -- you're using `$PWD` which contains the symlink itself as an argument -- pointing outside of the repository (pre-commit should maybe forbid you from running against files outside of your repository, but it currently is mostly harmless to allow that and tedious to prevent it)

we also can't use `realpath` here as it changes the meaning of arguments -- for example if you have a symlink checked in pointing at a directory it would redirect to an unrelated thing:

```pycon
>>> import identify.identify
>>> identify.identify.tags_from_path('sym')
{'symlink'}
>>> import os.path
>>> identify.identify.tags_from_path(os.path.realpath('sym'))
{'directory'}
```

wavexx:
> you're using $PWD which contains the symlink itself as an argument -- pointing outside of the repository

I'm not sure I follow here. They're both pointing to the exact same file within the same repository. This is the path that both `git rev-parse --show-toplevel` and os.getcwd() return.

asottile:
they're not though

when you're in the symlink `$PWD` contains the symlink path, and getcwd() and git rev-parse do not.  having a working directory of a "symlink" is an invention of your shell and not the actual working directory (this is done to be a convenience for shell operations).

so while you're "inside" the symlink (according to your shell) the following things are true:

- `$PWD`: sym/
- `git rev-parse --show-toplevel`: repo/
- `os.getcwd()`: `repo/` 

so from there, $PWD is *outside* the repository as it refers to an unrelated path by the path name (it does not start with the absolute path of the repository) and so in a world where pre-commit would reject files outside the working repository that would be rejected as well

wavexx:
Ok, I see what you mean, and the rationale. Although I can agree with it (and it's definitely a convoluted case - this was caused by some internal tooling doing automated tests), this does not match how git itself is resolving paths - which is the largest disconnect for me.

For git both full paths are reduced to "subdir/file" inside the same tree. You can verify this by substituting "pre-commit run" in the test with "git add -v" or a similar command.

asottile:
if you can come up with a heuristic which doesn't break checked in symlinks it will be considered -- but your current solution does not

wavexx:
Afaik git doesn't accept paths involving symlinks inside the repository itself for this reason:

``fatal: pathspec 'xy' is beyond a symbolic link``

asottile:
you can check in symlinks into a git repository and pre-commit needs to run linters associated with symlinks when passed as `--files` on the commandline

wavexx:
Yes, that's definitely supported also in git, although the symlink is only accepted if that's the last component of a path inside the tree. I/e, you cannot refer to a path through a symlinked directory inside the tree itself. So there's no ambiguity, but it's true realpath incorrectly resolves the last element.

Is there already a test case involving symlinks or similar enough that could adapted to test such a scenario?

PR #2074, opened 2021-10-03
pre_commit: expose fail-fast argument for run subcommand

ljmf00:
This patch exposes the fail-fast feature as a command-line argument for the run subcommand, allowing to run pre-commit to stop at the first failed hook without the need of temporarily writing it to the configuration file when temporarily needed.

Signed-off-by: Luís Ferreira <contact@lsferreira.net>


asottile:
I'd rather have one or the other but not both

ljmf00:
> I'd rather have one or the other but not both

Well, in fact, this is mostly for convenience and I don't think this introduces much complexity. The rationale behind this is to avoid running costly fixers temporarily when changing a beefy file, on development without having to change the configuration file.

asottile:
what's the actual use case?  does `pre-commit run hookid` solve the same problem?  ideally it's best to have an issue discussion before sending a feature patch

tests/commands/run_test.py
```diff
@@ -974,6 +974,12 @@ def test_pass_filenames(
 
 
 def test_fail_fast(cap_out, store, repo_with_failing_hook):
+    ret, printed = _do_run(
+        cap_out, store, repo_with_failing_hook, run_opts(fail_fast=True),
+    )
+    # it should have only run one hook
+    assert printed.count(b'Failing hook') == 1
```
asottile, line 981: - this should be a separate test
- this test doesn't fail on master (it's 1 failing hook independent of fail_fast)

pre_commit/main.py
```diff
@@ -85,6 +85,10 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
         '--all-files', '-a', action='store_true', default=False,
         help='Run on all the files in the repo.',
     )
+    mutex_group.add_argument(
+        '--fail-fast', action='store_true', default=False,
```
asottile, line 89: `store_true` implies `default=False`

ljmf00:
> what's the actual use case?

Where cases that fit the reasons of #240 but not regularly, in the sense that only a few files are costly and are not worth changing the configuration file. In my case, I was running in combination with `--all-files` and that will hit those costly files. I understand that this can't introduce a huge benefit since it is just a convenient thing to have.

> does `pre-commit run hookid` solve the same problem?

I don't think that fits because if the user's configuration has 30 hooks and only one is time-consuming, I would have to run 29 hooks by hand, unless the contrary is possible, e.g. `exclude-hook` or some similar option.

I tried to run pre-commit with a bash trick using a `vfs` file but I couldn't although that is a bit out of the scope of this change.

```bash
pre-commit run --config <(cat .pre-commit-config.yaml <(echo "fail_fast: true"))
```

> ideally it's best to have an issue discussion before sending a feature patch

Noted. If you prefer, I can create one for referencing this and discuss it there.

pre_commit/main.py
ljmf00: Ok, I will change it. My change was based on the previous entry, which is a `store_true` and also has `default=False`.

asottile:
> by hand, unless the contrary is possible, e.g. exclude-hook or some similar option.

that already exists via `SKIP`

tests/commands/run_test.py
ljmf00: Amended.

ljmf00:
> > by hand, unless the contrary is possible, e.g. exclude-hook or some similar option.
> 
> that already exists via `SKIP`

Sure, that covers mostly my practical use case for this. I leave this up to you to close this and reopen it if anyone has a stronger valid use case.

PR #2087, opened 2021-10-11
Add a `files_contain` option to hook configuration

cread:
We have a large existing codebase we're slowly trying to roll out some pre-commit checks for. We're currently using the old traditional hand rolled shell scripts to do this, but we'd love to be able to use `pre-commit` for this instead.

This PR adds a hook selector called `files_contain`, which is applied after all the existing selectors. Only files that contain the given string are passed to the hook.

Using a simple pre-defined string instead of a regular expression allows us to still have reasonable performance when using this option.


asottile:
hi, in the future please discuss features before implementing them

this is too specific to be accepted, but thank you for the patch nonetheless

cread:
Thanks for the fast reply. Where would be the best place to discuss something like?

PR #2093, opened 2021-10-15
Refactor imports to remove duplication.

ShivnarenSrinivasan:
Cleanup repeated import statements.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -2,32 +2,26 @@ import contextlib
 import functools
 import os
 import sys
-from typing import (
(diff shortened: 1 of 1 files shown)


asottile:
it would not, for instance if you go from:

```
from foo import bar
```

and added `from foo import baz` you would have to modify both lines always whereas with the style in `reorder-python-imports` you preserve blame and only introduce a single new line of code (rather than 1 deletion and 4 insertions)

also in general you shouldn't send unprompted style PRs to projects you're not a member of -- especially when that project is about enforcing style 😆

PR #2094, opened 2021-10-16
Ignore BOM for autoupdate migration

fristhon:
PowerShell (v5.1 and before) output is in UTF-16 and also contains BOM (Byte-Order Mark)
pyyaml just support UTF-8 so this makes a problem when `pre-commit sample-config > .pre-commit-config.yaml` runs.

I changed default encoding of migrate_config to 'utf-8-sig' to handle this issue.

https://stackoverflow.com/questions/68487529/how-to-ensure-python-prints-utf-8-and-not-utf-16-le-when-piped-in-powershell
https://docs.python.org/3/library/codecs.html#module-encodings.utf_8_sig

### DIFF
--- a/pre_commit/commands/migrate_config.py
+++ b/pre_commit/commands/migrate_config.py
@@ -39,14 +39,14 @@ def _migrate_sha_to_rev(contents: str) -> str:
 
 
 def migrate_config(config_file: str, quiet: bool = False) -> int:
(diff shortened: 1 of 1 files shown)


asottile:
yeah that isn't really helpful here -- only UTF-8 is supported

PR #2096, opened 2021-10-18
Add 'install-hooks' to COMMANDS_NO_GIT

janeklb:
The `install-hooks` command seems like it does not require a git directory. If that's correct, could it be added to `COMMANDS_NO_GIT`?

My use-case: to build a warmed-up pre-commit image with hook environments ready go. To be used in a CI environment, where the same image can be used for running pre-commit hooks on various different repos (that share the same pre-commit hooks) by mounting the repo into the container. This saves having to add a `git init` in the image :)

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -37,7 +37,7 @@ os.environ.pop('__PYVENV_LAUNCHER__', None)
 
 
 COMMANDS_NO_GIT = {
-    'clean', 'gc', 'init-templatedir', 'install-hooks', 'sample-config'
+    'clean', 'gc', 'init-templatedir', 'install-hooks', 'sample-config',
 }
 
 


asottile:
the git repository is still used to locate the configuration:

```bash
cd pre_commit
pre-commit install-hooks
```

*before*:

```console
$ cd pre_commit
$ pre-commit install-hooks
$
```

*after*

```console
$ cd pre_commit
$ pre-commit install-hooks
An error has occurred: InvalidConfigError: 
=====> .pre-commit-config.yaml is not a file
Check the log at /home/asottile/.cache/pre-commit/pre-commit.log
```

janeklb:
The `.pre-commit-config.yaml` can exist outside of a git repo

asottile:
it can, but the default behaviour is to use the git repo to locate the pre-commit configuration which this PR breaks

janeklb:
This is a very minor thing really -- it's very little overhead to run `git init` (and clean up, if need be!).

janeklb:
> I'd rather not, there isn't a "normal" use case for this and for your case you can just run `git init`

No worries, appreciate you taking a look. 

Have you come across use-cases like this in the past (as described in the PR description)?

janeklb:
> yeah I've seen something like this for a docker image:
> 
> ```dockerfile
> WORKDIR /src
> RUN git init
> COPY .pre-commit-config.yaml .
> RUN pre-commit install-hooks
> 
> # ...
> ```

👍 alright, [that](https://github.com/pre-commit/pre-commit/pull/2096#issuecomment-946263174)'s essentially what I have too.


> https://pre-commit.ci has a similar setup, though done completely differently (through mounts)

meaning you mount the hook environments into the container doing the `pre-commit` check (rather than installing them via the container)?

asottile:
without giving too much away about how it works yes -- though they're always built inside the same container

PR #2097, opened 2021-10-19
Support fail_fast on check level

### LINKED ISSUE #1143: Support fail_fast on checck level
I would love to have fail_fast configuration option on check level not only at the level of the whole configuration file. 

The use case I have in mind - i have one specific pre-commit check that performs docker builds of images that are used by later checks. This is fairly complex build and I have some optimisations in place to build it once and not re-run Docker build . again (because just running Docker build . later on will introduce some delays). I would like to mark that single first check only as "fail_fast". If it fails, none of the other checks should run. But if any other check fails, It should be perfectly OK to run all other tasks without failing fast.

colens3:
This PR adds support for fail_fast on per-hook level. Additional test created to cover new use case, in addition to that the old test was updated due to the change in Hook class (added fail_fast attribute).

Closes #1143

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -70,6 +70,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
     ),
     cfgv.Optional('args', cfgv.check_array(cfgv.check_string), []),
     cfgv.Optional('always_run', cfgv.check_bool, False),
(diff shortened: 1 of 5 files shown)


asottile:
should be fixed on master -- please try rebasing

Anthchirp:
I have now started using this, and it works exactly as you'd expect. Thank you :heart:

PR #2129, opened 2021-11-16
Expose hook stage to hook scripts as env var

Vynce:
This change makes it possible for one hook script to run in several stages and change its behavior based on context.

My primary use case is to write a single hook for Git LFS that runs during the `push`, `post-checkout`, `post-commit`, and `post-merge` stages.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -391,6 +391,9 @@ def run(
     if args.rewrite_command:
         environ['PRE_COMMIT_REWRITE_COMMAND'] = args.rewrite_command
 
+    if args.hook_stage:
+        environ['PRE_COMMIT_HOOK_STAGE'] = args.hook_stage
+
     # Set pre_commit flag
     environ['PRE_COMMIT'] = '1'
 


asottile:
can you speak more about the use case? it sounds like you could instead provide 4 hooks which internally do the right thing (and are selected by `stages`)

Vynce:
Yeah, four hooks would work fine, but projects would need to specify all four in their `.pre-commit-config.yaml` file. This change would allow the four pre-commit hooks to be combined in a single hook. I think it's a worthwhile usability improvement, but not essential.

`.pre-commit-config.yaml` prior to this change:
``` yaml
repos:
  - repo: ssh://host/myhooks
    rev: 1.0.0
    hooks:
      - id: git-lfs-push
      - id: git-lfs-post-checkout
      - id: git-lfs-post-commit
      - id: git-lfs-post-merge
```

`.pre-commit-config.yaml` after this change:
``` yaml
repos:
  - repo: ssh://host/myhooks
    rev: 1.0.0
    hooks:
      - id: git-lfs
```

asottile:
I kinda think the _former_ is better, since it better communicates that you need to set up each step of `git-lfs` when running `pre-commit install --hook-stage ... --hook-stage ...`

Vynce:
This PR may be premature. I'm running into other problems with pre-commit/LFS interaction. I still need to investigate more, but it seems to have something to do with the LFS smudge/clean filters not always running and/or the way pre-commit saves off patches of modified files.

It also looks like only the LFS `pre-push` hook is needed for basic LFS operation. The other three hooks (`post-checkout`, `post-commit`, and `post-merge`) just help manage file locks, which don't seem to be supported by many LFS backends.

PR #2136, opened 2021-11-22
support gitconfig from env

emzeat:
Add exceptions to the git env so externally configured gitconfig
values set via `GIT_CONFIG_KEY_<n>`, `GIT_CONFIG_VALUE_<n>` and
`GIT_CONFIG_COUNT` get passed through. This is e.g. useful when using
the env to inject insteadOf configurations to force use of proxy servers
in a CI environment.

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -41,9 +41,10 @@ def no_git_env(
     return {
         k: v for k, v in _env.items()
         if not k.startswith('GIT_') or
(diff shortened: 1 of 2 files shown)


asottile:
can you link a reference?

emzeat:
> can you link a reference?

Sure, see http://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGCOUNT

asottile (approved):

PR #2152, opened 2021-12-03
Fix typo in "resources/hook-tmpl"

kubijo:
Pretty much self-explanatory here ... I've looked into the installed hook scripts & found out that there was a typo in missing space.

While python doesn't, for some reason, seem to have a problem with it, I think we can agree that it's just cleaner this way.


asottile:
this is not a typo

PR #2154, opened 2021-12-04
fix: Add missing warning for regular expression with [\\/]

### LINKED ISSUE #2151: More on warnings for regular expressions
Hi!

As briefly discussed here:
https://github.com/pre-commit/pre-commit/pull/2053#issuecomment-983436239

`pre-commit` could detect the `[\\/]` regular expression as well.
As of now (pre-commit 2.16.0), such a regex does not trigger any warnings.

Besides, IIRC, the regex `[\/]`, currently detected, matches the single character `/` (not `/` and `\`), so I think it shouldn't be mentioned as an erroneous way for matching both kinds of slashes.
(see https://regex101.com/r/FRPiLZ/1)

Thanks!

kuviokelluja:
test: Test case parameters for said regular expression

refactor: For-loop for regex warnings instead of multiple if statements

resolves #2151

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -144,7 +144,7 @@ class OptionalSensibleRegexAtHook(cfgv.OptionalNoDefault):
                 f"regex, not a glob -- matching '/*' probably isn't what you "
                 f'want here',
             )
-        for fwd_slash_re in [r'[\\/]', r'[\/]', r'[/\\]']:
+        for fwd_slash_re in (r'[\\/]', r'[\/]', r'[/\\]'):
             if fwd_slash_re in dct.get(self.key, ''):
                 logger.warning(
                     fr'pre-commit normalizes slashes in the {self.key!r} '
@@ -162,7 +162,7 @@ class OptionalSensibleRegexAtTop(cfgv.OptionalNoDefault):
                 f'The top-level {self.key!r} field is a regex, not a glob -- '
                 f"matching '/*' probably isn't what you want here",
             )
-        for fwd_slash_re in [r'[\\/]', r'[\/]', r'[/\\]']:
+        for fwd_slash_re in (r'[\\/]', r'[\/]', r'[/\\]'):
             if fwd_slash_re in dct.get(self.key, ''):
                 logger.warning(
                     fr'pre-commit normalizes the slashes in the top-level '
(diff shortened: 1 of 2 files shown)


pre_commit/clientlib.py
```diff
@@ -144,18 +144,13 @@ def check(self, dct: Dict[str, Any]) -> None:
...
-                fr'in hook {dct.get("id")!r} to forward slashes, so you '
-                fr'can use / instead of [\/]',
-            )
-        if r'[/\\]' in dct.get(self.key, ''):
-            logger.warning(
-                fr'pre-commit normalizes slashes in the {self.key!r} field '
-                fr'in hook {dct.get("id")!r} to forward slashes, so you '
-                fr'can use / instead of [/\\]',
-            )
+        for fwd_slash_re in [r'[\\/]', r'[\/]', r'[/\\]']:
```
asottile, line 147: this should be a tuple

while the code was originally written as separate if statements to make sure everything is covered, this is fine I guess

sanjioh, line 147: Also: should `r'[\/]'` trigger the warning at all? The matched character _is_ a forward slash, mentioning "normalization to forward slashes" seems odd in this case.

kuviokelluja: This is something I expected you to comment on, and I'm glad for that. 

Since this was accepted in the [previous pull request regarding these regexes](https://github.com/pre-commit/pre-commit/pull/2053#issuecomment-983436239), without any comments on the subject, I felt like confirming if this is meant to be a style guideline?
In this case, the warning would (sort of) make sense.

kuviokelluja: Fixed the first, and making excuses for the second:
Figured I'd remove some duplication, but if requested, I will revert it back to if statements.

sanjioh: Hello, and thanks for working on this.

Yes, you’re right that pre-commit could warn about `r'[\/]'` as a suggestion about style (I assume the issue didn’t come up in the original PR for an oversight).
But in that case, I’d be more inclined to extract the check into a separate one, with its own proper message.

I understand that this could be quite a nit pick, and that grouping all the regexes under the same message would be equally acceptable, although a little less precise with regards to semantics.

@asottile any thoughts about this?

asottile: yeah so admittedly the original task was a _mistake_ on my part since `r'[\/]'` is the same as `r'[/]'` -- but since there's already several levels of where escaping is questionable (yaml rules, then python rules, then regex rules) I think it's still correct to warn about that

the flip `r'[/\]'` doesn't need a warning because it's an invalid regex (unclosed `]`) so something will already let the user know something is broken

asottile (approved):

PR #2158, opened 2021-12-12
Add Lua language support

mblayman:
This PR adds support for Lua as a language using Luarocks for package management.

I have the Lua tests working locally. I suspect that CI is going to break until I change something about the `azure-pipelines.yml`.

Also, this PR has zero documentation about the new language support. Is documentation done in a different repo? If so, how can I help get something appropriate written?


tests/repository_test.py
```diff
@@ -1128,3 +1128,27 @@ def test_non_installable_hook_error_for_additional_dependencies(store, caplog):
...
+def test_local_lua_additional_dependencies(store):
+    config = {
+        'repo': 'local',
+        'hooks': [{
+            'id': 'local-lua',
+            'name': 'local-lua',
+            'entry': 'lua',
+            'language': 'lua',
+            'args': ['-e', 'require "inspect"; print("hello world")'],
+            'additional_dependencies': ['inspect'],
```
mblayman, line 1149: `inspect` pretty prints tables (because the default print of a table spits out a fairly useless memory address). This seemed like a tame additional dependency to check against.

testing/get-lua.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+sudo apt install luarocks
```
mblayman, line 2: Sigh... the version of `luarocks` available to apt is 2.4.2 which came out in 2016. I think I have to install `luarocks` from source to get past this current error with the build command.

asottile: I think if you bump the version of azure-pipeline-templates it'll upgrade to ubuntu 20.04 which should have a new enough version

mblayman: It's a good thought, but 20.04 isn't going to help unfortunately. 2.4.2 is the version that's available in Ubuntu 20.10. 😒 https://packages.ubuntu.com/impish/luarocks

testing/get-lua.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Install the runtime.
+sudo apt install lua5.3 liblua5.3-dev
+# Ubuntu doesn't add a `lua` binary. Symlink to make the test config easier.
+sudo ln -s /usr/bin/lua5.3 /usr/bin/lua
```
mblayman, line 7: I don't feel great about this line. An Linux user seems unlikely to symlink a specific version to `lua` so this is setting up the test environment for an unrealistic circumstance.

I could update the code to search every directory in the PATH and attempt to find `lua*`. This becomes even more relevant since I had to add in some code that searches for the version number to put in `LUA_PATH` and `LUA_CPATH`. That part of this branch feels brittle to me at the moment and I'm open to suggestions to improve it.

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,104 @@
...
+                os.path.join(env, 'lib', 'lua', version, '?.so;'),
+                Var('LUA_CPATH'),
+            ),
+        ),
+    )
+
+
+def _get_lua_path_version() -> str:
+    """Get the Lua version used in file paths."""
+    returncode, stdout, _ = cmd_output('lua', '-e', 'print(package.path)')
```
mblayman, line 93: This is the line that is brittle if there is only a version-specific lua (e.g., `lua5.3`) on the PATH.

mblayman:
@asottile, things are now working with 100% coverage on POSIX CI.

What strategy would you like me to apply for Windows? I think it's likely to be quite rough to get tests running for Windows (e.g., the Lua docs automatically assume the presence of `make` when building from source so I'm not even sure where to start for this Windows setup. cygwin?). I see some tests that are for POSIX only. Should I do that for the tests? If I did that, I think the coverage on the Windows run would be less than 100% so I'm not sure how to handle that.

I've invested a fair amount of time into this PR, and I'd like to see it merged ultimately because I think it will be valuable for Lua users, so please let me know what I can do to help get this branch where it needs to be.

Thanks in advance for any guidance on this!

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def install_environment(
+    prefix: Prefix,
+    version: str,
+    additional_dependencies: Sequence[str],
+) -> None:
+    helpers.assert_version_default('lua', version)
+    # luarocks, Lua's package manager, is separate from the runtime.
+    helpers.assert_version_default('luarocks', version)
```
asottile, line 29: this doesn't do anything -- you can remove this here -- the first parameter in this call here is just for the error message and doesn't do any additional checking

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+
+    envdir = _envdir(prefix, version)
+    with clean_path_on_failure(envdir):
+        with in_env(prefix, version):
+            _bootstrap_environment(envdir)
+
+            # Sometimes a rockspec is not available (like with a local hook).
```
asottile, line 36: we should add a placeholder rockspec into the empty template -- for example see what I did for [dart](https://github.com/pre-commit/pre-commit/pull/2027/files#diff-8372cd96f090f7f0d3997aa2d99bf35c9d153aa89fa76d389343ca432251471b)

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+def _bootstrap_environment(envdir: str) -> None:
+    """Bootstrap the environment directory.
+
+    luarocks doesn't bootstrap a tree prior to installing
+    so ensure the directory exists.
+    """
+    try:
+        os.mkdir(envdir)
+    except FileExistsError:
+        pass
```
asottile, line 113: this is `os.makedirs(dirname, exist_ok=True)`

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
+import contextlib
```
asottile, line 1: I tend to order functions in the opposite order of what you've written here -- where the helpers are at the top and the api is at the bottom and it calls things above it

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+def get_env_patch(env: str) -> PatchesT:
+    version = _get_lua_path_version()
+    return (
+        ('PATH', (os.path.join(env, 'bin'), os.pathsep, Var('PATH'))),
+        (
+            'LUA_PATH', (
+                os.path.join(env, 'share', 'lua', version, '?.lua;'),
+                os.path.join(env, 'share', 'lua', version, '?', 'init.lua;'),
+                Var('LUA_PATH'),
+            ),
```
asottile, line 77: inheriting LUA_PATH seems maybe not right here -- I think this should just _set_ the value? (that way things outside of the environment can't interfere)

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+        ),
+    )
+
+
+def _get_lua_path_version() -> str:
+    """Get the Lua version used in file paths."""
+    returncode, stdout, _ = cmd_output('lua', '-e', 'print(package.path)')
+    if returncode != 0:
+        return 'unknown'
```
asottile, line 92: this seems kind of fragile -- I think it would be better to raise an error if it can't be detected at all?

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+    match = re.search('share/lua/(.*?)/', stdout)
```
asottile, line 93: I don't think this is going to work on windows with the slashes (?)

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,113 @@
...
+    if match:
+        return match.group(1)
```
asottile, line 95: this is better written as `match[1]`

testing/get-lua.sh
asottile: yeahhhh this is not a good idea, at the very least it should be `/usr/local/bin/lua` -- but it also strikes me as odd that there isn't a non-versioned executable -- then again I haven't written much lua

asottile: maybe pre-commit should try and discovery any executable that looks like `lua-#.#`?  what do lua developers usually do?

testing/get-lua.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Install the package manager. The version available to Ubuntu
+# (as of 12/12/21) is 2.4.2 and fails on the build command.
+LUAROCKS_VERSION="3.8.0"
```
asottile, line 11: is there a way we could make this work with older versions?  seems a little unfortunate to require such a new version of the package manager 🤔

tests/languages/lua_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+from pre_commit.languages import lua
+
+
+def test_existing_bootstrap_environment(tmpdir):
+    """An existing environment is permitted."""
+    lua._bootstrap_environment(str(tmpdir))
+
+
+@mock.patch('pre_commit.languages.lua.cmd_output')
```
asottile, line 11: I prefer to always use `with mock.patch.object(...)` -- `mock.patch('string')` has hard-to-trace side-effects and has been a really annoying source of bugs in the past

asottile (commented):
overall seems fine, I haven't looked at the comments in the actual PR yet so I will address those next!

asottile:
really excited about this!  seems cool!

> Also, this PR has zero documentation about the new language support. Is documentation done in a different repo? If so, how can I help get something appropriate written?

yep -- that'll happen in https://github.com/pre-commit/pre-commit.com

> What strategy would you like me to apply for Windows?

hmm yeah it would be ideal to test it on windows as well -- but that can come later if you'd like.  the tests really only need to work in azure pipelines so I would see if someone else has a lua setup that works well for them

mblayman:
Thanks for the thorough review. I'll try to address the comments in the next couple of days.

asottile:
> Thanks for the thorough review. I'll try to address the comments in the next couple of days.

no rush!  and thanks for working on this!  (I reviewed it on stream today if you want to watch the vod, and gave you a nice shoutout!)

testing/get-lua.sh
mblayman: > what do lua developers usually do?

I have no idea. I'm fairly new to Lua myself. I'll try to make the script do some discovery.

testing/resources/lua_repo/hello-dev-1.rockspec
```diff
@@ -0,0 +1,15 @@
+package = "hello"
+version = "dev-1"
+
+source = {
+   url = "git+ssh://git@github.com/pre-commit/pre-commit.git"
```
mblayman, line 5: I noticed the confused expression on your face on the stream when you got to this line. `luarocks` needs a URL. The value doesn't really matter for the operation of the `build` command, so I just opted to use something related to pre-commit. I could probably put something random here if you'd prefer.

asottile: this is fine, it's just test data anyway

pre_commit/languages/lua.py
mblayman: I'm a "most abstract thing first" kind of guy, but I have no problem switching the order to your preference. 😄

asottile:
a thought I just had -- perhaps `language_version` does make sense for lua if there's potentially many different point releases installed (similar to python?)

testing/get-lua.sh
mblayman: Yeah, you're right, older luarocks support would be better (and, as a bonus, would avoid all this compilation nonsense).

I'll try this with the older version again. I didn't dig in too deeply because the error was pretty weird at the moment (and I don't remember what it was now). Maybe this will go better now that the empty rockspec template is in place.

mblayman:
@asottile, I finally got a passing build by skipping the Windows side for now. I believe I've addressed almost all of your comments now.

Can you help me understand the purpose of the `language_version` stuff? I'm not sure what that's supposed to enable in pre-commit.

Also, it looks like the return for `get_default_version` is either the value of `C.DEFAULT` or the name of the programming language's executable that is discovered. Did I get that right?

I could potentially provide an implementation for `get_default_version`, but I don't understand why that would help right now.

asottile:
so `language_version` is intended to allow hooks to specify that they need a specific version of a programming language

for lua it might look like this:
- `default_language_version()` -> discover all of the `lua` / `lua#.#` executables and pick the newest or most appropriate one
- allow the user to set `language_version` which would then use a specific `lua#.#` executable to do the installation and execution

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,137 @@
...
+    so discovery is needed to find a valid executable.
+    """
+    paths = os.environ.get('PATH', '').split(os.pathsep)
+    possible_luas = []
+    for path in paths:
+        try:
+            candidate_executables = os.listdir(path)
+        except FileNotFoundError:
+            # Invalid path on PATH.
+            continue
```
asottile, line 36: this can be `with contextlib.suppress(FileNotFoundError):` -- though I suspect this should just suppress all `OSError`s since there's possibility for `PermissionError` as well

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,137 @@
...
+
+        for candidate in candidate_executables:
+            if candidate == 'lua':
+                # Found one. Bail early.
+                return os.path.join(path, candidate)
+            elif candidate.startswith('lua'):
+                possible_luas.append((path, candidate))
```
asottile, line 43: this needs to check that they are executable and a a file -- otherwise this will discover non-executables

one way you could do this ~slightly easier would be to collect all the possible basenames and then use `parse_shebang.find_executable` to return the first non-None one -- that'd reuse most of the `PATH` parsing for you

pre_commit/languages/lua.py
```diff
@@ -0,0 +1,137 @@
...
+            # so ensure the directory exists.
+            os.makedirs(envdir, exist_ok=True)
+
+            make_cmd = ['luarocks', '--tree', envdir, 'make']
+            # Older luarocks (e.g., 2.4.2) expect the rockspec as an argument.
+            filenames = [
+                filename for filename in os.listdir(prefix.prefix_dir)
+                if filename.endswith('.rockspec')
+            ]
+            make_cmd.extend(filenames[:1])
```
asottile, line 120: there's `Prefix.star` which does this for you -- it's used in `ruby` as well

pre_commit/store.py
```diff
@@ -188,7 +188,8 @@ def _git_cmd(*args: str) -> None:
 
     LOCAL_RESOURCES = (
         'Cargo.toml', 'main.go', 'go.mod', 'main.rs', '.npmignore',
-        'package.json', 'pre_commit_placeholder_package.gemspec', 'setup.py',
+        'package-dev-1.rockspec', 'package.json',
```
asottile, line 191: package-dev-1.rockspec should probably have a less generic name so it's clear it comes from `pre-commit`

pre_commit/languages/lua.py
mblayman: This try/except is in the for loop so I'm not sure how the code could continue if the exception was suppressed. I think if I used `suppress` and the check failed, then the code could hit an `UnboundLocalError`.

Good call on the `PermissionError`. I'll switch to catching `OSError`.

asottile: oh I see now, I think I missed the indentation

pre_commit/languages/lua.py
mblayman: I'm not sure how to make it work with `find_executable`. I like the idea of reusing the `PATH` parsing, but I don't think I'd be able to construct the possible basenames to check against. The problem that I see currently is that the information about the version isn't available.

For instance, the Ubuntu packaging scenario makes building basenames difficult. The user may not provide a version so pre-commit would be set to use the default. Since the Ubuntu package doesn't provide a `lua` executable and only provides something like `lua-5.3`, there's no way to guess the 5.3 value. The basenames could be `['lua']` in this case and, on Ubuntu, pre-commit would fail to find the executable.

As far as I can tell, sniffing through all the executables with the regex is the only way I can see to get the lua executable in that scenario.

Any ideas on how to work around that scenario?

asottile: basically something like this:

```python
choices = {'lua'}
for path in os.environ.get('PATH', '').split(':'):
    try:
        candidates = os.listdir(path)
    except OSError:
        continue
    else:
        for candidate in candidates:
            if re.match('^lua-\d+\.\d+(\.exe)?$', candidate):
                choices.add(candidate)

found_exes = [
    exe for exe in choices
    if find_executable(exe)
]
# ... pick the one that's newest or is `lua`
```

mblayman: Ah, I see what you mean. I got the impression that you thought their was a way to do this by iterating over the PATH paths one time.

I'll take your approach. I think I can simplify the regex too because the `.exe` extension handling is factored in to `find_executable`. Thanks for the code sketch.

mblayman:
I made `language_version` work with the `_find_lua` function so I think this will now work if users request a different language version.

Assuming the tests pass, I think this is ready for another review. I hope this is getting close to the finish line. 😁

asottile:
alright, apologies for taking so long on this -- I'm going to finish it up for you since I've taken so much of your time here -- thank you!

I did a little research last night and it looks like even though multiple lua versions could be installed, luarocks (or at least the current version of LTS releases) doesn't have a good way of selecting between the versions.  As such, I think we'll leave out `language_version` support for now and just support whatever version `luarocks` indicates it is installing for.

I'll work on implementing this!  thank you for the basis, this is really helpful!

mblayman:
Great! I appreciate the help on this.

My endgame here is to make two hooks: one for [LuaCheck](https://github.com/lunarmodules/luacheck) and one for [LuaFormatter](https://github.com/Koihik/LuaFormatter).

If I can help test this out with a real hook, I'd be happy to be a tester/initial consumer. Since this is a pre-released version of pre-commit on a branch, I would appreciate any config pointers if you have them.

asottile:
alrighty this should be good to go -- I tried it out with `luacheck` (in the test) mostly so I could remove the `lua` binary finding bits -- now the whole thing works just using `luarocks`

asottile (approved):

asottile:
I also checked and it works on windows! (if you can manage to get lua installed which was a big struggle)

PR #2166, opened 2021-12-22
pre_commit: prefer pythonX over pythonX.Y

ljmf00:
This patch prevent reinstallation of pre-commit hooks or any script generated
by shebang function to just update the script shebang, since the scripts are
not heavily dependent on a specific Python 3 version:

    /usr/bin/env: ‘python3.9’: No such file or directory

Signed-off-by: Luís Ferreira <contact@lsferreira.net>

---

Since e622f793c3de2276d6975358c7e35e9b890fbdfd is a thing, shebang is unused, perhaps change this to delete it completely?

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -59,8 +59,8 @@ def shebang() -> str:
         py, _ = os.path.splitext(SYS_EXE)
     else:
         exe_choices = [
-            f'python{sys.version_info[0]}.{sys.version_info[1]}',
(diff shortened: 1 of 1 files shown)


asottile:
hah this code is actually dead and doesn't do anything so your princess is in another castle

ljmf00:
> hah this code is actually dead and doesn't do anything so your princess is in another castle

Sure, then:
> Since e622f79 is a thing, shebang is unused, perhaps change this to delete it completely?

Issue #2189, opened 2022-01-05
PRE_COMMIT_HOME with user home directory (~) is not expanded


tenevdev:
### describe your issue

If `PRE_COMMIT_HOME` is set without the user home directory expanded, such as `PRE_COMMIT_HOME=~/.cache/pre-commit`, the cache is not created in the expected directory. I got this issue in my custom Github workflow and fixed it by not setting `PRE_COMMIT_HOME`.

Does it make sense to just expand [here](https://github.com/pre-commit/pre-commit/blob/cccbc9d2807fb268300d07e71db0d3d332db8daa/pre_commit/store.py#L33) or that's not a common practice? The current code with `os.path.realpath('~/.cache/pre-commit')` just ends up appending it to the working directory afaik which is weird to me.

### pre-commit --version

pre-commit 2.16.0

### .pre-commit-config.yaml

```yaml
n/a
```

### ~/.cache/pre-commit/pre-commit.log (if present)

n/a

asottile:
`~` expansion is a shell feature, pre-commit takes the value literally.

you probably want to use [whatever substitution your CI provides](https://pre-commit.com/#managing-ci-caches) -- or if you're on github just use https://pre-commit.ci

PR #2191, opened 2022-01-05
fix: expand user in the default dir for the store

tenevdev:
fixes https://github.com/pre-commit/pre-commit/issues/2189


asottile:
hi, sorry I do not want this as indicated in https://github.com/pre-commit/pre-commit/issues/2189#issuecomment-1006008824

PR #2207, opened 2022-01-15
Add mamba support to `language: conda`

### LINKED ISSUE #2204: Allow to `mamba install` environments
Thanks for maintaining pre-commit, we really enjoying working with it!

When installing conda environments, [mamba](https://github.com/mamba-org/mamba) is often the preferred option, as it is faster and [requires less memory](https://github.com/conda/conda/issues/5003) when solving an environment.

As of today, pre-commit uses conda for creating environments, as this is hard-coded into the source code:

https://github.com/pre-commit/pre-commit/blob/b944395d6638629a93756ed7279b4368473177af/pre_commit/languages/conda.py#L62-L66

We have a use case, where we want to install pre-commit hooks in CI, but the installation fails because of out-of-memory errors. Installing the same environment with mamba however, succeeds.

It would be nice if one could define whether pre-commit uses conda or mamba for installing environments. This could for example be accomplished by introducing an environment variable `PRE_COMMIT_USE_MAMBA`. If there is a better way to pass such configuration to pre-commit, please let me know.

I'd be glad to work on this feature, if this is something you'd like to see added to pre-commit.

xhochy:
Fixes #2204

### DIFF
--- a/pre_commit/languages/conda.py
+++ b/pre_commit/languages/conda.py
@@ -50,6 +50,15 @@ def in_env(
         yield
 
 
+def _conda_exe() -> str:
+    if os.environ.get('PRE_COMMIT_USE_MICROMAMBA'):
+        return 'micromamba'
+    elif os.environ.get('PRE_COMMIT_USE_MAMBA'):
+        return 'mamba'
+    else:
+        return 'conda'
+
+
 def install_environment(
         prefix: Prefix,
         version: str,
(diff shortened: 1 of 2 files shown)


xhochy:
@janjagusch This is missing tests for the `*mamba` paths. My proposal would be to mock the output of `shutil.which` and check whether the right `conda_exe` is used depending on the environment variables set. I wouldn't install `mamba` / `micromamba` in CI as it isn't a responsibility of `pre-commit` to check whether their install commands are drop-ins. If they don't match, this is a bug in the tools themselves.

asottile:
I went ahead and adjusted this in a few ways and wrote tests:
- if the user opts in via the environment variable don't fall back
- treat any non-empty present variable as set (consistent with the rest of the project)

xhochy:
Thanks @asottile !

Issue #2219, opened 2022-01-24
Improve message for `pre-commit install`


Greedquest:
Running `pre-commit install` gives the message:
 > `pre-commit installed at .git\hooks\pre-commit`

... irrespective of whether or not pre-commit is already set up as a hook. I mistakenly thought it was necessary to run `pre-commit install` every time I added a new entry to my `.pre-commit-config.yaml` because I had to "install" the new package. It would be useful if there was a note saying:

 > `pre-commit already installed at .git\hooks\pre-commit`

So I know through experimentation that I don't need to keep running this command every time.

Greedquest:
No the docs are good and clear and eventually I read them. This is purely a UX feature request - to give a bit more info at the point of running the command so users _don't need_ to use the docs, they can just play around and experiment to use the tool.

I understand that from an implementation p.o.v the script may not ever need to gather info on the presence of a file - as you say it just overwrites what's there/isn't - however if it isn't too costly to add this check purely to improve the CLI output and make it more explicit about what is being done/ what isn't, then I can see that improving the user experience.

By the way I'm talking about vanilla `pre-commit install` without the eager `--install-hooks`.

asottile:
it doesn't hurt to re-run it, I don't think this is worth doing

Greedquest:
The harm is only that a user might waste a bit of time doing it, as I did, but now that I understand the idempotent implementation of the command a bit better I'm inclined to agree, probably not worth the added complexity of checking.

Whether it is good to silently overwrite whatever file is there is another consideration though, and if you ever decided to check whether a custom file exists in the hooks dir already then that would be a good opportunity to add this UX improvement at the same time.

asottile:
it does check if there's a custom file there already, that's what `--force` is for

Greedquest:
Oh I see, sorry should've checked the code first. Ah ok in that case, the script kinda already knows whether it will be creating a new file, `--force`fully overwriting an existing one or effectively a NO-OP, so it shouldn't be any additional cost to inform the user of this.

Guess it just comes down to your personal preference, whether you'd like the user to be able to experiment and learn that way or prefer them to always consult the docs first, your call 🤷

PR #2221, opened 2022-01-24
Modify output text for `pre-commit install` when hook script is already present

Greedquest:
Slightly modify the output message to indicate whether a new hook file was created - see #2219


asottile:
yeah sorry, I don't want to do this as indicated in the issue -- I think it makes it more confusing (and certainly more complicated) rather than less

Greedquest:
@asottile Ok no worries!

PR #2242, opened 2022-02-09
Improving DinD detection

jtyr:
I'm trying to run `docker-image` hook in GitHub Actions inside a Docker image and with the current `pre-commit` implementation it doesn't detect correctly that I'm trying to run it already in a Docker container. The [problem is](https://github.com/pre-commit/pre-commit/blob/master/pre_commit/languages/docker.py#L22-L27) that it's trying to search for the word `docker` in the `/proc/1/cgroup` file which is not there.

This PR is trying to address it by searching the file for `cpuset` followed by path with an ID of a specific length. I think this is valid scenario because the [`_get_container_id` function](https://github.com/pre-commit/pre-commit/blob/master/pre_commit/languages/docker.py#L30-L38) is doing basically the same check.

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -19,7 +19,7 @@ get_default_version = helpers.basic_get_default_version
 healthy = helpers.basic_healthy
 
 
-def _is_in_docker() -> bool:
+def _is_in_docker_cgroup() -> bool:
     try:
         with open('/proc/1/cgroup', 'rb') as f:
             for line in f.readlines():
@@ -37,7 +37,27 @@ def _is_in_docker() -> bool:
         return False
 
 
-def _get_container_id() -> str:
+def _is_in_docker_sched() -> bool:
+    try:
+        with open('/proc/1/sched', 'rb') as f:
+            line = f.readline()
+
+            if line.startswith(b'systemd ') or line.startswith(b'init '):
+                return False
+
+            return True
+    except FileNotFoundError:
+        return False
+
+
+def _is_in_docker() -> bool:
+    if _is_in_docker_cgroup() or _is_in_docker_sched():
+        return True
+
+    return False
+
+
+def _get_container_id_cgroup() -> str:
     # It's assumed that we already check /proc/1/cgroup in _is_in_docker. The
     # cpuset cgroup controller existed since cgroups were introduced so this
     # way of getting the container ID is pretty reliable.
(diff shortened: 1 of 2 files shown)


asottile:
this doesn't seem specific enough -- the example just looks like a normal cgroup and doesn't help us identify a container runtime at all

jtyr:
I have checked several machines (physical and VMs) and non of them have no ID in the path:

```
12:blkio:/init.scope
11:hugetlb:/
10:cpuset:/
9:devices:/init.scope
8:memory:/init.scope
7:rdma:/
6:pids:/init.scope
5:freezer:/
4:net_cls,net_prio:/
3:perf_event:/
2:cpu,cpuacct:/init.scope
1:name=systemd:/init.scope
0::/init.scope
```

Another solution could be to introduce a specific flag on the level of the `pre-commit` config that would allow explicitly specify that the hook is expected to run as Docker in Docker scenario.

pre_commit/languages/docker.py
```diff
@@ -37,15 +37,118 @@ def _is_in_docker() -> bool:
...
 
 
-def _get_container_id() -> str:
+def _is_in_docker_sched() -> bool:
+    try:
+        with open('/proc/1/sched', 'rb') as f:
+            line = f.readline()
+
+            if line.startswith(b'systemd ') or line.startswith(b'init '):
+                return False
```
asottile, line 46: this isn't a valid detection either -- this just checks what process pid1

jtyr: Well, checking for word `docker` anywhere in the cgroups is also not very exact solution.

asottile: the scheduling information is just wrong so that one should not be checked

jtyr: So are you in favor of the `.dockerenv` solution?

jtyr: > the scheduling information is just wrong so that one should not be checked

Btw. why do you think it's wrong? I would rather ignore users who run systemd/initd in Docker (it's bad practice anyway) than make `pre-commit` fail on users who run legitimit containers.

asottile: the identity of pid1 has nothing to do with whether it is running in docker or not

jtyr: > the identity of pid1 has nothing to do with whether it is running in docker or not

The same we could say if there is the word `docker` anywhere in the cgroups.

I still believe that adding more detection methods would make more people happy than unhappy. Or can you think of any other solution that could solve this problem?

jtyr:
Just for the record, I have created `pre-commit` hook that implements what was proposed by this PR:
https://github.com/jtyr/pre-commit-hooks#docker-image

PR #2252, opened 2022-02-17
Use the correct .git path in case of worktrees. 

daschuer:
The issue happens because `get_git_dir()` returns `git rev-parse --git-common-dir` and not `git rev-parse --git-dir` as expected. This PR introduces `get_git_common_dir()` and uses it where the common dir is required and the local .git dir of the current working copy in all other cases.   

This fixes https://github.com/pre-commit/pre-commit/issues/1972, false positives in case of worktrees. 

This PR is in the draft state to debug CI issues.


daschuer:
All green now!

asottile (approved):

PR #2272, opened 2022-03-04
GIT_HTTP_PROXY_AUTHMETHOD kept in env variables

### LINKED ISSUE #2271: Add GIT_HTTP_PROXY_AUTHMETHOD to kept environment variables
### describe your issue

On the cluster I work on, there’s a proxy. It’s… Let’s just say that proxy is being a pain, and to make it work we have to set the environment variable GIT_HTTP_PROXY_AUTHMETHOD.

In pre_commit/git.py however, only a small subset of variables are kept, and that one is not among them. So, sure, I can (and did) edit the script to keep that one too (and it works now) but it’s not ideal.

### pre-commit --version

pre-commit 2.17.0

### .pre-commit-config.yaml

```yaml
Not relevant
```

_No response_

VincentBerthier:
resolves #2271

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -43,6 +43,7 @@ def no_git_env(
         k in {
             'GIT_EXEC_PATH', 'GIT_SSH', 'GIT_SSH_COMMAND', 'GIT_SSL_CAINFO',
             'GIT_SSL_NO_VERIFY', 'GIT_CONFIG_COUNT',
+            'GIT_HTTP_PROXY_AUTHMETHOD',
         }
     }
 


asottile (approved):

asottile:
cc @lorenzwalthert -- any ideas on the R failures here?

lorenzwalthert:
yes, a new version of the package {rlang} got released on CRAN. Since binaries take sometimes a few days to arrive, this CI run tried a source installation, and failed because of some system requirements for compiling code. The problem should go away in a few days. Alternative is to re-write the test to not involve a package like {rlang} that has system requirements as build time dependencies.

asottile:
> yes, a new version of the package {rlang} got released on CRAN. Since binaries take sometimes a few days to arrive, this CI run tried a source installation, and failed because of some system requirements for compiling code. The problem should go away in a few days. ~Alternative is to re-write the test to not involve a package like {rlang} that has system requirements as build time dependencies.~
> 
> Maybe a more general solution that would also serve the user (instead of just hacking a test) would be to always prioritise binary packages over source package when initializing an environment, even if the source package offer a newer version, e.g. with [` options(install.packages.compile.from.source = "never")`](https://community.rstudio.com/t/prefer-install-packages-from-binaries-in-windows/50760/3)

since you know way more about this -- could you help with a PR to add that?

asottile (approved):

PR #2293, opened 2022-03-17
coursier: Add support for both `cs` and `coursier` executable names

Holzhaus:
On some systems, the executable might be named `coursier` instead of
`cs`. For example, this is the case on Arch Linux when using the AUR
package, or when following the official instructions when installing the
JAR-based launcher:

  https://get-coursier.io/docs/cli-installation#jar-based-launcher

### DIFF
--- a/pre_commit/languages/coursier.py
+++ b/pre_commit/languages/coursier.py
@@ -10,6 +10,7 @@ from pre_commit.envcontext import PatchesT
 from pre_commit.envcontext import Var
 from pre_commit.hook import Hook
 from pre_commit.languages import helpers
+from pre_commit.parse_shebang import find_executable
 from pre_commit.prefix import Prefix
 from pre_commit.util import clean_path_on_failure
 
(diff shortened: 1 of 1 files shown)


pre_commit/languages/coursier.py
```diff
@@ -27,6 +27,14 @@ def install_environment(
     helpers.assert_version_default('coursier', version)
     helpers.assert_no_additional_deps('coursier', additional_dependencies)
 
+    # Support both possible executable names (either "cs" or "coursier")
+    executable = next(filter(helpers.exe_exists, ('cs', 'coursier')), None)
```
asottile, line 31: unnecessary FP -- we avoid `map` / `filter` / `reduce` 

this should be `exe('cs') or exe('coursier')` once you find the right function (`helpers.exe_exists` isn't it -- that returns a `bool`)

Holzhaus: > this should be `exe('cs') or exe('coursier')` once you find the right function (`helpers.exe_exists` isn't it -- that returns a `bool`)

Do you want me to write a new one or is there something that I can use? `shutil.which` would be a candiate, or maybe `pre_commit.parse_shebang.find_executable` which kind of duplicates that functionality.

asottile: `find_executable` is the right one probably yes

asottile (approved):
thanks!

PR #2319, opened 2022-03-31
Update ruby-build to latest available (v20220324)

jalessio:
Similar to PRs https://github.com/pre-commit/pre-commit/pull/1913 and https://github.com/pre-commit/pre-commit/pull/2205, updates `ruby-build` to the latest release which  makes more recent versions of Ruby available to pre-commit.

Files were generated by running `./testing/make-archives`.

### ruby-build 20220324

https://github.com/rbenv/ruby-build/releases/tag/v20220324


asottile:
weird, something changed in either gzip or tarfile which makes a different reproducible tar on python3.9 or python3.10 -- I just regenerated using python3.8 to eliminate the diffs

asottile:
apparently this is the change in cpython that causes a different output -- for now I'm going to continue generating these using python3.8 but later in april I'll switch them to the 3.10 generation:  https://github.com/python/cpython/commit/674935b8caf33e47c78f1b8e197b1b77a04992d2

PR #2353, opened 2022-04-22
hook: enable calling out of tree entries

priv-kweihmann:
in the following setup

├── common
│   ├── .pre-commit-config.yaml
│   ├── script1.sh
│   └── script-config
├── repo1
│   └── code_file
└── repo2
    └── code_file

a common configuration from 'common' folder will be installed using an
absolute path to repo1 and repo2.
As local tools needs to be run (script1.sh with its
configuration script-config) we need to pass the path to common
somehow into the hook cmdline.

Add support for a template variable %CONFIG_BASEPATH% which
is automatically expanded to the configuration directory when
pre-commit run is called

This helps to create a single source of truth setup, while still
keeping the possibility to call non public tools and configurations

Signed-off-by: Konrad Weihmann <kweihmann@outlook.com>

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -397,7 +397,7 @@ def run(
 
         config = load_config(config_file)
         hooks = [
-            hook
+            hook.expand(args, config_file)
             for hook in all_hooks(config, store)
             if not args.hook or hook.id == args.hook or hook.alias == args.hook
             if args.hook_stage in hook.stages
--- a/pre_commit/hook.py
+++ b/pre_commit/hook.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import argparse
 import logging
+import os
 import shlex
 from typing import Any
 from typing import NamedTuple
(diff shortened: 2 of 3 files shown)


asottile:
pre-commit intentionally does not support interpolation or inclusion

as is general best practice for open source please in the future search the issue tracker and start with an issue before jumping to implementation

priv-kweihmann:
Then how can one tackle the issue to include configuration?
I'm okay with the tools needed to be installed but configuration **files**, how would one supply them into pre-commit hooks?

asottile:
have you tried `--config`?

priv-kweihmann:
Not for pre-commit itself, but for the tools used in the hook?
For instance I was trying out relint, which works exclusively with config-files.
If I can't include a global one from a template project then I need to copy them into every using repository - making it a maintenance hell (talking about 20+ projects that should use the same tool configuration file)

priv-kweihmann:
What I can think of is to package the needed config files separately and inject them as dependencies for a hook, but how can I reference this installed - can I mix like python dependencies (likely they are installed via pip) with like script dependencies (whose would likely be just http downloads or something) - in my understanding additional_dependencies uses the same install mechanism as the main hook, right?

asottile:
how are you doing it today?

priv-kweihmann:
> how are you doing it today?

basically copying, leading to the effect that always one repo is out of sync - so for upcoming works I'd like to avoid it if there's way to achieve this

asottile:
you also don't even need the thing you're substituting in this PR -- pre-commit will always execute from the root of the repository so if you need to access things relatively outside the repo you can just do so

pre-commit is designed with the fact in mind that most things cannot be uniform, each repository is going to need its own separate configuration and customizations so attempting to "unify" them is a foolish endeavor and there aren't really mechanisms to support that (beyond `--config` and that symlinks "work")

priv-kweihmann:
so you're saying that the only way out would be using docker (one general or a specific container for each hook) - that feels bloated to me, but okay

PR #2169, opened 2021-12-22
To comply with shellcheck

alanwilter:
The `pre-commit` script fails bash linter shellcheck.

PR #2381, opened 2022-05-06
Fix shellcheck warnings

cmuck:
Found some shellcheck warnings

* https://github.com/koalaman/shellcheck/wiki/SC2068
* https://github.com/koalaman/shellcheck/wiki/SC2016

and no shellcheck but when I'm at it...

* https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md034---bare-url-used

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,4 +6,4 @@
 
 A framework for managing and maintaining multi-language pre-commit hooks.
 
-For more information see: https://pre-commit.com/
+For more information see: <https://pre-commit.com/>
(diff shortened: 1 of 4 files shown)


asottile:
dupe #2169

cmuck:
@asottile thanks for the fast feedback, taking the time and having a look at the pr.

Compared to the dupe, SC2016 and MD034 are on-top and valid findings to be fixed.

To SC2068

>your changed output is worse for no improvement (the lint rule you listened to is incorrect)

My change differs a little bit because I did not change the output, so identical to before without becoming worse I think.
It's true that  there is no issue right now because no expression is used atm but I think the finding could be considered, the improvement is "better safe than sorry".

It was a drive-by patch so fine to close the pr for me but maybe you re-think about again ;-)

asottile:
your patch falls under the "no improvement whatsoever" category -- it was not asked for and it does not make things better in any way.  but it does waste your and my time to review it and integrate it

asottile:
and in some ways it makes things *worse*

cmuck:
sry :v: didn't want to waste your time

PR #2384, opened 2022-05-07
Force gem installation into `GEM_HOME`

### LINKED ISSUE #2383: Ruby hooks cannot be installed when `gem` is globally configured with `--user-install`

gem, ruby

### describe your issue

On Arch Linux, `gem` is configured to always use `--user-install` via a global option set in `/etc/gemrc`: https://github.com/archlinux/svntogit-packages/blob/7a52ba210a99eee8df25ad27aa5700bc1f3f10d0/trunk/gemrc

This breaks Ruby tool installations for `pre-commit`, because it makes `gem` ignore `GEM_HOME`.

In a fresh `git` repo:
```console
$ ls -A
.git  .pre-commit-config.yaml  t.rb
$ file t.rb
t.rb: empty
$ pre-commit clean
[...]
$ pre-commit run rubocop -a
[INFO] Initializing environment for https://github.com/rubocop/rubocop.
[INFO] Installing environment for https://github.com/rubocop/rubocop.
rubocop..................................................................Failed
- hook id: rubocop
- exit code: 1

Executable `rubocop` not found

$
```

This can also be reproduced in a fresh docker container using `archlinux:latest` and installing `python-virtualenv python-pip ruby git`.

Adding `--norc` (or `--no-install-user`) to the `gem install` invocation here https://github.com/pre-commit/pre-commit/blob/cc9d950601cd3eba27e8395a7edcd455262705d9/pre_commit/languages/ruby.py#L136-L143 fixes the issue for me (though I haven’t written any tests).

### pre-commit --version

pre-commit 2.19.0

### .pre-commit-config.yaml

```yaml
repos:
  - repo: https://github.com/rubocop/rubocop
    rev: v1.29.0
    hooks:
      - id: rubocop
```

not present

narpfel:
When `--user-install` is set in the gemrc config file, `gem` ignores
`GEM_HOME`. `--no-user-install` prevents this behaviour.

Resolves #2383

### DIFF
--- a/pre_commit/languages/ruby.py
+++ b/pre_commit/languages/ruby.py
@@ -138,6 +138,7 @@ def install_environment(
                 (
                     'gem', 'install',
                     '--no-document', '--no-format-executable',
+                    '--no-user-install',
                     *prefix.star('.gem'), *additional_dependencies,
                 ),
             )
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -332,6 +332,13 @@ def test_run_a_ruby_hook(tempdir_factory, store):
     )
 
 
+def test_run_a_ruby_hook_with_user_install_set(tempdir_factory, store, tmpdir):
+    gemrc = tmpdir.join('gemrc')
+    gemrc.write('gem: --user-install\n')
+    with envcontext((('GEMRC', str(gemrc)),)):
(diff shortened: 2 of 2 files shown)


asottile (approved):
awesome!  and thanks for the test as well :D

Issue #2389, opened 2022-05-13
pre-commit hooks having execution problems on solaris (smartos)


gaige:
### search tried in the issue tracker

solaris

### describe your issue

After using pre-commit on my local workstation for some time, I installed (using pip) on my SmartOS system (Solaris derivative).  Install went fine, and execution without changed files (`pre-commit run`) executed without problems.

As soon as I have a file to commit, or I run `pre-commit run -a`, pre-commit fails with:

    termios.error: (22, 'Invalid argument')

The backtrace is:

````python
Traceback (most recent call last):
  File "/opt/local/lib/python3.9/site-packages/pre_commit/error_handler.py", line 73, in error_handler
    yield
  File "/opt/local/lib/python3.9/site-packages/pre_commit/main.py", line 389, in main
    return run(args.config, store, args)
  File "/opt/local/lib/python3.9/site-packages/pre_commit/commands/run.py", line 416, in run
    return _run_hooks(config, hooks, skips, args)
  File "/opt/local/lib/python3.9/site-packages/pre_commit/commands/run.py", line 285, in _run_hooks
    current_retval, prior_diff = _run_single_hook(
  File "/opt/local/lib/python3.9/site-packages/pre_commit/commands/run.py", line 192, in _run_single_hook
    retcode, out = language.run_hook(hook, filenames, use_color)
  File "/opt/local/lib/python3.9/site-packages/pre_commit/languages/python.py", line 230, in run_hook
    return helpers.run_xargs(hook, hook.cmd, file_args, color=color)
  File "/opt/local/lib/python3.9/site-packages/pre_commit/languages/helpers.py", line 137, in run_xargs
    return xargs(cmd, file_args, **kwargs)
  File "/opt/local/lib/python3.9/site-packages/pre_commit/xargs.py", line 164, in xargs
    for proc_retcode, proc_out, _ in results:
  File "/opt/local/lib/python3.9/concurrent/futures/_base.py", line 609, in result_iterator
    yield fs.pop().result()
  File "/opt/local/lib/python3.9/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/opt/local/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/opt/local/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
[...]

PR #2390, opened 2022-05-14
Switch pty use to fix solaris

gaige:
Use the child instead of parent fd when manipulating pty for color.

Fixes https://github.com/pre-commit/pre-commit/issues/2389


gaige:
Making my way through the tests here. 

- [ ] SmartOS (Solaris)
- [ ] macOS
- [ ] Ubuntu

asottile (approved):

Issue #2391, opened 2022-05-14
Tests fail on systems configured for default branch other than master


gaige:
### search tried in the issue tracker

git master

### describe your issue

While preparing to submit a PR for the solaris issue, I was testing on my desktop which has a default branch of main instead of master.  There are a number of tests that use regex to check the output of git commands which fail if the default branch is anything other than master.  In a CI environment (for now at least) that's likely to work fine, since the default git branch is main.

I'll submit a PR with a small change to use the `--initial-branch=master` argument (available in 2.28+) in order to make sure that tests succeed when used on systems that have a different default branch set, or if git changes the default branch name in the future (current documentation warns that the name is subject to change).

The problem can be readily reproduced by using git config to set `init.defaultbranch=main`

### pre-commit --version

pre-commit 2.19.0

### .pre-commit-config.yaml

```yaml
Not specifically relevant to this bug, see the tox tests
```

PR #2393, opened 2022-05-14
force default branch name for tests

gaige:
Fixes https://github.com/pre-commit/pre-commit/issues/2391

More modern git versions have the ability to set the default branch name, and recent documentation warns against expecting the branch name to be stable. As such, added an explicit set to  `master` (as that was the minimal change and is backwards compatible with older versions of git.

Call is made in a try/except block checking for errors and stepping back to the previous version without initial-branch

### DIFF
--- a/testing/fixtures.py
+++ b/testing/fixtures.py
@@ -38,7 +38,7 @@ def copy_tree_to_path(src_dir, dest_dir):
 
 def git_dir(tempdir_factory):
     path = tempdir_factory.get()
-    cmd_output('git', 'init', path)
+    cmd_output('git', '-c', 'init.defaultBranch=master', 'init', path)
     return path
 
 


testing/fixtures.py
```diff
@@ -38,7 +39,10 @@ def copy_tree_to_path(src_dir, dest_dir):
 
 def git_dir(tempdir_factory):
     path = tempdir_factory.get()
-    cmd_output('git', 'init', path)
+    try:
+        cmd_output('git', 'init', '--initial-branch=master', path)
+    except CalledProcessError:  # pragma: no cover (pre 2.28 git)
+        cmd_output('git', 'init', path)
```
asottile, line 45: without needing special version-specific coverage:

```bash
git -c init.defaultBranch=master init path
```

testing/resources/python3_hooks_repo/build/lib/py3_hook.py
```diff
@@ -0,0 +1,8 @@
+import sys
```
asottile, line 1: oops, [don't run `git add .`](https://youtu.be/GgIZbO2Hb5s)

gaige: Shoot. I thought I'd fixed that.... Will do.

testing/fixtures.py
gaige: OK, didn't realize that worked. I'll fix that when I'm back from running a few errands. Thanks!

gaige:
basically a one-line patch at this point.

asottile (approved):

PR #2407, opened 2022-05-28
Expose prepare-commit-msg arguments as environment vars

### LINKED ISSUE #2406: prepare-commit-msg Not passing all parameters

Yes

### describe your issue

When I run `git commit` with certain arguments I expect to get two or more arguments passed to argv of my script as per the specification:

image
 

```python
def main(argv, argc):
     print(argv, argc)
```

Will print 
```
['tools/hooks/add-commit-template.py', '.git/COMMIT_EDITMSG'] 2
```
It looks to me from looking at your auto-created hook installed that you are passing all args to the script so it maybe I haven't configured this correctly.

Thanks

### pre-commit --version

pre-commit 2.19.0

### .pre-commit-config.yaml

```yaml
repos:
   - repo:
     hooks:
        - id: commit-template-local
          name: Provide Commit Message Template
          entry: tools/hooks/add-commit-template.py
          language: python
          stages: [ prepare-commit-msg ]
```

_No response_

M-Whitaker:
Closes #2406.

This adds the ability to pass environ args that are not files to the prepare-commit-msg hook as defined by the git specification:

[Git hook documentation](https://git-scm.com/docs/githooks#_prepare_commit_msg)

If you want some tests for this more than happy to think up some, however, I have tried to only add to existing code and use optional arguments so all existing tests are not affected.


asottile:
so it shouldn't treat them as arguments, pre-commit only deals in files

they should be environment variables to match up with the other hooks

M-Whitaker:
Ah! I see what you mean now... 

> so it shouldn't treat them as arguments, pre-commit only deals in files
> 
> they should be environment variables to match up with the other hooks

I have changed it accordingly.

pre_commit/commands/hook_impl.py
```diff
@@ -202,8 +206,20 @@ def _run_ns(
...
+        if (len(args) == 2):
+            return _ns(
+                hook_type, color, commit_msg_filename=args[0],
+                commit_msg=args[1],
+            )
+        else:
+            return _ns(
+                hook_type, color, commit_msg_filename=args[0],
+                commit_msg=args[1], commit_sha=args[2],
+            )
```
asottile, line 222: backslashes are gross and parens are unneeded -- I would write this as 4 disparate conditions:

```
elif commit-msg: ...
elif prepare-commit-msg and len(...) == 1: ...
elif prepare-commit-msg and len(...) == 2: ...
elif prepare-commit-msg and len(...) == 3: ...
```

pre_commit/main.py
```diff
@@ -107,6 +107,17 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
         '--commit-msg-filename',
         help='Filename to check when running during `commit-msg`',
     )
+    parser.add_argument(
+        '--commit-msg',
```
asottile, line 111: this is a bad name for this -- perhaps `prepare-commit-message-source` ?  it doesn't contain the commit message itself so calling it `commit-msg` doesn't make sense

pre_commit/main.py
```diff
@@ -107,6 +107,17 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
...
+        help=(
+            'Type of commit string to pass though as ENV' +
+            'during `prepare-commit-msg`'
+        ),
+    )
+    parser.add_argument(
+        '--commit-sha',
```
asottile, line 118: the docs call the the "commit object name"

M-Whitaker:
This was the most elegant way I thought of to keep this underneath the PEP8 line length. Maybe you have an opinion on this...

```python
if args.prepare_commit_message_source:
         environ['PRE_COMMIT_COMMIT_MSG_SOURCE'] \
             = args.prepare_commit_message_source
```

I have implemented all the other name changes as you suggested.

asottile (approved):

Konfekt:
Hello, is there any documentation in pre-commit on these environment vars?

twistfire92:
> Hello, is there any documentation in pre-commit on these environment vars? I suspected a check for `os.getenv("PRE_COMMIT_COMMIT_MSG_SOURCE")` to work, but this is always `None` even when amending, say (see `man 5 githooks`).

@asottile Same problem

@Konfekt did you resolve it?

Konfekt:
Hello, I think I got interested in it for a commit message generation script and resorted to detecting rebases by detecting more than one command-line argument https://gist.github.com/Konfekt/d9640c390deea4beafcf74e29410b8a6#file-chatgpt-write-msg-py-L8 but actually cannot remember at all

Issue #2337, opened 2022-04-10
Ability to include additional hooks 


varac:
First of all I'm super happy with pre-commit, thanks for it!
I'd like to address an issue that currently isn't solved when using pre-commit, which is that there's no way to add local hooks to an existing `pre-commit-config.yaml`.

As an example: I'm working on an open-source project where we found a common ground of pre-commit hooks that result in a shared `.pre-commit-config.yaml`. We couldn't agree on using [conventional commits](https://www.conventionalcommits.org), and decided that this would be optional and anybody could use this convention or not.
After searching through related issues (i.e. #2233, #718, #663) it looks like there is no way to achive this.

I'm *not* proposing any feature that would *install* hooks in `core.hooksPath`, but rather look for ways how to achieve this particular issue, since I also noticed in the issue tracker that a lot of people are looking for similar functionality.
The already suggested `git templatedir` doesn't solve the problem since it's only good for new projects, and even with newly setup project the question remains how you would make changes to hooks in multiple git repos at once.

So I see two ways of achieving this:

1. A global hook that one could install manually using `core.hooksPath`, which would run first and then execute the pre-commit configured in the project
2. A way of including another pre-commit yaml config from `.pre-commit-config.yaml`, i.e 

```
include:
  - ~/.pre-commit-config.yaml
```

which would merge both configs. This is imo the best (and easiest) solution, and gives users the freedom to include custom, local hooks that can't get included in the main config for one reason or another (no matter if you agree with them or not).
3. Maybe there's another way I haven't thought of ?

What do you think ?

varac:
Oh, I just realized that the exact feature was already requested 2 days ago (#2335).

@asottile There you wrote:

> inclusion is intentionally not supported and will not be supported -- please search the issue tracker next time there are many duplicates

Could you please summarize why you don't want to support this, given the `many duplicated` which indicates that this is an important feature for may users ? I mean, you don't need to agree with every user and don't need to make use of it, but please consider giving other users the freedom to use pre-commit their way.

asottile:
you can share those by putting them in a `.pre-commit-hooks.yaml` and using it as a repository. this is the normal way to do things, `local` is the escape hatch

varac:
> you can share those by putting them in a .pre-commit-hooks.yaml and using it as a repository. this is the normal way to do things

[...]

Issue #2335, opened 2022-04-08
allow yaml !include to have a distributed config file


christian-steinmeyer:
Hi!
Perhaps I am missing something, but I would like to be able to import content from another config file in my `.pre-commit-config.yaml`. Something like

```yaml
!include some-shared-pre-commit-config.yaml

repos:
  - # ...
```

Right now, the behavior when trying this is
```bash
An error has occurred: InvalidConfigError: 
==> File .pre-commit-config.yaml
=====> mapping values are not allowed in this context
  in "<unicode string>", line 3, column 6
Check the log at <CACHE_DIR>/pre-commit/pre-commit.log

```

Note, I am not looking for any logic on the level of pre-commit, but rather the support of an (almost) standard feature of yaml files, something like [pyyaml-include](https://github.com/tanbro/pyyaml-include).

I understand, that you are hesitant against the distribution of the config file (#554, #548), but in our case this would be actually really helpful. In contrast to the referenced issues, I also don't think, this would come with a lot of effort. What do you think? I'd be happy to support with docs or similar.

asottile:
inclusion is intentionally not supported and will not be supported -- please search the issue tracker next time there are many duplicates

christian-steinmeyer:
I did (see issue references above) and couldn't find any duplicates for a plain yaml include. Would you mind sharing one of them that contains the rational of why it's not supported?

PR #2422, opened 2022-06-15
Add PRE_COMIMT_ARGV environment variable

jordiclariana:
This pull request adds an environment variable called `PRE_COMMIT_ARGV` cointaining the actual `pre-commit` arguments (escaped).

The main reason to add this environment variable is so the arguments can be used on inner `pre-commit` executions. This use case is for requests like [this](https://github.com/pre-commit/pre-commit/issues/2337) and [this](https://github.com/pre-commit/pre-commit/issues/2335). Having this env var allows you to have something like:

### main repository

 `.pre_commit_config.yaml`:
```yaml
repos:
- repo: git@github.com:mycompany/pre-commit.git
  rev: v0.1
  hooks:
    - id: custom_hooks
```

### `mycompany/pre-commit` repository
`.pre-commit-hooks.yaml`:
```yaml
- id: required
  name: Custom checks
  description: Custom checks
  entry: pre_commit_hooks/custom_hooks.sh
  language: script
  verbose: true
  pass_filenames: false
```

`pre_commit_hooks/custom_hooks.sh`:
```bash
#!/usr/bin/env bash

CUSTOM_HOOKS_DIR="$(dirname "$(readlink -f "$0")")"

eval pre-commit $PRE_COMMIT_ARGV -c "$CUSTOM_HOOKS_DIR/pre-commit-config.yaml"
```

`pre_commit_hooks/pre-commit-config.yaml`:
```yaml
repos:
- repo: https://github.com/antonbabenko/pre-commit-terraform
  rev: v1.64.0
  hooks:
    - id: terraform_fmt
    - id: terraform_tflint
```

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import logging
 import os
+import re
 import sys
 from typing import Sequence
 
@@ -345,6 +346,8 @@ def main(argv: Sequence[str] | None = None) -> int:
     if len(argv) == 0:
         argv = ['run']
     args = parser.parse_args(argv)
+
+    os.environ['PRE_COMMIT_ARGV'] = ' '.join([re.escape(x) for x in argv])
 
     if args.command == 'help' and args.help_cmd:
         parser.parse_args([args.help_cmd, '--help'])


asottile:
neither of those requests are supported workflows

asottile:
I'm not going to add and maintain functionality that only provides value for unsupported workflows.  that's what unsupported means

not to mention, your patch is completely untested and isn't even correct (`re.escape` doesn't have anything to do with commandline) -- a proper design would probably need an actual serialization format (such as `json`) -- and it's not worth having these discussions because I'm not accepting the feature in the first place.

the general best practice for open source is to start with issues for proposing new features rather than jumping to solutioneering

jordiclariana:
Ok, thanks anyway.

PR #2459, opened 2022-07-12
adjust relative --commit-msg-filename if in subdir

asottile:


asottile:
@lorenzwalthert any ideas for the windows R failure here?

lorenzwalthert:
Seems again related to missing built-time dependencies for source install, this time for transversive dependencies. Trying to enforce with #2460.

asottile:
huh looks like I should have just clicked retry because it appears to be working now!  thanks again for checking on this!

lorenzwalthert:
Yeah, because the dependency (without version specification) was on a package  that was not yet available as a binary.

PR #2460, opened 2022-07-12
Enforce binary installs also for dependencies in R

lorenzwalthert:
Should solve problems detected in #2459.


lorenzwalthert:
Can we get this merged? It makes R more robust to install old binaries instead of new source versions that may need system dependencies for the build.

PR #2479, opened 2022-08-10
respect aliases SKIP when installing environments

### LINKED ISSUE #2478: SKIP skips invocation by alias but doesn't skip "Installing environment"

SKIP alias

### describe your issue

When I run 

```sh
SKIP=pylint-alias2 git commit
```

the hook aliased as `pylint-alias2` is properly skipped, however, its environment is still initialized. I would expect that it also skips the lengthy initialization of the environment.

Indeed, running a single hook ([run.py#153](https://github.com/pre-commit/pre-commit/blob/49f95b9ef391639d09136666be375d99a3dccd04/pre_commit/commands/run.py#153)) checks against `hook.id` and `hook.alias`, however `install_hook_envs` is invoked with only filtering against `hook.id` ([run.py#423](https://github.com/pre-commit/pre-commit/blob/49f95b9ef391639d09136666be375d99a3dccd04/pre_commit/commands/run.py)

Is this inconsistency deliberate? I couldn't find a use case for it. This quick and dirty patch works for us:

```diff
-        to_install = [hook for hook in hooks if hook.id not in skips]
+        to_install = [hook for hook in hooks if hook.id not in skips and hook.alias not in skips]
```

In our use case Initialization is not needed for everyone (monorepo where some non-python devs also do non-python work) and also very expensive unfortunately (6minutes currently to install already cached pip packages)

(An even better solution would be if initialization wouldn't happen at all if the hook run would we skipped in general (because of `SKIP` variable or because there is no files changed, however, that would be a but bigger patch I suppose))

### pre-commit --version

pre-commit 

### .pre-commit-config.yaml

```yaml
repos:
  - repo: https://github.com/PyCQA/pylint
    rev: v2.14.5
    hooks:
      - name: pylint-alias1
        alias: pylint-alias1
        id: pylint
        additional_dependencies:
          - arrow
      - name: pylint-alias2
        alias: pylint-alias2
        id: pylint
        additional_dependencies:
          - hdfs
```

_No response_

kmARC:
Fixes #2478

### DIFF
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -618,10 +618,18 @@ def test_skip_bypasses_installation(cap_out, store, repo_with_passing_hook):
         'hooks': [
             {
                 'id': 'skipme',
-                'name': 'skipme',
+                'name': 'skipme-1',
                 'entry': 'skipme',
+                'alias': 'skipme-1',
                 'language': 'python',
                 'additional_dependencies': ['/pre-commit-does-not-exist'],
+            },
+            {
+                'id': 'skipme',
+                'name': 'skipme-2',
+                'entry': 'skipme',
+                'alias': 'skipme-2',
+                'language': 'python',
             },
         ],
     }
@@ -633,6 +641,13 @@ def test_skip_bypasses_installation(cap_out, store, repo_with_passing_hook):
         {'SKIP': 'skipme'},
     )
     assert ret == 0
+
+    ret, printed = _do_run(
+        cap_out, store, repo_with_passing_hook,
+        run_opts(all_files=True),
+        {'SKIP': 'skipme-1'},
+    )
+    assert ret == 1
 
 
 def test_hook_id_not_in_non_verbose_output(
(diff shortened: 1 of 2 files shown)


tests/commands/run_test.py
```diff
@@ -634,6 +642,13 @@ def test_skip_bypasses_installation(cap_out, store, repo_with_passing_hook):
     )
     assert ret == 0
 
+    ret, printed = _do_run(
+        cap_out, store, repo_with_passing_hook,
+        run_opts(all_files=True),
+        {'SKIP': 'skipme-1'},
+    )
+    assert ret == 1
```
asottile, line 650: this test now validates two separate things -- it should be two separate tests

asottile (commented):
just future open source advice, don't make PRs from your `main` branch as it makes it significantly more difficult to review / fix

tests/commands/run_test.py
kmARC: > just future open source advice, don't make PRs from your `main` branch as it makes it significantly more difficult to review / fix

Sure. I was in the gerrit workflow mindset.

I can put it in a separate test method.

I see the CI failing, should I rebase on `v2.20.0` or is this fine?

kmARC:
I am closing this PR and will open one from our company GH account.

PR #2480, opened 2022-08-11
respect aliases in SKIP when installing environments

### LINKED ISSUE #2478: SKIP skips invocation by alias but doesn't skip "Installing environment"

kmARC:
Fixes #2478

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -420,7 +420,11 @@ def run(
             return 1
 
         skips = _get_skips(environ)
-        to_install = [hook for hook in hooks if hook.id not in skips]
+        to_install = [
+            hook
+            for hook in hooks
+            if hook.id not in skips and hook.alias not in skips
+        ]
         install_hook_envs(to_install, store)
 
         return _run_hooks(config, hooks, skips, args)
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -635,6 +635,32 @@ def test_skip_bypasses_installation(cap_out, store, repo_with_passing_hook):
     assert ret == 0
 
 
+def test_skip_alias_bypasses_installation(
+        cap_out, store, repo_with_passing_hook,
+):
+    config = {
+        'repo': 'local',
+        'hooks': [
+            {
+                'id': 'skipme',
+                'name': 'skipme-1',
+                'alias': 'skipme-1',
+                'entry': 'skipme',
+                'language': 'python',
+                'additional_dependencies': ['/pre-commit-does-not-exist'],
+            },
+        ],
+    }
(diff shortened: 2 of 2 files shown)


asottile:
augh the company account is even worse because I can't fix the branch at all!

asottile (approved):

kmARC:
> augh the company account is even worse because I can't fix the branch at all!

Hmmm... I'm not sure what do you mean by that.

I'm planning another PR that helps a lot with monorepos, but that would also come from teh company account. Do you have any guideline on how to make it easier for you?

asottile:
(1) please do not send feature PRs without first discussing them -- monorepos have been a hot topic and I have already set my expectations that there will not be special support for them

(2) when you make PRs from organizations I cannot fixup the branch like I can with user accounts

asottile:
and for (1) make sure to search around first -- I'm almost certain it's about `cd` and there's several discussions about this already

PR #2482, opened 2022-08-15
Remove `.` from `Installing environment for` log

sobolevn:
This is a minor thing, but a bit frustrating.
Some services / terminals like xterm, hyper, and github actions detect links:
Снимок экрана 2022-08-15 в 12 14 15

But, when clicked, they lead to a url with `.` in it, and it does not exist:
Снимок экрана 2022-08-15 в 12 14 26


asottile:
there's a duplicate somewhere -- arguably this is a bug in your terminal emulator and on some you can adjust the clickable pattern

I don't really want to make this change

PR #2489, opened 2022-08-18
Remove period from 'Stashing unstaged files to {patch_filename}.'

digitaltembo:
This extra period makes it ever so slightly more difficult to copy the path in the rare circumstance where I actually care about the path.


asottile:
dupe #2482

Issue #969, opened 2019-03-14
Colors


ghostsquad:
I find the white on yellow color for hooks that have passed hard to read. Also, yellow typically means warning. Can the colors be configurable?

asottile:
`Passed` is green not yellow, it uses whatever color your terminal emulator has configured for green.

For instance, this is what it looks like on ubuntu:

https://asottile.github.io/pygments-pre-commit/demo/test_basic_pre_commit.html

ghostsquad:
Hmm but can this be made configurable? Like maybe the ability to remove the
background color?

Cheers,
Wes

asottile:
yes, you can disable colors with [`--color=never`](https://pre-commit.com/#cli)

avatar-lavventura:
> yes, you can disable colors with [`--color=never`](https://pre-commit.com/#cli)

Is it possible to change the font color?

my precommit `...PASSED` and `...Skipped` colors are as follows (font color is white where it is not readable on top of green background):

[![enter image description here][1]][1]

Here https://pre-commit.com/#usage, its much more readable:

[![enter image description here][2]][2]

**[Q]** How could I change the font and background color of `Passed` and `Skipped` that is located at the end of each check?

  [1]: https://i.stack.imgur.com/uQPvL.png
  [2]: https://i.stack.imgur.com/VEfns.png

asottile:
the font color comes from your terminal emulator -- I have no control over it but you do!

avatar-lavventura:
@asottile I do not know how does the readable colours in my local terminal config is set. Small example would be helpful

asottile:
I literally can't know, I don't know what terminal emulator you are using there are thousands of them.  it's probably in your terminal preferences somewhere not in a dotfile on disk

avatar-lavventura:
@asottile  I am using `zsh`

asottile:
that's your shell, your terminal emulator is the program you launch on your computer (such as iterm or windows terminal or gnome-terminal, etc.)

avatar-lavventura:
I am using `iTerm` that has dark background theme along with white foreground color. Does `Passed` pulling the foreground color? 

My configuration:
<img width="551" alt="image" src="https://user-images.githubusercontent.com/18537398/116809340-3169a680-ab46-11eb-9032-964628d9d694.png">

avatar-lavventura:
I assume there is a configuration bug in pre-commit end that it is configured for the white background theme and dark background theme is ignored.

[...]

PR #2495, opened 2022-08-23
Use reversed colors for better legibility

dhruvkb:
# Problem

The current set of colors have the following problems.

- The green and red texts are hard to read because the grey color of the theme text is not readable against those backgrounds.
- The yellow and turquoise texts look weird on light backgrounds due to the use of dark text.

These issues have been raised before: #1897, #969

## Screenshots

**Dark theme:**
Screenshot 2022-08-23 at 11 22 36 AM

**Light theme:**
Screenshot 2022-08-23 at 11 22 56 AM

# Solution

This PR changes the ANSI codes to use the colors on the foreground (using ANSI codes 3x) and then reverses them with the background (using ANSI code 7). This makes sure that text is always readable and looks harmonious with the background.

## Screenshots

**Dark theme:**
Screenshot 2022-08-23 at 11 26 58 AM

**Light theme:**
Screenshot 2022-08-23 at 11 27 41 AM

The new colors are legible in both modes.

### DIFF
--- a/pre_commit/color.py
+++ b/pre_commit/color.py
@@ -56,10 +56,10 @@ if sys.platform == 'win32':  # pragma: no cover (windows)
 else:  # pragma: win32 no cover
     terminal_supports_color = True
 
-RED = '\033[41m'
-GREEN = '\033[42m'
-YELLOW = '\033[43;30m'
-TURQUOISE = '\033[46;30m'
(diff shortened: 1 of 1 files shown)


asottile:
sorry but as I'm sure you've read in the things you've linked I am not interested in changing the branding

PR #2506, opened 2022-09-07
Add shell completions by shtab

### LINKED ISSUE #1119: Completions for CLI
Following up on #1118, it would be great to have cross-shell completions for the CLI. 

These could be statically autogenerated from the output of --help options.

Would be great to have
1. cross-shell completions (bash, zsh, fish, powershell)
2. help introspection where supported i.e. 
```
pre-commit i<TAB>
install  (Install the pre-commit script)  install-hooks  (Install hook environments...)
```

3. complete for installed hooks (i.e.
```
pre-commit run check-<TAB>
check-added-large-files (Prevent giant files from being committed.)
check-ast   (Simply check whether the files parse as valid python.)
check-toml      (This hook checks toml files for parseable syntax.)
```
4. a command line option to install the completions (i.e. `pre-commit install-completions --shell zsh`).

Freed-Wu:
Maybe fix #1119. Just offer a few commonplace remarks by way of introduction in the
hope that others may come up with valuable opinions.

> complete for installed hooks

Not fixed.

> don't really want to add a dependency for this

Introduce an extras_require. Even if shtab is not installed, it still can work.

```sh
pre-commit --print-completion bash | sudo tee /usr/share/bash-completion/completions/pre-commit
pre-commit --print-completion zsh | sudo tee /usr/share/zsh/site-functions/_pre-commit
pre-commit --print-completion tcsh | sudo tee /etc/profile.d/pre-commit.completion.csh
```

### DIFF
--- a/pre_commit/_shtab.py
+++ b/pre_commit/_shtab.py
@@ -1,0 +1,16 @@
+from __future__ import annotations
+
+from argparse import ArgumentParser
+from typing import Any
+
+FILE = None
+DIRECTORY = DIR = None
+
+
+def add_argument_to(
+    parser: ArgumentParser, *args: list[Any], **kwargs: dict[Any, Any],
+) -> ArgumentParser:
+    from argparse import Action
+
+    Action.complete = None  # type: ignore
+    return parser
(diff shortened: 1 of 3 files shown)


asottile:
as stated in #1119 I do not want to add a dependency for this

asottile:
especially one which adds a bunch of monkeypatch hacks

Issue #2243, opened 2022-02-10
Switch to `tomli`


dhruvkb:
`pre-commit` seems to have a dependency on `toml`.
https://github.com/pre-commit/pre-commit/blob/7858ad066f2dfd11453c2f7e25c8f055ba4de931/setup.cfg#L30

That package
- has not been updated for longer than a year
- is not compatible with Python 3.10

Most of the Python community has shifted to `tomli` ([PyPI](https://pypi.org/project/tomli/), [repo](https://github.com/hukkin/tomli)) instead. Could pre-commit also do the same?

I'm happy to contribute to this.

dhruvkb:
@asottile if switching to `tomli` is not something you'd even consider, could you please help me with this error? Installing `pre-commit` on a Python 3.10 project with Poetry does not work because `toml` breaks the installation.

```
Using version ^2.17.0 for pre-commit

Updating dependencies
Resolving dependencies... (0.5s)

Writing lock file

Package operations: 19 installs, 0 updates, 0 removals

  • Installing distlib (0.3.4)
  • Installing filelock (3.4.2)
  • Installing platformdirs (2.5.0)
  • Installing pyparsing (3.0.7)
  • Installing six (1.16.0)
  • Installing attrs (21.4.0)
  • Installing cfgv (3.3.1)
  • Installing identify (2.4.9)
  • Installing more-itertools (8.12.0)
  • Installing nodeenv (1.6.0)
  • Installing packaging (21.3)
  • Installing pluggy (0.13.1)
  • Installing py (1.11.0)
  • Installing pyyaml (6.0)
  • Installing toml (0.10.2): Failed

  RuntimeError

  Invalid hashes (sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855) for toml (0.10.2) using archive toml-0.10.2-py2.py3-none-any.whl. Expected one of sha256:806143ae5bfb6a3c6e736a764057db0e6a0e05e338b5630894a5f779cabb4f9b, sha256:b3bda1d108d5dd99f4a20d24d9c348e91c4db7ab1b749200bded2f839ccbe68f.

  at /Library/Python/3.8/site-packages/poetry/installation/executor.py:627 in _download_link
      623│                     )
      624│                 )
      625│
      626│             if archive_hashes.isdisjoint(hashes):
    → 627│                 raise RuntimeError(
[...]

PR #2516, opened 2022-09-16
Migrate from dependency toml because it has not yet reached v1.0

cclauss:
https://pypi.org/project/toml -- Currently v0.10.2 is still zero-ver and does not align with the recommendations at https://docs.python.org/3.11/library/tomllib.html

This pull request recommends the adoption of:
* https://pypi.org/project/tomli -- Currently v2.0.1
* https://pypi.org/project/tomli-w -- Currently v1.0.0


asottile:
dupe #2243

PR #2524, opened 2022-09-23
extend warning if globs are used instead of regex to local hooks

### LINKED ISSUE #2521: warnings for regexes don't apply to `repo: local`
looks like these were left out -- should be a pretty easy patch similar to #1707 / #1750

chrisRedwine:
Resolves #2521 
Apologies for the diff - those two classes were just moved up w/o modification

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -298,6 +298,14 @@ CONFIG_HOOK_DICT = cfgv.Map(
     OptionalSensibleRegexAtHook('files', cfgv.check_string),
     OptionalSensibleRegexAtHook('exclude', cfgv.check_string),
 )
+LOCAL_HOOK_DICT = cfgv.Map(
+    'Hook', 'id',
+
+    *MANIFEST_HOOK_DICT.items,
+
+    OptionalSensibleRegexAtHook('files', cfgv.check_string),
+    OptionalSensibleRegexAtHook('exclude', cfgv.check_string),
+)
 CONFIG_REPO_DICT = cfgv.Map(
     'Repository', 'repo',
 
@@ -308,7 +316,7 @@ CONFIG_REPO_DICT = cfgv.Map(
         'repo', cfgv.NotIn(LOCAL, META),
     ),
     cfgv.ConditionalRecurse(
-        'hooks', cfgv.Array(MANIFEST_HOOK_DICT),
+        'hooks', cfgv.Array(LOCAL_HOOK_DICT),
         'repo', LOCAL,
     ),
     cfgv.ConditionalRecurse(
--- a/tests/clientlib_test.py
+++ b/tests/clientlib_test.py
@@ -15,6 +15,8 @@ from pre_commit.clientlib import DEFAULT_LANGUAGE_VERSION
 from pre_commit.clientlib import MANIFEST_SCHEMA
 from pre_commit.clientlib import META_HOOK_DICT
 from pre_commit.clientlib import MigrateShaToRev
+from pre_commit.clientlib import OptionalSensibleRegexAtHook
+from pre_commit.clientlib import OptionalSensibleRegexAtTop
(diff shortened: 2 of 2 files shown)


pre_commit/clientlib.py
```diff
@@ -81,6 +118,8 @@ def _make_argparser(filenames_help: str) -> argparse.ArgumentParser:
     cfgv.Optional('require_serial', cfgv.check_bool, False),
     cfgv.Optional('stages', cfgv.check_array(cfgv.check_one_of(C.STAGES)), []),
     cfgv.Optional('verbose', cfgv.check_bool, False),
+    OptionalSensibleRegexAtHook('files', cfgv.check_string),
+    OptionalSensibleRegexAtHook('exclude', cfgv.check_string),
```
asottile, line 122: we don't want to warn users about things outside their control (remote repositories doing this incorrectly)

chrisRedwine: makes sense, thanks for the direction. I think I have a better understanding of the `cfgv` stuff after looking at it a bit, so I'll post a revision here in a minute.

chrisRedwine: I reworked it per your call-out. 

Also noticed that the existing validators for this were calling `super().check` and passing a check func of `check_string`, which as far as I can tell is superfluous work since there are already validators on `CONFIG_HOOK_DICT` and `CONFIG_SCHEMA` for `files` & `exclude` which use `check_string_regex`. So, I removed the `super().check` calls and just passed `check_any` instead, similar to how you did the `NotAllowed` validator. Let me know if that doesn't jive, though. 

Thanks!

asottile: compare:

```python
x = OptionalSensibleRegexAtHook('files', cfgv.check_string)
x.check({'files': None})
```

with super:

```console
$ python3 t.py 
  File "/tmp/add-trailing-comma/t.py", line 25, in <module>
    x.check({'files': None})
  File "/tmp/add-trailing-comma/t.py", line 7, in check
    super().check(dct)
  File "/tmp/add-trailing-comma/venv/lib/python3.10/site-packages/cfgv.py", line 53, in _check_optional
    with validate_context(f'At key: {self.key}'):
  File "/usr/lib/python3.10/contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/tmp/add-trailing-comma/venv/lib/python3.10/site-packages/cfgv.py", line 34, in validate_context
    raise ValidationError(e, ctx=msg).with_traceback(tb) from None
  File "/tmp/add-trailing-comma/venv/lib/python3.10/site-packages/cfgv.py", line 31, in validate_context
    yield
  File "/tmp/add-trailing-comma/venv/lib/python3.10/site-packages/cfgv.py", line 54, in _check_optional
    self.check_fn(dct[self.key])
  File "/tmp/add-trailing-comma/venv/lib/python3.10/site-packages/cfgv.py", line 325, in check_type_fn
    raise ValidationError(
cfgv.ValidationError: 
==> At key: files
=====> Expected string got NoneType
```

without super:

```console
$ python3 t.py 
  File "/tmp/add-trailing-comma/t.py", line 25, in <module>
    x.check({'files': None})
  File "/tmp/add-trailing-comma/t.py", line 9, in check
    if '/*' in dct.get(self.key, ''):
TypeError: argument of type 'NoneType' is not iterable
```

chrisRedwine: roger that - made the update

chrisRedwine: I added a test for the counterexample. Definitely need to make sure I can commit the same amount of attention I would to closed source in poking around / exploring edge cases / understanding the code base / writing tests / etc., rather than letting the novelty & excitement of open source tempt me to just grab a ticket and push something quickly. So, I apologize & appreciate the patience here - I'll do better!

asottile (approved):

Issue #2525, opened 2022-09-24
 Executable `python` not found when committing new file


swiss-knight:
### search tried in the issue tracker

Executable `python` not found

### System information
Ubuntu 22.04.1
Python 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0] on linux
git version 2.34.1
pre-commit 2.20.0

Running this basic git command:
```
$ git add foo.py
$ git commit -m "Update foo.py"
```

leads to the error log hereunder. 

Please, also notice that `python` is no more a valid command on Ubuntu 22,04:
```
$ python
Command 'python' not found, did you mean:
  command 'python3' from deb python3
  command 'python' from deb python-is-python3
  
$ which python3
/usr/bin/python3
```

### pre-commit --version

2.20.0

### .pre-commit-config.yaml

```yaml
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
default_language_version:
    python: python3.10

repos:

  - repo: https://github.com/pycqa/isort
    rev: "5.10.1"
    hooks:
      - id: isort
        args: ["--profile", "black"]

  - repo: https://github.com/psf/black
    rev: "22.8.0"
    hooks:
      - id: black
        language_version: python3.10
```

### ~/.cache/pre-commit/pre-commit.log (if present)

$ cat /home/username/.cache/pre-commit/pre-commit.log

### version information

```
pre-commit version: 2.20.0
git --version: git version 2.34.1
sys.version:
    3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
sys.executable: /usr/bin/python3
os.name: posix
sys.platform: linux
```

### error information

```
An unexpected error has occurred: CalledProcessError: command: ('python', '-mpip', 'install', '.')
return code: 1
expected return code: 0
stdout:
    Executable `python` not found
stderr: (none)
```

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/pre_commit/error_handler.py", line 73, in error_handler
    yield
  File "/usr/local/lib/python3.10/dist-packages/pre_commit/main.py", line 358, in main
    return hook_impl(
  File "/usr/local/lib/python3.10/dist-packages/pre_commit/commands/hook_impl.py", line 254, in hook_impl
    return retv | run(config, store, ns)
  File "/usr/local/lib/python3.10/dist-packages/pre_commit/commands/run.py", line 424, in run
[...]

Issue #2503, opened 2022-08-26
Bug during installing environment


cawoon:
### search tried in the issue tracker

Looked for all similar issues with CalledProcessError, tried changing virtualenv version a lot of times, doesn't help

### describe your issue

Trying to run pre-commit after I updated to python3.10, having this issue

```
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Restored changes from /home/cawoon/.cache/pre-commit/patch1661510071-33086.
An unexpected error has occurred: CalledProcessError: command: ('python', '-mpip', 'install', '.')
return code: 1
expected return code: 0
stdout:
    Executable `python` not found
stderr: (none)
Check the log at /home/cawoon/.cache/pre-commit/pre-commit.log
```

### pre-commit --version

2.20.0

### .pre-commit-config.yaml

```yaml
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.3.0
    hooks:
      - id: check-docstring-first
      - id: check-json
      - id: check-yaml
        args: ["--allow-multiple-documents"]
      - id: debug-statements
      - id: fix-encoding-pragma
        args: ["--remove"]

  - repo: https://github.com/asottile/pyupgrade
    rev: v2.37.3
    hooks:
      - id: pyupgrade

  - repo: https://github.com/psf/black
    rev: 22.6.0
    hooks:
      - id: black
        args: ["--line-length=99"]

  - repo: https://github.com/pycqa/isort
    rev: 5.10.1
    hooks:
      - id: isort
        args: ["--gitignore", "--line-length=99", "--profile", "black"]

  - repo: https://github.com/pycqa/flake8
    rev: 5.0.4
    hooks:
      - id: flake8
        additional_dependencies: [
          '-e', 'git+https://github.com/pycqa/pyflakes#egg=pyflakes',
          '-e', 'git+https://github.com/pycqa/pycodestyle#egg=pycodestyle',
        ]
        args: ["--max-line-length=99"]
```

### version information

```
pre-commit version: 2.20.0
git --version: git version 2.34.1
sys.version:
    3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
sys.executable: /usr/bin/python3.10
os.name: posix
sys.platform: linux
```

### error information

```
[...]

Issue #2299, opened 2022-03-22
Installing precommit hooks fails due to non-existent python bin


russkel:
Looks like a call to `python` is hard coded when it shouldn't be.

```
↳ pre-commit run --all-files
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/asottile/pyupgrade.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://github.com/pycqa/flake8.
[INFO] Initializing environment for https://github.com/codespell-project/codespell.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('python', '-mpip', 'install', '.')
return code: 1
expected return code: 0
stdout:
    Executable `python` not found
stderr: (none)
Check the log at /home/russ/.cache/pre-commit/pre-commit.log
```
Running on Ubuntu jammy with python 3.10. `/usr/bin/python` does not exist.

### version information

```
pre-commit version: 2.17.0
git --version: git version 2.34.1
sys.version:
    3.10.3 (main, Mar 16 2022, 17:19:40) [GCC 11.2.0]
sys.executable: /usr/bin/python3
os.name: posix
sys.platform: linux
```

### error information

```
An unexpected error has occurred: CalledProcessError: command: ('python', '-mpip', 'install', '.')
return code: 1
expected return code: 0
stdout:
    Executable `python` not found
stderr: (none)
```

```
Traceback (most recent call last):
  File "/home/russ/.local/lib/python3.10/site-packages/pre_commit/error_handler.py", line 70, in error_handler
    yield
  File "/home/russ/.local/lib/python3.10/site-packages/pre_commit/main.py", line 396, in main
    return run(args.config, store, args)
  File "/home/russ/.local/lib/python3.10/site-packages/pre_commit/commands/run.py", line 416, in run
    install_hook_envs(to_install, store)
  File "/home/russ/.local/lib/python3.10/site-packages/pre_commit/repository.py", line 224, in install_hook_envs
    _hook_install(hook)
  File "/home/russ/.local/lib/python3.10/site-packages/pre_commit/repository.py", line 82, in _hook_install
    lang.install_environment(
[...]

PR #2526, opened 2022-09-25
Fix for creating a new python virtualenv inside a virtualenv/venv

marcelotduarte:
This fixes a bug similar to that reported in issues #2525 #2503 and #2299.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -211,6 +211,9 @@ def install_environment(
     envdir = prefix.path(helpers.environment_dir(ENVIRONMENT_DIR, version))
     venv_cmd = [sys.executable, '-mvirtualenv', envdir]
     python = norm_version(version)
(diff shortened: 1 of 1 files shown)


asottile:
no this isn't right or the right place to fix a bug in debian

marcelotduarte:
I meant similar because I'm using ubuntu 22.04, python 3.10.6
The fixes reported in the issues do not solve my issue because it is different.
And I don't understand why you can't get around a problem if you have a solution. This prevents us from continuing to use a great tool you've developed. I work on an open-source project and I really like it when you have a solution that makes life easier for others.

asottile:
I promise you the solutions for 22.04 and 3.10.4 also work for 3.10.6 because I am using them

asottile:
and I'm case you're having a hard time with the threads because there's admittedly a lot of them because it's a common problem: install python3-distutils and then clear the virtualenv cache in ~/.local/share

also your patch is not a good one and breaks windows

PR #2531, opened 2022-09-28
Add error handling for relpath ValueError

SitiSchu:


Syke04 (approved):

asottile:
#2571 is slightly closer to fixing this -- but needs an additional assertion cc @m-rsha

PR #2534, opened 2022-09-30
Rust as 1st class language

Holzhaus:
This adds support for pinning a specific rust version and bootstrapping rust with rustup/rustup-init.

I tried to follow the desired behavior described here: https://github.com/pre-commit/pre-commit/pull/1863#issuecomment-817185697

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -1,13 +1,20 @@
 from __future__ import annotations
 
 import contextlib
+import functools
 import os.path
+import platform
+import shutil
+import sys
+import tempfile
+import urllib.request
 from typing import Generator
 from typing import Sequence
 
 import toml
 
 import pre_commit.constants as C
+from pre_commit import parse_shebang
 from pre_commit.envcontext import envcontext
 from pre_commit.envcontext import PatchesT
 from pre_commit.envcontext import Var
@@ -16,24 +23,61 @@ from pre_commit.languages import helpers
 from pre_commit.prefix import Prefix
 from pre_commit.util import clean_path_on_failure
 from pre_commit.util import cmd_output_b
+from pre_commit.util import make_executable
+from pre_commit.util import win_exe
 
 ENVIRONMENT_DIR = 'rustenv'
-get_default_version = helpers.basic_get_default_version
 health_check = helpers.basic_health_check
 
 
-def get_env_patch(target_dir: str) -> PatchesT:
+@functools.lru_cache(maxsize=1)
+def get_default_version() -> str:
+    # If rust is already installed, we can save a bunch of setup time by
+    # using the installed version.
+    #
+    # Just detecting the executable does not suffice, because if rustup is
+    # installed but no toolchain is available, then `cargo` exists but
+    # cannot be used without installing a toolchain first.
+    if cmd_output_b('cargo', '--version', retcode=None)[0] == 0:
+        return 'system'
+    else:
+        return C.DEFAULT
+
+
+def _rust_toolchain(language_version: str) -> str:
+    """Transform the language version into a rust toolchain version."""
+    if language_version == C.DEFAULT:
+        return 'stable'
+    else:
+        return language_version
+
+
+def _envdir(prefix: Prefix, version: str) -> str:
+    directory = helpers.environment_dir(ENVIRONMENT_DIR, version)
+    return prefix.path(directory)
+
+
+def get_env_patch(target_dir: str, version: str) -> PatchesT:
     return (
+        ('CARGO_HOME', target_dir),
         ('PATH', (os.path.join(target_dir, 'bin'), os.pathsep, Var('PATH'))),
+        # Only set RUSTUP_TOOLCHAIN if we don't want use the system's default
+        # toolchain
+        *(
+            (('RUSTUP_TOOLCHAIN', _rust_toolchain(version)),)
+            if version != 'system' else ()
+        ),
     )
 
 
 @contextlib.contextmanager
-def in_env(prefix: Prefix) -> Generator[None, None, None]:
-    target_dir = prefix.path(
-        helpers.environment_dir(ENVIRONMENT_DIR, C.DEFAULT),
-    )
-    with envcontext(get_env_patch(target_dir)):
+def in_env(
+    prefix: Prefix,
+    language_version: str,
+) -> Generator[None, None, None]:
+    with envcontext(
+        get_env_patch(_envdir(prefix, language_version), language_version),
+    ):
         yield
 
 
@@ -52,15 +96,45 @@ def _add_dependencies(
         f.truncate()
 
 
+def install_rust_with_toolchain(toolchain: str) -> None:
+    with tempfile.TemporaryDirectory() as rustup_dir:
+        with envcontext((('RUSTUP_HOME', rustup_dir),)):
+            # acquire `rustup` if not present
+            if parse_shebang.find_executable('rustup') is None:
+                # We did not detect rustup and need to download it first.
+                if sys.platform == 'win32':  # pragma: win32 cover
+                    if platform.machine() == 'x86_64':
+                        url = 'https://win.rustup.rs/x86_64'
+                    else:
+                        url = 'https://win.rustup.rs/i686'
+                else:  # pragma: win32 no cover
+                    url = 'https://sh.rustup.rs'
+
+                resp = urllib.request.urlopen(url)
+
+                rustup_init = os.path.join(rustup_dir, win_exe('rustup-init'))
+                with open(rustup_init, 'wb') as f:
+                    shutil.copyfileobj(resp, f)
+                make_executable(rustup_init)
+
+                # install rustup into `$CARGO_HOME/bin`
+                cmd_output_b(
+                    rustup_init, '-y', '--quiet', '--no-modify-path',
+                    '--default-toolchain', 'none',
+                )
+
+            cmd_output_b(
+                'rustup', 'toolchain', 'install', '--no-self-update',
+                toolchain,
+            )
+
+
 def install_environment(
         prefix: Prefix,
         version: str,
         additional_dependencies: Sequence[str],
 ) -> None:
-    helpers.assert_version_default('rust', version)
-    directory = prefix.path(
-        helpers.environment_dir(ENVIRONMENT_DIR, C.DEFAULT),
-    )
+    directory = _envdir(prefix, version)
 
     # There are two cases where we might want to specify more dependencies:
     # as dependencies for the library being built, and as binary packages
@@ -84,17 +158,21 @@ def install_environment(
         packages_to_install: set[tuple[str, ...]] = {('--path', '.')}
         for cli_dep in cli_deps:
             cli_dep = cli_dep[len('cli:'):]
-            package, _, version = cli_dep.partition(':')
-            if version != '':
-                packages_to_install.add((package, '--version', version))
+            package, _, crate_version = cli_dep.partition(':')
+            if crate_version != '':
+                packages_to_install.add((package, '--version', crate_version))
             else:
                 packages_to_install.add((package,))
 
-        for args in packages_to_install:
-            cmd_output_b(
-                'cargo', 'install', '--bins', '--root', directory, *args,
-                cwd=prefix.prefix_dir,
-            )
+        with in_env(prefix, version):
+            if version != 'system':
+                install_rust_with_toolchain(_rust_toolchain(version))
+
+            for args in packages_to_install:
+                cmd_output_b(
+                    'cargo', 'install', '--bins', '--root', directory, *args,
+                    cwd=prefix.prefix_dir,
+                )
 
 
 def run_hook(
@@ -102,5 +180,5 @@ def run_hook(
         file_args: Sequence[str],
         color: bool,
 ) -> tuple[int, bytes]:
-    with in_env(hook.prefix):
+    with in_env(hook.prefix, hook.language_version):
         return helpers.run_xargs(hook, hook.cmd, file_args, color=color)
--- a/tests/languages/rust_test.py
+++ b/tests/languages/rust_test.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+from unittest import mock
+
+import pytest
+
+import pre_commit.constants as C
+from pre_commit import parse_shebang
+from pre_commit.languages import rust
+from pre_commit.prefix import Prefix
+from pre_commit.util import cmd_output
+
+ACTUAL_GET_DEFAULT_VERSION = rust.get_default_version.__wrapped__
+
+
+@pytest.fixture
+def cmd_output_b_mck():
+    with mock.patch.object(rust, 'cmd_output_b') as mck:
+        yield mck
+
+
+def test_sets_system_when_rust_is_available(cmd_output_b_mck):
+    cmd_output_b_mck.return_value = (0, b'', b'')
+    assert ACTUAL_GET_DEFAULT_VERSION() == 'system'
+
+
+def test_uses_default_when_rust_is_not_available(cmd_output_b_mck):
+    cmd_output_b_mck.return_value = (127, b'', b'error: not found')
+    assert ACTUAL_GET_DEFAULT_VERSION() == C.DEFAULT
+
+
+@pytest.mark.parametrize('language_version', (C.DEFAULT, '1.56.0'))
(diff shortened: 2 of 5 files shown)


Holzhaus:
@asottile Unsure how to remove cargo/rust from the CI env to test the bootstrapping code. Maybe these tests can be added in a separate PR?

asottile:
> @asottile Unsure how to remove cargo/rust from the CI env to test the bootstrapping code. Maybe these tests can be added in a separate PR?

I think you can probably take the same approach that the other 1st-party languages use? where they stub out the "get default" to return `None` and observe the results

pre_commit/languages/rust.py
```diff
@@ -52,15 +93,57 @@ def _add_dependencies(
...
+        )
+        with envcontext(rustup_env):
+            if rustup_exe is None:
+                # We did not detect rustup and need to download it first.
+                if sys.platform == 'win32':
+                    if platform.machine() == 'x86_64':
+                        url = 'https://win.rustup.rs/x86_64'
+                    else:
+                        url = 'https://win.rustup.rs/i686'
+                else:
```
asottile, line 111: I think windows can also run on arm64 -- might need to make a mapping of architectures and platforms to select the url out of that

pre_commit/languages/rust.py
```diff
@@ -90,17 +173,23 @@ def install_environment(
...
 
-        for args in packages_to_install:
-            cmd_output_b(
-                'cargo', 'install', '--bins', '--root', directory, *args,
-                cwd=prefix.prefix_dir,
-            )
+        with in_env(prefix, language_version):
+            toolchain = os.getenv('RUSTUP_TOOLCHAIN')
+            if toolchain is not None:
+                install_rust_with_toolchain(toolchain)
```
asottile, line 179: rather than ping-ponging through environment variables as globals this should just check directly -- I believe that'd be `language_version != 'system'`

pre_commit/languages/rust.py
```diff
@@ -52,15 +93,57 @@ def _add_dependencies(
...
+            _, _, _ = cmd_output_b(
+                rustup_exe, 'toolchain', 'install', '--no-self-update',
+                toolchain, retcode=0,
+            )
+
+
 def install_environment(
         prefix: Prefix,
-        version: str,
+        language_version: str,
```
asottile, line 143: ah -- yeah I think the better fix is to rename the `version` below to not clobber this (since all the languages use this argument name, best to keep the consistency here)

pre_commit/languages/rust.py
```diff
@@ -52,15 +93,57 @@ def _add_dependencies(
...
 
 
+def install_rust_with_toolchain(toolchain: str) -> None:
+    # Check if rustup is already present in PATH
+    rustup_exe = parse_shebang.find_executable('rustup')
+    with tempfile.TemporaryDirectory() as rustup_dir:
+        rustup_env = (
+            ('RUSTUP_HOME', rustup_dir),
+        )
+        with envcontext(rustup_env):
```
asottile, line 103: I *think* you can do this using `ExitStack` since creating this directory and such is conditional? (that way it doesn't need to write a directory and change the environment for noops)

pre_commit/languages/rust.py
```diff
@@ -16,27 +22,62 @@
...
         yield
 
 
+def health_check(prefix: Prefix, language_version: str) -> str | None:
+    with in_env(prefix, language_version):
+        retcode, _, _ = cmd_output_b('cargo', '--version', retcode=None)
+        if retcode != 0:
+            return f'`cargo --version` returned {retcode}'
+        else:
+            return None
```
asottile, line 78: `health_check` might not be needed, though curious to hear your thoughts (most of the `health_check` implementations are for situations where OS upgrades pull out the necessary parts of built environments) -- think like `libssl1.1` -> `libssl3.0` and then the `.so` files are missing links.  my understanding of rust is that it's self-contained ?

pre_commit/languages/rust.py
```diff
@@ -16,27 +22,62 @@
...
+TOOLCHAIN_FROM_VERSION = {
+    'system': None,
+    C.DEFAULT: 'stable',
+}
 
 
-def get_env_patch(target_dir: str) -> PatchesT:
+@functools.lru_cache(maxsize=1)
+def get_default_version() -> str:
+    # if ust is already installed, we can save a bunch of setup time by
```
asottile, line 37: rust

pre_commit/languages/rust.py
Holzhaus: I'm not a windows expert, but IIUC Windows on ARM has x86 emulation built in: https://learn.microsoft.com/en-us/windows/arm/overview#build-windows-apps-that-run-on-arm

pre_commit/languages/rust.py
Holzhaus: https://github.com/pre-commit/pre-commit/pull/2534/commits/c81a7e2830240cae6627e46dc203f1e686b31582

pre_commit/languages/rust.py
Holzhaus: https://github.com/pre-commit/pre-commit/pull/2534/commits/fc081b498f883a38b223e12f78d2c61d3b267cc5

pre_commit/languages/rust.py
Holzhaus: https://github.com/pre-commit/pre-commit/pull/2534/commits/901e7541dc2383ee075babb8d6d069fb8f73e477

pre_commit/languages/rust.py
Holzhaus: We still need the toolchain type, but you're right that we don't need to get it from the env. See https://github.com/pre-commit/pre-commit/pull/2534/commits/7c22138df19335864f08738e2d860686c61194fa.

asottile:
anyway -- wanted to say I'm pretty excited for this PR :D

pre_commit/languages/rust.py
```diff
@@ -84,23 +164,29 @@ def install_environment(
...
             else:
                 packages_to_install.add((package,))
 
-        for args in packages_to_install:
-            cmd_output_b(
-                'cargo', 'install', '--bins', '--root', directory, *args,
-                cwd=prefix.prefix_dir,
-            )
+        with in_env(prefix, version):
+            toolchain = TOOLCHAIN_FROM_VERSION.get(version, version)
```
asottile, line 174: I don't think `TOOLCHAIN_FROM_VERSION` is all that useful -- this condition here is just `if version != 'system':` rather than needing to go through this mapping

pre_commit/languages/rust.py
```diff
@@ -52,15 +90,57 @@ def _add_dependencies(
...
+    # Check if rustup is already present in PATH
+    rustup_exe = parse_shebang.find_executable('rustup')
+    with tempfile.TemporaryDirectory() as rustup_dir:
+        rustup_env = (
+            ('RUSTUP_HOME', rustup_dir),
+        )
+        with envcontext(rustup_env):
+            if rustup_exe is None:
+                # We did not detect rustup and need to download it first.
+                if sys.platform == 'win32':  # pragma: no cover
```
asottile, line 103: I think this'd be `# pragma: win32 cover` and the else would be `# pragma: win32 no cover`

pre_commit/languages/rust.py
```diff
@@ -16,24 +22,56 @@
...
+
+
+def get_env_patch(target_dir: str, version: str) -> PatchesT:
+    toolchain = TOOLCHAIN_FROM_VERSION.get(version, version)
     return (
+        ('CARGO_HOME', target_dir),
         ('PATH', (os.path.join(target_dir, 'bin'), os.pathsep, Var('PATH'))),
+        # Only set RUSTUP_TOOLCHAIN if we don't want use the system's default
+        # toolchain
+        *((('RUSTUP_TOOLCHAIN', toolchain),) if toolchain else ()),
```
asottile, line 63: this then becomes `if version != 'system' else ()`

pre_commit/languages/rust.py
```diff
@@ -52,15 +90,57 @@ def _add_dependencies(
...
+                # Now, rustup should be present in `$CARGO_HOME/bin`.
+                rustup_exe = parse_shebang.find_executable('rustup')
+
+            if rustup_exe is None:
+                raise AssertionError(
+                    'failed to find rustup even though it should have been '
+                    'installed at this point -- this is probably a bug',
+                )
+
+            _, _, _ = cmd_output_b(
```
asottile, line 132: you don't need to assign the return value here

tests/languages/rust_test.py
```diff
@@ -0,0 +1,108 @@
...
+    shutil.rmtree(tmpdir.join('target'))
+
+    with rust.in_env(prefix, '1.64.0'):
+        assert cmd_output('hello_world')[1] == 'Hello, world!\n'
+
+
+@pytest.mark.skipif(
+    cmd_output('cargo', '--version', retcode=None)[0] != 0,
+    reason='requires working system installation of rust/cargo',
+)
```
asottile, line 84: since there's no environment CI or I use without this I'd remove this (and the CONTRIBUTING mentions needing a system rust for some of the tests)

tests/languages/rust_test.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    tmpdir.join('Cargo.toml').ensure().write(
+        toml.dumps({
+            'package': {
+                'name': 'hello_world',
+                'version': '0.1.0',
+                'edition': '2021',
+            },
+            'dependencies': {},
+        }),
```
asottile, line 99: can probably skip `toml` and just write a string here

pre_commit/languages/rust.py
Holzhaus: ok, this is a bit embarassing :hand_over_mouth:

tests/languages/rust_test.py
Holzhaus: I adapted this from the node tests which use `json.dumps()` for the `package.json` instead of writing a string directly. And we have a dependency on `toml` for reading the `Cargo.toml` file anyway. But I can change it if you want.

Holzhaus:
The windows tests are failing because Windows has problems with long paths:

```
E           pre_commit.util.CalledProcessError: command: ('C:\\Users\\VssAdministrator\\.cargo\\bin\\cargo.EXE', 'install', '--bins', '--root', 'C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_local_rust_additional_dep0\\0\\.pre-commit\\repo04ryo5oe\\rustenv-system', 'hello-cli', '--version', '0.2.2')
E           return code: 101
E           expected return code: 0
E           stdout: (none)
E           stderr:
E               error: failed to initialize index git repository (in "C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_local_rust_additional_dep0\\0\\.pre-commit\\repo04ryo5oe\\rustenv-system\\registry\\index\\github.com-1ecc6299db9ec823")
E               
E               Caused by:
E                 path too long: 'C:/Users/VssAdministrator/AppData/Local/Temp/pytest-of-unknown/pytest-0/test_local_rust_additional_dep0/0/.pre-commit/repo04ryo5oe/rustenv-system/registry/index/github.com-1ecc6299db9ec823/.git/'; class=Filesystem (30)
```

Wouldn't have expected that a path of this length is an issue in 2022, but here we are now. Upstream issue seems to be https://github.com/rust-lang/cargo/issues/9770. Ideas?

asottile:
> The windows tests are failing because Windows has problems with long paths:
> 
> ```
> E           pre_commit.util.CalledProcessError: command: ('C:\\Users\\VssAdministrator\\.cargo\\bin\\cargo.EXE', 'install', '--bins', '--root', 'C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_local_rust_additional_dep0\\0\\.pre-commit\\repo04ryo5oe\\rustenv-system', 'hello-cli', '--version', '0.2.2')
> E           return code: 101
> E           expected return code: 0
> E           stdout: (none)
> E           stderr:
> E               error: failed to initialize index git repository (in "C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_local_rust_additional_dep0\\0\\.pre-commit\\repo04ryo5oe\\rustenv-system\\registry\\index\\github.com-1ecc6299db9ec823")
> E               
> E               Caused by:
> E                 path too long: 'C:/Users/VssAdministrator/AppData/Local/Temp/pytest-of-unknown/pytest-0/test_local_rust_additional_dep0/0/.pre-commit/repo04ryo5oe/rustenv-system/registry/index/github.com-1ecc6299db9ec823/.git/'; class=Filesystem (30)
> ```
> 
> Wouldn't have expected that a path of this length is an issue in 2022, but here we are now. Upstream issue seems to be [rust-lang/cargo#9770](https://github.com/rust-lang/cargo/issues/9770). Ideas?

might be able to use a shorter temporary directory -- we had `TEMP: 'C:\temp'` for a while for the windows tests -- not sure if that's still there or not or if we should bring it back

pre_commit/languages/rust.py
Holzhaus: I only added the mapping to have a single source of truth for mapping `C.DEFAULT` to `stable`. I think I incorporated the changes in https://github.com/pre-commit/pre-commit/pull/2534/commits/8d545adedbcf33c731a995f7b2fe22fa1be2bef4

pre_commit/languages/rust.py
Holzhaus: thanks. https://github.com/pre-commit/pre-commit/pull/2534/commits/59abe3875eb893fed407a410125c8910bcc3d2cf

pre_commit/languages/rust.py
Holzhaus: https://github.com/pre-commit/pre-commit/commit/[8d545ad](https://github.com/pre-commit/pre-commit/commit/8d545adedbcf33c731a995f7b2fe22fa1be2bef4)edbcf33c731a995f7b2fe22fa1be2bef4

pre_commit/languages/rust.py
Holzhaus: https://github.com/pre-commit/pre-commit/pull/2534/commits/5250fbbf932e8312d7bb7b291221fe4e1f7fa648

tests/languages/rust_test.py
Holzhaus: https://github.com/pre-commit/pre-commit/pull/2534/commits/da5ca6ff03185176911040af24bec5c2629f45d5

tests/languages/rust_test.py
Holzhaus: https://github.com/pre-commit/pre-commit/pull/2534/commits/da5ca6ff03185176911040af24bec5c2629f45d5

Holzhaus:
I tried to shorten the path passing tox' envtmpdir as pytest `--basetemp`. This fixes the path length issue. Unfortunately, 3 other tests unrelated to to rust fail now :(

tox.ini
```diff
@@ -1,13 +1,22 @@
...
-    coverage report
+    unix: coverage erase
+          coverage run -m pytest --basetemp="/tmp/pytest-{envname}" {posargs:tests}
+          coverage report
+    windows: coverage erase
+             coverage run -m pytest --basetemp="C:\tmp\pytest-{envname}" {posargs:tests}
+             coverage report
 
 [testenv:pre-commit]
 skip_install = true
```
asottile, line 22: I'd rather not change this -- it's intentionally simple.  I meant to just add the temp override in `azure-pipelines.yml` since that's really the only place the paths are going to be so deep

Holzhaus: @asottile Thanks for taking care of this!

Holzhaus:
Looks like CI passed except for the failing `test_pre_merge_commit_integration` on Windows, but that one also fails on current `main` so it's not related to this PR. Anything else to do?

asottile (approved):

Holzhaus:
Any ETA when this will land in a stable release?

I have a rust pre-commit hook that I want to use on a non-rust project and don't want to bother contributors with setting up rust themselves, so my plan is to pin a rust toolchain and set `minimum_pre_commit_version` to 2.21.0 to prevent people from running into issues.

Holzhaus:
> it's sorta rude to demand releases like that -- it'll happen when it happens

I'm sorry, that was not my intention. I was just interested because I don't know what the release roadmap for pre-commit is. If there is something blocking the release, I can try to help out.

PR #2561, opened 2022-10-27
Support Golang major versions

arjen-ag5:
Golang has a special treatment of major versions. If you go from v1.x.x to v2.x.x you are supposed to put all your code in a v2 directory in the root of the repository. This break pre-commit because pre-commit tries to do a `go install ./...` in the root, this fails because now the `go.mod` file is missing. See this repository for example: https://github.com/nicksnyder/go-i18n.

This patch checks if there is a go.mod in the root of the project, if it is not it will scan all subdirectories matching the v[0-9]+ pattern until it finds one containing a go.mod file. This will be the _cwd_ for the install command.

For reference:
https://go.dev/doc/modules/major-version

### DIFF
--- a/pre_commit/languages/golang.py
+++ b/pre_commit/languages/golang.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import os.path
+import re
 import sys
 from typing import Generator
 from typing import Sequence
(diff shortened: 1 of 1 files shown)


asottile:
the docs you linked are for a subpath install. pre-commit only supports the convention of the project being at the root.  guessing is a surprising behavior that I can't support

PR #2564, opened 2022-10-27
allow GIT_ASKPASS as a pass through env var 

mattp-:
documented via man gitcredentials,  it is used to provide a script/input for git to fetch creds in a no-tty usecase. used among other things by jenkins to pass credentials down to git for authentication.

https://github.com/jenkinsci/git-plugin/blob/1e3488a730a169778ba0863dd4edbb1dc29154a1/README.adoc#git-bindings https://github.com/jenkinsci/git-plugin/blob/9429e7d05df3dbb4060ac6ab4da6538bb0eb50ba/src/main/java/jenkins/plugins/git/GitUsernamePasswordBinding.java#L130

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -43,6 +43,7 @@ def no_git_env(_env: Mapping[str, str] | None = None) -> dict[str, str]:
             'GIT_SSL_NO_VERIFY', 'GIT_CONFIG_COUNT',
             'GIT_HTTP_PROXY_AUTHMETHOD',
             'GIT_ALLOW_PROTOCOL',
+            'GIT_ASKPASS',
         }
     }
 


pre_commit/git.py
```diff
@@ -43,6 +43,7 @@ def no_git_env(_env: Mapping[str, str] | None = None) -> dict[str, str]:
             'GIT_SSL_NO_VERIFY', 'GIT_CONFIG_COUNT',
             'GIT_HTTP_PROXY_AUTHMETHOD',
             'GIT_ALLOW_PROTOCOL',
+            'SSH_ASKPASS', 'GIT_ASKPASS', 'GIT_USERNAME', 'GIT_PASSWORD',
```
asottile, line 46: - `SSH_ASKPASS` doesn't start with `GIT_` so it's not relevant
- `GIT_USERNAME` isn't a variable supported by `git`
- `GIT_PASSWORD` isn't a variable supported by `git`

mattp-: oops, good point; removed ssh_askpass. although git_username/git_password isn't directly, it is the default used by jenkins git plugin; so without it theres no way to provide creds down to git from jenkins as a caller without it. is there any harm in adding it ? I don't think you can use jenkins credential store to execute precommit otherwise.

asottile: jenkins shouldn't be setting environment variables in other tools' namespaces -- I'm fine adding `GIT_ASKPASS` but not the other ones

mattp-: fair enough; removed in latest force push. commit is just GIT_ASKPASS now

asottile (approved):

mattp-:
in case anyone ever stumbles onto this from google, i was mistaken on GIT_USERNAME/GIT_PASSWORD, those aren't actually passed down by that plugin as env vars at all; just me failing to read java code properly. thanks for the merge 👍

PR #2571, opened 2022-10-30
Add `relpath` test for Windows error coverage

m-rsha:
Fixes #2531 , beep boop.


asottile:
@m-rsha I think this needs an assertion showing that it preserves the value -- and maybe some more realistic inputs / outputs to demonstrate it's trying to handle windows paths on different drives?

PR #2581, opened 2022-11-03
Add pip dependancy support in conda language

AlexisFetet:
Allow users to add pip dependancies in their conda hooks by adding "pip install" in the dependancy : 

conda install :
additionnal dependencies:
   - numpy

pip install :
additionnal dependencies:
   - pip install numpy


asottile:
please in the future open an issue describing your actual problem before wasting time on implementation -- the **why** is much more important than the **what**

I don't want to support this mixed-mode case -- this is also discouraged in general by conda as well since mixing the two managers can lead to problems with resolution and shared libraries

PR #2591, opened 2022-11-11
Attempt to maximize Rust coverage

m-rsha:
And by coverage, I mean... adding pragmas everywhere?


m-rsha:
Is this what you had wanted? I figured I'd ask now instead of waiting for tox to maybe fail again ^^

PR #2593, opened 2022-11-12
Add hook tags

auhsoja:
It seemed rather useful and easy-to-implement to have a way of tagging your hooks so that you can more closely control which subset is run/skipped. For example, you may want to designate some subset to only run during CI (like tests that might take a long time) or to only run locally (like compiling your dependencies).

The reason I started on this was mainly because I wanted local typechecking and linting to be fast, but to validate everything in CI, the tools had to be containerized, which digs into your flow a little bit.

If I were smarter and more experienced, I probably would've come up with a better solution to this, but again, this seemed pretty simple and potentially useful for other reasons.

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -80,6 +80,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
     cfgv.Optional('minimum_pre_commit_version', cfgv.check_string, '0'),
     cfgv.Optional('require_serial', cfgv.check_bool, False),
     cfgv.Optional('stages', cfgv.check_array(cfgv.check_one_of(C.STAGES)), []),
+    cfgv.Optional('tags', cfgv.check_array(cfgv.check_string), []),
     cfgv.Optional('verbose', cfgv.check_bool, False),
 )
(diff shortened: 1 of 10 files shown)


asottile:
please don't waste time on implementation before discussing features

this has been discussed before and rejected, please search the issue tracker next time

asottile:
note also this is what `manual` is for

auhsoja:
1. Fair enough. Again. The rationale was that it was so easy to implement and I was hoping that already having the code would make people get less mad at me if they saw the idea as unnecessary work (I suppose that was a mistake).
2. Ironically enough, doing this whole thing made me think that perhaps I could do it with stages if I thought about it some more, so again fair enough.
3. I did check the issue tracker, but the previous guy asked 3 years ago and used different terminology so I missed it. Sorry. I feel like the point still stands, though, that if you use the one-at-a-time solution posed, you'll add unnecessary bloat from scripting away the groups, and then you'll have more files that need to be maintained in concert.
4. Just a random thing I saw. Is it the case that the skip tests have a bug? Looks like they pass no matter what by virtue of having no files to check.

PR #2604, opened 2022-11-18
update tests to change flake8 repo from gitlab to github

tylerpotts:
Tests were still pointing to old gitlab repo for flake8, needed updating to github


asottile:
the strings here don't matter, they're never actually cloned

PR #2613, opened 2022-11-25
Add sys.executable escapes on Windows

vidartf:
This fixes a lack of escaping backslashes when installing hooks on Windows. This is pretty critical, since backslashes are the main path separator, so will lead to catastrophic errors like:

```
git/hooks/pre-commit: line 6: exec: C:mypythoninstalldirpython.exe: not found
```

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -102,7 +102,11 @@ def _install_hook_script(
             hook_file.write('#!/bin/sh\n')
 
         hook_file.write(before + TEMPLATE_START)
-        hook_file.write(f'INSTALL_PYTHON={shlex.quote(sys.executable)}\n')
+        sys_exec = sys.executable
+        if sys.platform == 'win32':  # pragma: win32 cover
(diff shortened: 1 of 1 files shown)


asottile:
it works fine for me on windows (as demonstrated by the test suite as well), please start with a bug report demonstrating what's actually happening

vidartf:
Yeah, I see it now. I had somehow ended up with a hook file like this (no idea really where or when I got it):
```
#!/usr/bin/env bash
INSTALL_PYTHON=C:\mypython\installdir\python.exe
ARGS=(hook-impl --config=.pre-commit-config.yaml --hook-type=pre-commit)
HERE="$(cd "$(dirname "$0")" && pwd)"
ARGS+=(--hook-dir "$HERE" -- "$@")
exec "$INSTALL_PYTHON" -m pre_commit "${ARGS[@]}"
```
It seems that when there are quotes around it, the lack of backslash escapes is not an issue. I blindly assumed that quotes wouldn't have an effect, and couldn't understand how this would ever work for anyone on Window :)  Sorry for the noise!

asottile:
yeah that file contents is not possible from our code so it must be something on your end

PR #2616, opened 2022-11-27
Update revision of pre-commit-hooks in sample-config

waweber:
Trivial change to bring the rev specified in the `sample-config` to the current version


asottile:
this isn't meant to be kept up to date

waweber:
@asottile according to your comment, it is meant to be periodically updated (and it even suggests fetching the latest version via git somehow?). This could probably either be updated to reflect that its purely an example, or just removed

asottile:
yeah the comment is from 2017 I'll just remove the comment

PR #2623, opened 2022-12-02
add in an option to disable a hook when run with --all-files

mdering:
This adds in a simple option for a hook to disable it when pre-commit is run with the `--all-files` flag. The option is True by default. This should help with heavier weight hooks or making incremental improvements to a large repo. This way, if a user runs `pre-commit run --all-files` it will not run this hook, but it will run normally during the pre-commit stage


asottile:
please don't waste time on features before discussing them in an issue

use SKIP, it already works for this

mdering:
skip does not work for this. I will create a ticket.

asottile:
it absolutely does, please do not

PR #2631, opened 2022-12-06
dotnet: ignore nuget source during tool install

rkm:
Removes `nuget.org` from the default lookup configuration so package resolution is performed against the local `build_dir` only.

This fixes an issue where dotnet attempts to install unwanted packages or older versions, as `nuget.org` has higher precedence than the specified sources.


rkm:
Seems to be 2 unrelated CI failures here:
- Swift install failing
- Coverage failing on rust hooks

PR #2634, opened 2022-12-06
fix CI

asottile:


pre_commit/languages/rust.py
```diff
@@ -99,10 +98,7 @@ def install_rust_with_toolchain(toolchain: str) -> None:
             if parse_shebang.find_executable('rustup') is None:
                 # We did not detect rustup and need to download it first.
                 if sys.platform == 'win32':  # pragma: win32 cover
-                    if platform.machine() == 'x86_64':
-                        url = 'https://win.rustup.rs/x86_64'
-                    else:
-                        url = 'https://win.rustup.rs/i686'
+                    url = 'https://win.rustup.rs/x86_64'
```
Holzhaus, line 101: Won't this break support for 32-bit windows? Maybe we should use something like:

```suggestion
                    if platform.machine() in {'x86_64', 'AMD64', 'x64'}:
                        url = 'https://win.rustup.rs/x86_64'
                    else:
                        url = 'https://win.rustup.rs/i686'
```
(`AMD64` and `x64` are just examples, I have no idea what values are possible...)

asottile: yes but the year is 2022

PR #2637, opened 2022-12-09
tidied and sorted imports

marksmayo:
removed ones not required as well, groups and spaced.

experimental, as I'm having trouble with tox locally.  used isort.


asottile:
please don't do this to anyone

Issue #2513, opened 2022-09-12
Autoupdate attempts to update to pre-release tags


christianbundy:
### search tried in the issue tracker

autoupdate alpha beta pre-release

## Problem

When I run `pre-commit autoupdate`, I am surprised to see what my dependencies are updated to pre-release versions (alpha, beta, release candidate, etc).

Concrete example: https://github.com/pre-commit/mirrors-prettier/releases/tag/v3.0.0-alpha.0

I understand that 'latest tag in a repo' is a simple heuristic that usually works well, but in this case it's back-firing.

First of all: Is this a problem that pre-commit should aim to solve?

## Solution

If it _is_ a problem we'd like for pre-commit to solve, one option would be to add 'autoupdate version' regex or similar, which would allow customization about which tags to update to. This would also provide the ability to add arbitrary version constraints (e.g. patch/minor versions only) without requiring that every dependency uses the same versioning scheme (i.e. semver).

### pre-commit --version

pre-commit 2.20.0

### .pre-commit-config.yaml

```yaml
- repo: https://github.com/pre-commit/mirrors-prettier
    rev: 'v2.7.1 '
    hooks:
    -   id: prettier
```

asottile:
there's a bunch of dupes -- https://github.com/pre-commit/pre-commit/issues?q=is%3Aissue+autoupdate+pre-release+is%3Aclosed

PR #2638, opened 2022-12-09
Add `pre-commit autoupdate --semver`

jepler:
This new, optional mode chooses the best version number according to semantic versioning, rather than according to `git describe`. Prereleases are only chosen if `--semver-prerelease` is also specified.

This behavior has been requested by multiple pre-commit users in closed issues such as https://github.com/pre-commit/pre-commit/issues/2513

I'm opening this PR in part to gauge interest. I think it's likely that the final form of this feature would e.g., look in the pre-commit-config.yaml file to find out for each repo whether semantic versioning should be used or not. A commandline flag to override and upgrade to latest ref or to prerelease versions would still be useful as a way to check whether there are any problems "coming down the pike", as it were.

### DIFF
--- a/pre_commit/commands/autoupdate.py
+++ b/pre_commit/commands/autoupdate.py
@@ -6,6 +6,8 @@ from typing import Any
 from typing import NamedTuple
 from typing import Sequence
 
+from semver import VersionInfo
+
 import pre_commit.constants as C
 from pre_commit import git
 from pre_commit import output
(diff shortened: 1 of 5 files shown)


asottile:
I already said no to this feature. please read the threads you linked

jepler:
If you have any insight into the failure on pre-commit.ci and how to fix it, please let me know. I do not see that failure locally.  It looks like it may be running the version of pre-commit in this PR without installing the new requirement.

jepler:
Do you know if there are any active forks of pre-commit that accept contributions?

asottile:
no nor is my issue tracker an advertising space for them

PR #2644, opened 2022-12-16
Introduce SBT (as a 0th class) language into the pre-commit framework

stewartHutchins:
Adds a simple way (and sub-optimal) way of running SBT commands via pre-commit.

This pull request:
 - Runs a user's commands via the shell "sbt <your command>".
 - Sets up testing for future improvements.

TODO:
 - [ ] ensure that the commit titled "Add script to install SBT, via coursier." installs SBT such that tests run correctly.

Future improvements:
 - Run SBT commands via the network interface, avoiding JVM startup time (see: https://github.com/stewartHutchins/pre-commit/pull/1)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.tox
 /dist
 .vscode/
+.idea/
--- a/pre_commit/languages/sbt.py
+++ b/pre_commit/languages/sbt.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+from pre_commit.hook import Hook
+from pre_commit.languages import helpers
+
+ENVIRONMENT_DIR = None
+install_environment = helpers.no_install
+health_check = helpers.basic_health_check
+get_default_version = helpers.basic_get_default_version
+
+
+def run_hook(
+        hook: Hook,
+        file_args: Sequence[str],
+        color: bool,
+) -> tuple[int, bytes]:
+    # TODO: Improve impl to connect to run commands via SBT server
+    return run_sbt_hook_via_commandline(hook, file_args, color)
+
+
+def run_sbt_hook_via_commandline(
+        hook: Hook,
+        file_args: Sequence[str],
+        color: bool,
+) -> tuple[int, bytes]:
+    """
+    Run an SBT hook, via the commandline. The command to be run is:
+        sbt ${entry} ${args} ${files}
+    The entry and args will not be quoted (so should be wrapped in quotes as
+    appropriate by the hook author),however files will be quoted, so any
+    filenames with spaces will be interpreted as a single argument by SBT
+    """
+    entry_part = hook.entry
+    args_part = ' '.join(hook.args)
+    files_part = ' '.join(_quote(file) for file in file_args)
+    sbt_command = f'{entry_part} {args_part} {files_part}'
+    shell_cmd = ('sbt', sbt_command)
+    return helpers.run_xargs(hook, shell_cmd, [], color=color)
+
+
+def _quote(s: str) -> str:
+    return f"\"{s}\""
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import functools
 import io
 import logging
 import os.path
+from pathlib import Path
 from unittest import mock
 
 import pytest
@@ -16,6 +17,7 @@ from pre_commit.util import cmd_output
 from pre_commit.util import make_executable
 from testing.fixtures import git_dir
 from testing.fixtures import make_consuming_repo
+from testing.fixtures import make_repo
 from testing.fixtures import write_config
 from testing.util import cwd
 from testing.util import git_commit
@@ -250,3 +252,9 @@ def set_git_templatedir(tmpdir_factory):
     tdir = str(tmpdir_factory.mktemp('git_template_dir'))
     with envcontext((('GIT_TEMPLATE_DIR', tdir),)):
         yield
+
+
+@pytest.fixture
+def sbt_project_with_touch_command(tempdir_factory):
+    project_repo = make_repo(tempdir_factory, 'sbt_repo_with_touch_command')
+    return Path(project_repo)
--- a/tests/languages/sbt_test.py
+++ b/tests/languages/sbt_test.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+from itertools import product
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from pre_commit.hook import Hook
+from pre_commit.languages import sbt
+from testing.util import cwd
+from testing.util import skipif_cant_run_sbt
+
+
+@skipif_cant_run_sbt
+@pytest.mark.parametrize(
+    ['args', 'files'],
+    product(
+        [
+            [], ['argfile1.txt'], ['argfile1.txt', 'argfile2.txt'],
+            ['\"arg file1.txt\"'], ['\"arg file1.txt\"', '\"arg file2.txt\"'],
+        ],
+        [
+            [], ['filesfile1.txt'], ['filesfile1.txt', 'filesfile2.txt'],
+            ['files file1.txt'], ['files file1.txt', 'files file2.txt'],
+        ],
+    ),
+)
+def test_sbt_hook(
+        sbt_project_with_touch_command: Path,
+        args: list[str],
+        files: list[str],
+) -> None:
+    # arrange
+    project_root = sbt_project_with_touch_command
+    hook = _create_hook(
+        language='sbt',
+        entry='touch',
+        args=args,
+    )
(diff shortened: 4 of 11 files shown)


pre_commit/languages/sbt.py
```diff
@@ -0,0 +1,44 @@
...
+    entry_part = hook.entry
+    args_part = ' '.join(hook.args)
+    files_part = ' '.join(_quote(file) for file in file_args)
+    sbt_command = f'{entry_part} {args_part} {files_part}'
+    shell_cmd = ('sbt', sbt_command)
+    return helpers.run_xargs(hook, shell_cmd, [], color=color)
+
+
+def _quote(s: str) -> str:
+    return f"\"{s}\""
```
asottile, line 44: shlex.quote exists and is better

pre_commit/languages/sbt.py
```diff
@@ -0,0 +1,44 @@
...
+    The entry and args will not be quoted (so should be wrapped in quotes as
+    appropriate by the hook author),however files will be quoted, so any
+    filenames with spaces will be interpreted as a single argument by SBT
+    """
+    entry_part = hook.entry
+    args_part = ' '.join(hook.args)
+    files_part = ' '.join(_quote(file) for file in file_args)
+    sbt_command = f'{entry_part} {args_part} {files_part}'
+    shell_cmd = ('sbt', sbt_command)
+    return helpers.run_xargs(hook, shell_cmd, [], color=color)
```
asottile, line 40: this isn't going to work -- this will exhaust the command line limit (that's the point of the xargs helper)

tests/conftest.py
```diff
@@ -4,6 +4,7 @@
 import io
 import logging
 import os.path
+from pathlib import Path
```
asottile, line 7: sorry, won't be accepting anything with pathlib

tests/conftest.py
```diff
@@ -250,3 +252,9 @@ def set_git_templatedir(tmpdir_factory):
     tdir = str(tmpdir_factory.mktemp('git_template_dir'))
     with envcontext((('GIT_TEMPLATE_DIR', tdir),)):
         yield
+
+
+@pytest.fixture
+def sbt_project_with_touch_command(tempdir_factory):
+    project_repo = make_repo(tempdir_factory, 'sbt_repo_with_touch_command')
+    return Path(project_repo)
```
asottile, line 260: this isn't a useful fixture -- it saves ~1 line of code at best

tests/languages/sbt_test.py
```diff
@@ -0,0 +1,63 @@
...
+        [
+            [], ['filesfile1.txt'], ['filesfile1.txt', 'filesfile2.txt'],
+            ['files file1.txt'], ['files file1.txt', 'files file2.txt'],
+        ],
+    ),
+)
+def test_sbt_hook(
+        sbt_project_with_touch_command: Path,
+        args: list[str],
+        files: list[str],
```
asottile, line 32: [typing tests is a waste of time](https://youtu.be/lGCU0j-otls)

.gitignore
```diff
@@ -4,3 +4,4 @@
 /.tox
 /dist
 .vscode/
+.idea/
```
asottile, line 7: don't touch gitignore in projects you don't own -- you may want to set up global gitignore in your personal dotfiles

pre_commit/languages/sbt.py
```diff
@@ -0,0 +1,44 @@
...
+get_default_version = helpers.basic_get_default_version
+
+
+def run_hook(
+        hook: Hook,
+        file_args: Sequence[str],
+        color: bool,
+) -> tuple[int, bytes]:
+    # TODO: Improve impl to connect to run commands via SBT server
+    return run_sbt_hook_via_commandline(hook, file_args, color)
```
asottile, line 20: you can write this as `run = run_sbt_hook_via_commandline` but that would be silly -- there's no reason to have two functions is my point

tests/languages/sbt_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def _unquote(s: str) -> str:
+    return s.strip("\"")
+
+
+def _create_hook(**kwargs: Any) -> Hook:
+    default_values = {field: None for field in Hook._fields}
+    actual_values = {**default_values, **kwargs}
+    return Hook(**actual_values)  # type: ignore
```
asottile, line 63: there's `_hook` which does this for you

tests/languages/sbt_test.py
```diff
@@ -0,0 +1,63 @@
...
+    assert ret == 0
+    for file in args + files:
+        unquoted_file = _unquote(file)
+        expected_file = project_root.joinpath(unquoted_file).absolute()
+        assert expected_file.exists()
+        assert f'Creating file: {expected_file}' in output
+
+
+def _unquote(s: str) -> str:
+    return s.strip("\"")
```
asottile, line 57: shlex.split is the standard way to do this

tests/languages/sbt_test.py
```diff
@@ -0,0 +1,63 @@
...
+            ['files file1.txt'], ['files file1.txt', 'files file2.txt'],
+        ],
+    ),
+)
+def test_sbt_hook(
+        sbt_project_with_touch_command: Path,
+        args: list[str],
+        files: list[str],
+) -> None:
+    # arrange
```
asottile, line 34: these comments are bad, don't include them

tests/languages/sbt_test.py
```diff
@@ -0,0 +1,63 @@
...
+        entry='touch',
+        args=args,
+    )
+
+    # act
+    with cwd(project_root):
+        ret, out = sbt.run_hook(hook, files, False)
+
+    # assert
+    output = out.decode('UTF-8')
```
asottile, line 47: utf-8 is the default encoding so you should not include it

asottile (commented):
I don't see the point of this language, why doesn't `language: system` + `entry: sbt mycommand` work?

tests/languages/sbt_test.py
stewartHutchins: I personally, find them useful (even for tests), but I do understand your perspective in the video.

In this comment are you asking me to remove the type annotations from test code, or just suggesting that it is not a requirement to use type annotations in test code? (I don't mind either way)

asottile: yes please remove them unless they are necessary

pre_commit/languages/sbt.py
stewartHutchins: This makes more sense in a later PR. However I may need to change this anyway given [comment](https://github.com/pre-commit/pre-commit/pull/2644/files/163e41875472ff581c14d61ef91ce8f5bc2498f3#r1051139353)

stewartHutchins:
Sorry, missed this [comment](https://github.com/pre-commit/pre-commit/pull/2644#pullrequestreview-1221434702) - any other points are unimportant without addressing this first. I had tried to explain this in the description, however in retrospect there is probably some missing context.

You can run SBT commands in 2 ways:
1) Either run a command in the terminal, which starts SBT, loads the project's configuration, runs the command, and then shuts down SBT.
2) Or start SBT, keep it running in the background, and run commands in SBT's console as and when needed.

(Usually the 2nd option is used)

Starting SBT and loading the project's configuration takes an amount of time (e.g. it's about 14 secs on my computer to load typelevel/cats-effect's configuration). Currently SBT-based hooks are implemented (more-or-less) as you have suggested in the [comment](https://github.com/pre-commit/pre-commit/pull/2644#pullrequestreview-1221434702). This approach incurs this "loading SBT config overhead" upon *each* invocation of a hook.

Because it is not recommended to have long-running hooks (as developers will just turn the framework off) I felt it would be useful to add functionality to remove the overhead caused by re-starting/re-loading SBT for each hook.

The (eventual) aim is to use SBT's network interface, to avoid this unnecessary overhead for SBT-based pre-commit hooks. In the interest of breaking this task down, I wanted to split the feature into 3 PRs:
 - A PR (this one) which uses the command line (a poor design but allows testing to be added up-front)
 - A PR which adds network capability (this would still reload the project on each invocation, but removes JVM start-up time)
 - A PR which determines when the SBT project does not need to be reloaded, based on whether staged files will affect the SBT configuration (finally removing the overhead).

I currently have 90% of this logic implemented in a pre-commit-hook, however thought that implementing the logic within the framework it would allow any hook author to take advantage of this way of running SBT commands.

stewartHutchins:
I will be working on a better way to do implement this. I have found a much less complicated way to implement the desired functionality.

Given this code is not probably not salvageable (barring a few lines), It may be preferable to make appropriate edits to the comments, close this PR, and open a new one. I will leave this to your discretion.

asottile:
yeah sounds fine -- feel free to open a new one when you're ready!

PR #2651, opened 2022-12-23
Make Golang a first class language

### LINKED ISSUE #2649: Make Go a first class language
Hey, as many tools are now built using golang, it would be good to have it as a first class language.

I've hacked a little [POC](https://github.com/pre-commit/pre-commit/compare/main...taoufik07:pre-commit:golang_playground?expand=1) heavily inspired by the ruby lang and using https://github.com/stefanmaric/g as a go version manager.

Please let me know if that's what you had in mind before I go any further.

taoufik07:
Following #2672  
Closes #2649

### DIFF
--- a/pre_commit/languages/golang.py
+++ b/pre_commit/languages/golang.py
@@ -1,9 +1,21 @@
 from __future__ import annotations
 
 import contextlib
+import functools
+import json
 import os.path
+import platform
+import shutil
 import sys
+import tarfile
+import tempfile
+import urllib.error
+import urllib.request
+import zipfile
+from typing import ContextManager
 from typing import Generator
+from typing import IO
+from typing import Protocol
 from typing import Sequence
 
 import pre_commit.constants as C
@@ -17,20 +29,100 @@ from pre_commit.util import cmd_output
 from pre_commit.util import rmtree
 
 ENVIRONMENT_DIR = 'golangenv'
-get_default_version = helpers.basic_get_default_version
 health_check = helpers.basic_health_check
 
+_ARCH_ALIASES = {
+    'x86_64': 'amd64',
+    'i386': '386',
+    'aarch64': 'arm64',
+    'armv8': 'arm64',
+    'armv7l': 'armv6l',
+}
+_ARCH = platform.machine().lower()
+_ARCH = _ARCH_ALIASES.get(_ARCH, _ARCH)
 
-def get_env_patch(venv: str) -> PatchesT:
+
+class ExtractAll(Protocol):
+    def extractall(self, path: str) -> None: ...
+
+
+if sys.platform == 'win32':  # pragma: win32 cover
+    _EXT = 'zip'
+
+    def _open_archive(bio: IO[bytes]) -> ContextManager[ExtractAll]:
+        return zipfile.ZipFile(bio)
+else:  # pragma: win32 no cover
+    _EXT = 'tar.gz'
+
+    def _open_archive(bio: IO[bytes]) -> ContextManager[ExtractAll]:
+        return tarfile.open(fileobj=bio)
+
+
+@functools.lru_cache(maxsize=1)
+def get_default_version() -> str:
+    if helpers.exe_exists('go'):
+        return 'system'
+    else:
+        return C.DEFAULT
+
+
+def get_env_patch(venv: str, version: str) -> PatchesT:
+    if version == 'system':
+        return (
+            ('PATH', (os.path.join(venv, 'bin'), os.pathsep, Var('PATH'))),
+        )
+
     return (
-        ('PATH', (os.path.join(venv, 'bin'), os.pathsep, Var('PATH'))),
+        ('GOROOT', os.path.join(venv, '.go')),
+        (
+            'PATH', (
+                os.path.join(venv, 'bin'), os.pathsep,
+                os.path.join(venv, '.go', 'bin'), os.pathsep, Var('PATH'),
+            ),
+        ),
     )
 
 
+@functools.lru_cache
+def _infer_go_version(version: str) -> str:
+    if version != C.DEFAULT:
+        return version
+    resp = urllib.request.urlopen('https://go.dev/dl/?mode=json')
+    # TODO: 3.9+ .removeprefix('go')
+    return json.load(resp)[0]['version'][2:]
+
+
+def _get_url(version: str) -> str:
+    os_name = platform.system().lower()
+    version = _infer_go_version(version)
+    return f'https://dl.google.com/go/go{version}.{os_name}-{_ARCH}.{_EXT}'
+
+
+def _install_go(version: str, dest: str) -> None:
+    try:
+        resp = urllib.request.urlopen(_get_url(version))
+    except urllib.error.HTTPError as e:  # pragma: no cover
+        if e.code == 404:
+            raise ValueError(
+                f'Could not find a version matching your system requirements '
+                f'(os={platform.system().lower()}; arch={_ARCH})',
+            ) from e
+        else:
+            raise
+    else:
+        with tempfile.TemporaryFile() as f:
+            shutil.copyfileobj(resp, f)
+            f.seek(0)
+
+            with _open_archive(f) as archive:
+                archive.extractall(dest)
+        shutil.move(os.path.join(dest, 'go'), os.path.join(dest, '.go'))
+
+
 @contextlib.contextmanager
-def in_env(prefix: Prefix) -> Generator[None, None, None]:
-    envdir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, C.DEFAULT)
-    with envcontext(get_env_patch(envdir)):
+def in_env(prefix: Prefix, version: str) -> Generator[None, None, None]:
+    envdir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, version)
+    with envcontext(get_env_patch(envdir, version)):
         yield
 
 
@@ -39,15 +131,23 @@ def install_environment(
         version: str,
         additional_dependencies: Sequence[str],
 ) -> None:
-    helpers.assert_version_default('golang', version)
     env_dir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, version)
+
+    if version != 'system':
+        _install_go(version, env_dir)
 
     if sys.platform == 'cygwin':  # pragma: no cover
         gopath = cmd_output('cygpath', '-w', env_dir)[1].strip()
     else:
         gopath = env_dir
+
     env = dict(os.environ, GOPATH=gopath)
     env.pop('GOBIN', None)
+    if version != 'system':
+        env['GOROOT'] = os.path.join(env_dir, '.go')
+        env['PATH'] = os.pathsep.join((
+            os.path.join(env_dir, '.go', 'bin'), os.environ['PATH'],
+        ))
 
     helpers.run_setup_cmd(prefix, ('go', 'install', './...'), env=env)
     for dependency in additional_dependencies:
@@ -64,5 +164,5 @@ def run_hook(
         file_args: Sequence[str],
         color: bool,
 ) -> tuple[int, bytes]:
-    with in_env(hook.prefix):
+    with in_env(hook.prefix, hook.language_version):
         return helpers.run_xargs(hook, hook.cmd, file_args, color=color)
--- a/tests/languages/golang_test.py
+++ b/tests/languages/golang_test.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+import re
+from unittest import mock
+
+import pytest
+
+import pre_commit.constants as C
+from pre_commit.languages import golang
+from pre_commit.languages import helpers
+
+
+ACTUAL_GET_DEFAULT_VERSION = golang.get_default_version.__wrapped__
+
+
+@pytest.fixture
+def exe_exists_mck():
+    with mock.patch.object(helpers, 'exe_exists') as mck:
+        yield mck
+
+
+def test_golang_default_version_system_available(exe_exists_mck):
+    exe_exists_mck.return_value = True
+    assert ACTUAL_GET_DEFAULT_VERSION() == 'system'
+
+
+def test_golang_default_version_system_not_available(exe_exists_mck):
+    exe_exists_mck.return_value = False
+    assert ACTUAL_GET_DEFAULT_VERSION() == C.DEFAULT
+
+
+ACTUAL_INFER_GO_VERSION = golang._infer_go_version.__wrapped__
+
+
+def test_golang_infer_go_version_not_default():
+    assert ACTUAL_INFER_GO_VERSION('1.19.4') == '1.19.4'
+
+
+def test_golang_infer_go_version_default():
+    version = ACTUAL_INFER_GO_VERSION(C.DEFAULT)
+
+    assert version != C.DEFAULT
+    assert re.match(r'^\d+\.\d+\.\d+$', version)
--- a/tests/repository_test.py
+++ b/tests/repository_test.py
@@ -380,17 +380,36 @@ def test_swift_hook(tempdir_factory, store):
     )
 
 
-def test_golang_hook(tempdir_factory, store):
+def test_golang_system_hook(tempdir_factory, store):
     _test_hook_repo(
         tempdir_factory, store, 'golang_hooks_repo',
-        'golang-hook', [], b'hello world\n',
+        'golang-hook', ['system'], b'hello world from system\n',
+        config_kwargs={
+            'hooks': [{
+                'id': 'golang-hook',
+                'language_version': 'system',
+            }],
+        },
+    )
+
+
+def test_golang_versioned_hook(tempdir_factory, store):
+    _test_hook_repo(
+        tempdir_factory, store, 'golang_hooks_repo',
+        'golang-hook', [], b'hello world from go1.18.4\n',
+        config_kwargs={
+            'hooks': [{
+                'id': 'golang-hook',
+                'language_version': '1.18.4',
+            }],
+        },
     )
 
 
 def test_golang_hook_still_works_when_gobin_is_set(tempdir_factory, store):
     gobin_dir = tempdir_factory.get()
     with envcontext((('GOBIN', gobin_dir),)):
-        test_golang_hook(tempdir_factory, store)
+        test_golang_system_hook(tempdir_factory, store)
     assert os.listdir(gobin_dir) == []
 
 
@@ -677,7 +696,7 @@ def test_additional_golang_dependencies_installed(
     envdir = helpers.environment_dir(
         hook.prefix,
         golang.ENVIRONMENT_DIR,
-        C.DEFAULT,
+        golang.get_default_version(),
     )
     binaries = os.listdir(os.path.join(envdir, 'bin'))
     # normalize for windows
--- a/testing/resources/golang_hooks_repo/golang-hello-world/main.go
+++ b/testing/resources/golang_hooks_repo/golang-hello-world/main.go
@@ -3,7 +3,9 @@ package main
 
 import (
 	"fmt"
+	"runtime"
 	"github.com/BurntSushi/toml"
+	"os"
 )
 
 type Config struct {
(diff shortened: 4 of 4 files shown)


tests/git_test.py
```diff
@@ -282,3 +282,8 @@ def test_no_git_env():
 def test_init_repo_no_hooks(tmpdir):
     git.init_repo(str(tmpdir), remote='dne')
     assert not tmpdir.join('.git/hooks').exists()
+
+
+def test_get_remote_url(tmpdir):
+    git.init_repo(str(tmpdir), remote='dne')
+    assert git.get_remote_url(tmpdir) == 'dne'
```
asottile, line 289: if this function is dead it should just be removed -- originally this was building a `GOPATH`-like directory -- if that's not needed any more then there's a bunch of stuff that can be cleaned up I think ?

pre_commit/languages/golang.py
```diff
@@ -20,82 +24,139 @@
...
+
+
+def get_env_patch(
+        venv: str,
+        language_version: str,
+) -> PatchesT:
+    if language_version == 'system':
+        return (
+            ('PATH', (os.path.join(venv, 'bin'), os.pathsep, Var('PATH'))),
+        )
```
asottile, line 45: this is potentially going to install stuff outside of pre-commit's management isn't it?

pre_commit/languages/golang.py
```diff
@@ -20,82 +24,139 @@
...
 health_check = helpers.basic_health_check
 
 
-def get_env_patch(venv: str) -> PatchesT:
+@functools.lru_cache(maxsize=1)
+def get_default_version() -> str:
+    if helpers.exe_exists('go'):
+        return 'system'
+    else:
+        return '1.19.4'  # C.DEFAULT
```
asottile, line 35: this doesn't seem right?

pre_commit/languages/golang.py
```diff
@@ -20,82 +24,139 @@
...
+            raise
     else:
-        return 'unknown_src_dir'
+        with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+            shutil.copyfileobj(resp, tmp_file)
+            shutil.unpack_archive(
+                tmp_file.name,
+                dest,
+                format='zip' if sys.platform == 'win32' else 'gztar',
+            )
```
asottile, line 98: hmm I thought the whole point of unpack_archive was that it's automatic?

tests/git_test.py
taoufik07: Only get_remote_url is left I think. Done!

pre_commit/languages/golang.py
```diff
@@ -20,82 +24,139 @@
...
+    arch = platform.machine().lower()
+    if arch in ('x86_64',):
+        return 'amd64'
+    if arch in ('i386',):
+        return '386'
+    if arch in ('armv6l', 'armv7l'):
+        return 'armv6l'
+    if arch in ('arm64', 'aarch64', 'armv8'):
+        return 'arm64'
+    return arch
```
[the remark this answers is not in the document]
taoufik07: done... I think

pre_commit/languages/golang.py
taoufik07: No, since I've set a `GOPATH` later in the code which now I think can be moved to here

pre_commit/languages/golang.py
taoufik07: I was wondering if:
- I should return directly a default version `X.Y.Z`?
- or I should return `C.DEFAULT` and translate it somewhere in the code to a default version `X.Y.Z`?

pre_commit/languages/golang.py
taoufik07: That's what I thought at first, but when not provided, the format is only inferred using the filename extension.

asottile: should be easy to preserve the original extension right?

pre_commit/languages/golang.py
asottile: yeah this should just be `'default'` -- we absolutely cannot have a specific version hardcoded because I do not want to deal with spammy PRs bumping it all the time

pre_commit/languages/golang.py
taoufik07: The only way I can think of is to add it as a suffix to the named temp file (basically move the whole ternary there)

asottile: the url has the extension, no?

pre_commit/languages/golang.py
taoufik07: > yeah this should just be `'default'` -- we absolutely cannot have a specific version hardcoded because I do not want to deal with spammy PRs bumping it all the time

and translate it somewhere in the code to a default version `X.Y.Z`?  (this won't shoo spammy PRs, and I don't think Go provides an alias for stable or latest version)

asottile: yeah I can't accept that -- you'll have to figure some way to determine the latest available version

pre_commit/languages/golang.py
taoufik07: I've suffixed the whole name, might be useful for someone someday inspecting the temp dir :D

pre_commit/languages/golang.py
taoufik07: Looks it good now?

taoufik07:
@asottile Could you please approve me running workflows? Congrats on the migration btw 🎉

asottile:
if we don't need to build special `GOPATH` directories any more -- can you split that out to a separate PR so this is more straightforward?

pre_commit/languages/golang.py
```diff
@@ -39,30 +119,36 @@ def install_environment(
...
-    helpers.assert_version_default('golang', version)
     env_dir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, version)
 
-    if sys.platform == 'cygwin':  # pragma: no cover
-        gopath = cmd_output('cygpath', '-w', env_dir)[1].strip()
-    else:
-        gopath = env_dir
-    env = dict(os.environ, GOPATH=gopath)
-    env.pop('GOBIN', None)
+    with in_env(prefix, version):
```
asottile, line 124: I had factored this out in your other patch -- this shouldn't be needed here I think

testing/resources/golang_versioned_hooks_repo/go.mod
```diff
@@ -0,0 +1,5 @@
+module golang-hello-world
+
+go 1.18
+
+require github.com/BurntSushi/toml v1.1.0
```
asottile, line 5: you can probably take the existing local go test and set `language_version` -- that'd be simpler than a whole separate repo

tests/repository_test.py
```diff
@@ -380,20 +380,58 @@ def test_swift_hook(tempdir_factory, store):
     )
 
 
-def test_golang_hook(tempdir_factory, store):
+@pytest.fixture
+def _golang_lru_cache_clear():
+    golang.get_default_version.cache_clear()
+    yield
+    golang.get_default_version.cache_clear()
```
asottile, line 387: this shouldn't be needed

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
+
+    return (
+        ('GOROOT', os.path.join(venv, '.go')),
+        (
+            'PATH', (
+                os.path.join(venv, 'bin'), os.pathsep,
+                os.path.join(venv, '.go', 'bin'), os.pathsep, Var('PATH'),
+            ),
+        ),
+    )
```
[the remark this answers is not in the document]
taoufik07: All I can think of are some builtin commands that are useful to hookify (😆) and require access to a Go instance e.g. `gofmt`, `go generate`, `go mod ...` , `go vet`.
 
The question now is; do we want to give each an independent env? or expect/require the user to have a system installation? what if he uses a language version manager -- will it be cumbersome to active it?

If no to independent env, then we can yeet it.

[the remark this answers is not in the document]
asottile: I've been thinking about this a bit -- and it probably benefits pre-commit to store languages separate from the environments maybe?  that way they can all be shared instead of having N copies of `go` -- I'm not sure what it would look like yet (probably another entry in the store database?)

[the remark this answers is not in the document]
taoufik07: I had the same thought when I started working on this PR. It will mainly benefit the end user, thus pre-commit, especially when dealing with e.g. large sized languages in this case Go. Speaking of which, Go is a good candidate in this case, as it offers a simple way of achieving what we're looking for. I'm not sure if other languages offer the same experience -- I think that we can leverage `virtualenv` and `rbenv` for `python` and `ruby` -- if it's too complex or not worth it, we can have both structures coexist, maybe.

As a start, I think a very simple solution would be to pass a computed path to a language dir, install the language there, and make available a helper function that returns the language binary, this won't require a new database entry, but again this is a very simple solution based on my little knowledge of the tool internals.

pre_commit/languages/golang.py
taoufik07: Mainly because both hook install and run steps have shared env vars since I kept `go` accessible.

I've refactored it now after fixing #2677.

testing/resources/golang_versioned_hooks_repo/go.mod
taoufik07: done, now we only have one repo -- I have added a little arg in the `main.go`  that helps tell when running Go system, I'm not sure if it's sustainable.

tests/repository_test.py
taoufik07: I have added two tests, where I test default version resolving to system and to fetching latest version, so I needed to clear the `get_default_version` cache. Probably might be a stretch since we're already testing `get_default_version`, let me know if you think I should get rid of the tests.

pre_commit/languages/golang.py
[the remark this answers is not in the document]
asottile: yeah I think I'll design something separate from this -- I'm imaginging any "first class" language will support this (so ruby, js, rust, and now go) -- and any second class language (python) will continue doing nothing.  I'll need to think of the right api for this though (and hopefully something that makes it ~easy for pre-commit.ci to use as well)

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
+
+
+def _get_arch() -> str:  # pragma: no cover
```
asottile, line 56: I don't think this needs `no cover` any more

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
+    }
+    return _ALIASES.get(arch, arch)
+
+
+@functools.lru_cache
+def _infer_go_version(version: str) -> str:
+    if version != C.DEFAULT:
+        return version
+    resp = urllib.request.urlopen('https://go.dev/dl/?mode=json')
+    return json.loads(resp.read())[0]['version'].lstrip('go')
```
asottile, line 73: `json.load(resp)` is ~slightly more efficient -- `lstrip` isn't quite right here -- you probably want `removeprefix` but that's new in python 3.9

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
+
+
+def _get_url(version: str) -> str:
+    os_name = platform.system().lower()
+    ext = 'zip' if os_name == 'windows' else 'tar.gz'
```
asottile, line 78: the rest of the checks in the codebase use `sys.platform` -- just to be consistent let's do that here as well

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
-        ('PATH', (os.path.join(venv, 'bin'), os.pathsep, Var('PATH'))),
+        f'https://dl.google.com/go/go{version}.{os_name}-{_get_arch()}.{ext}'
     )
 
 
+def _install_go(version: str, dest: str) -> None:
+    try:
+        url = _get_url(version)
+        resp = urllib.request.urlopen(url)
+    except urllib.error.HTTPError as exec:  # pragma: no cover
```
asottile, line 89: `exec` is a bit of a weird name here -- if it's short for `exception` idk where the second `e` comes from 😆 -- I would call it just `e` or `exc` personally

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
+        if exec.code == 404:
+            raise ValueError(
+                'Could not find a version matching your system requirements '
+                f'(os={platform.system().lower()}; arch={_get_arch()})',
+            ) from exec
+        else:
+            raise
+    else:
+        with tempfile.NamedTemporaryFile(
+                delete=False,
```
asottile, line 99: I don't think this is right? this leaves the archive around after

pre_commit/languages/golang.py
```diff
@@ -17,20 +25,92 @@
...
+        shutil.move(os.path.join(dest, 'go'), os.path.join(dest, '.go'))
+
+
 @contextlib.contextmanager
-def in_env(prefix: Prefix) -> Generator[None, None, None]:
-    envdir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, C.DEFAULT)
-    with envcontext(get_env_patch(envdir)):
+def in_env(
+        prefix: Prefix,
+        language_version: str,
```
asottile, line 110: in a WIP branch I've standardized this function and with the name `version: str` (it'll also fit on one line) -- if you could do that here that'll save me a semantic conflict :D

pre_commit/languages/golang.py
```diff
@@ -39,15 +119,24 @@ def install_environment(
...
     env_dir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, version)
 
+    if version != 'system':
+        _install_go(version, env_dir)
+
     if sys.platform == 'cygwin':  # pragma: no cover
         gopath = cmd_output('cygpath', '-w', env_dir)[1].strip()
     else:
-        gopath = env_dir
+        gopath = os.path.join(env_dir)
```
asottile, line 130: `os.path.join` isn't doing anything here

pre_commit/languages/golang.py
```diff
@@ -39,15 +119,24 @@ def install_environment(
...
+
     env = dict(os.environ, GOPATH=gopath)
     env.pop('GOBIN', None)
+    if version != 'system':
+        env['GOROOT'] = os.path.join(env_dir, '.go')
+        env['PATH'] = (
+            os.path.join(env_dir, '.go', 'bin') +
+            os.pathsep + os.environ['PATH']
+        )
```
asottile, line 139: I would do `env['PATH'] = os.pathsep.join((os.path.join(env_dir, '.go', 'bin'), os.environ['PATH']))`

tests/languages/golang_test.py
```diff
@@ -0,0 +1,45 @@
...
+    assert ACTUAL_INFER_GO_VERSION('1.19.4') == '1.19.4'
+
+
+def test_golang_infer_go_version_default():
+    version = ACTUAL_INFER_GO_VERSION(C.DEFAULT)
+    assert version != C.DEFAULT
+
+    golang.get_default_version.cache_clear()
+    resp = urllib.request.urlopen(golang._get_url(version))
+    assert resp.code == 200
```
asottile, line 45: this is going to be pretty slow -- not sure it improves the test -- maybe just assert that it looks like a version?

tests/repository_test.py
asottile: you can set the language version directly for these tests -- you can pass in both `default` and `system`

pre_commit/languages/golang.py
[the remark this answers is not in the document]
taoufik07: Oh! python is a second class language, I forgot that, knowing that I've read it in the CONTRIBUTING docs

pre_commit/languages/golang.py
taoufik07: On windows, we can't open the named temp file a second time without first closing it, hence the `delete=False`. I've delegated the cleanup to the OS.

asottile: the store database has the same problem -- check out what's done there.

some people are ok delegating temp cleanup to the OS, I am not :P

pre_commit/languages/golang.py
taoufik07: done

pre_commit/languages/golang.py
taoufik07: done

pre_commit/languages/golang.py
taoufik07: done

pre_commit/languages/golang.py
taoufik07: Muscle memory typo probably 😅 -- done!

pre_commit/languages/golang.py
taoufik07: Today I learned that Windows doesn't automatically cleanup the temp dir. I've made a little context manager to perform the cleanup.

pre_commit/languages/golang.py
taoufik07: done!

pre_commit/languages/golang.py
taoufik07: done

pre_commit/languages/golang.py
taoufik07: done!

tests/languages/golang_test.py
taoufik07: done

tests/repository_test.py
taoufik07: done!

pre_commit/languages/golang.py
```diff
@@ -39,15 +135,22 @@ def install_environment(
...
     env_dir = helpers.environment_dir(prefix, ENVIRONMENT_DIR, version)
 
+    if version != 'system':
+        _install_go(version, env_dir)
+
+    gopath = env_dir
     if sys.platform == 'cygwin':  # pragma: no cover
         gopath = cmd_output('cygpath', '-w', env_dir)[1].strip()
-    else:
-        gopath = env_dir
```
asottile, line 48: the original code was better

tests/repository_test.py
```diff
@@ -380,17 +401,56 @@ def test_swift_hook(tempdir_factory, store):
...
+                tempdir_factory, store, 'golang_hooks_repo',
+                'golang-hook', [], b'hello world from go1.18.4\n',
+                config_kwargs={
+                    'hooks': [{
+                        'id': 'golang-hook',
+                        'language_version': 'default',
+                    }],
+                },
+            )
+            _infer_go_mck.assert_called_once_with(C.DEFAULT)
```
asottile, line 434: I don't think the mocks are useful here?  can't this just make an assertion based on the real execution and output?  this test is just a reimplementation of the implementation so it's not actually testing anything (same for the mock above as well)

pre_commit/languages/golang.py
```diff
@@ -17,20 +26,107 @@
...
+    """
+    f = tempfile.NamedTemporaryFile(**kwargs, delete=False)
+    try:
+        yield f
+    finally:  # pragma: no cover
+        try:
+            if delete:
+                os.unlink(f.name)
+        except OSError:
+            pass
```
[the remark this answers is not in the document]
taoufik07: I'll give it a look -- the `no cover` was necessary due to a coverage bug on the `if delete` https://github.com/nedbat/coveragepy/issues/867

I'm not sure if `seek(0)` will do the trick -- `shutil.unpack_archive` only accepts the filename.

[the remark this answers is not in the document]
taoufik07: It should be good now

pre_commit/languages/golang.py
taoufik07: done

tests/repository_test.py
```diff
@@ -62,21 +62,42 @@ def _get_hook(repo_config, store, hook_id):
     return hook
 
 
-def _test_hook_repo(
+def _try_hook_repo(
```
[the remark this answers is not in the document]
taoufik07: done

tests/repository_test.py
taoufik07: done

asottile:
I got rid of unpack_archive -- it's way simpler without it

asottile (approved):

PR #2655, opened 2022-12-27
drop python<3.8

asottile:


jkugler:
This doesn't like to an issue, so I am wondering why this was done. Python 3.7 is not EOL yet.

asottile:
because it's nonzero effort to support 3.7 and this release needed some stuff that's 3.8+ -- if you would like paid support for older python versions please reach out to me privately

PR #2671, opened 2023-01-02
simplify install state

asottile:
the additional bookkeeping has been unnecessary since b827694520be0f39bfc0599f3680b6c08b4516cf

unfortunately this will cause a rebuild of all hooks in order to be forward/backward compatible -- shrugs


asottile:
pre-commit.ci run

asottile:
hmm while I think this is a good simplification I actually think the reinstall is more annoying than I thought it would be so I'm going to back this out for now

PR #2677, opened 2023-01-05
Fix command normalization when a custom env is passed

taoufik07:
Custom envs can't be passed to `normalize_cmd` which causes `find_executable` to default to `os.environ` -- works as expected when injecting directly `os.environ`.

### DIFF
--- a/pre_commit/parse_shebang.py
+++ b/pre_commit/parse_shebang.py
@@ -20,13 +20,13 @@ def parse_filename(filename: str) -> tuple[str, ...]:
 
 
 def find_executable(
-        exe: str, _environ: Mapping[str, str] | None = None,
+        exe: str, *, env: Mapping[str, str] | None = None,
 ) -> str | None:
     exe = os.path.normpath(exe)
     if os.sep in exe:
         return exe
 
-    environ = _environ if _environ is not None else os.environ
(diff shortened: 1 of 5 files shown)


pre_commit/parse_shebang.py
```diff
@@ -20,7 +20,7 @@ def parse_filename(filename: str) -> tuple[str, ...]:
 
 
 def find_executable(
-        exe: str, _environ: Mapping[str, str] | None = None,
+        exe: str, *, _environ: Mapping[str, str] | None = None,
```
asottile, line 23: if it's public it shouldn't be underscored. `env` is probably what it sounds be called to match subprocess

tests/commands/install_uninstall_test.py
```diff
@@ -278,6 +278,7 @@ def test_environment_not_sourced(tempdir_factory, store):
         homedir = tempdir_factory.get()
         ret, out = git_commit(
             env={
+                **os.environ,
```
asottile, line 281: hmmm why did this need to change?

pre_commit/parse_shebang.py
taoufik07: done!

asottile (approved):

PR #2713, opened 2023-01-19
test golang directly

m-rsha:
I totally forgot to change the stuff in `repository_test.py`, but was hoping these tests would run on the CI servers. I can't get my new test to pass.

I was getting golang syntax errors about newlines in strings until I made `hello_world_go` a raw string, but now I either get a connection reset by peer while inside `_install_go` or it just fails to connect for ~five minutes before I interrupt it.

I'm guessing it's because I don't have go's toolchain installed, so hopefully the CI runs answer some stuff for me!

----

Okay, I derped again and forgot that making it a raw string while starting with a backslash is dumb lol, so I'm just gonna head to bed tonight. I'll have to poke at it more tomorrow!


tests/languages/golang_test.py
```diff
@@ -41,3 +45,100 @@ def test_golang_infer_go_version_default():
...
+
+type Config struct {
+        What string
+}
+
+func main() {
+        message := runtime.Version()
+        if len(os.Args) > 1 {
+                message = os.Args[1]
+        }
```
asottile, line 81: you can remove the args stuff -- this isn't actually testing anything

tests/repository_test.py
```diff
@@ -232,92 +229,6 @@ def test_system_hook_with_spaces(tempdir_factory, store):
...
-    cmd_output('git', '-C', str(repo), 'init', '.')
-    cmd_output('git', '-C', str(repo), 'add', '.')
-    cmd_output('git', '-C', str(repo), 'submodule', 'add', str(sub), 'sub')
-    git.commit(str(repo))
-
-    config = make_config_from_repo(str(repo))
-    hook = _get_hook(config, store, 'example')
-    ret, out = _hook_run(hook, (), color=False)
-    assert ret == 0
-    assert _norm_out(out) == b'hello hello world\n'
```
asottile, line 318: hmmm, I don't think there's anything testing that submodules are checked out any more -- this'll need to be replaced with a `language: script` test which has submodules

asottile (approved):

Issue #2582, opened 2022-11-04
Support for `pre-rebase` Hook


BrutalSimplicity:
Is there a pre-commit hook that can prevent rebase operations? It looks like git does support a `pre-rebase` hook, but I didn't see it in the docs.

I checked [here](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages).

Example `pre-rebase` hook:

https://gist.github.com/uasi/9384329

asottile:
here's some PRs you can draw inspiration from -- #2036 #1800

BrutalSimplicity:
Never mind, I think I'm actually looking for the standard `commit` hook functionality. I want to be able to prevent certain commits from being rewritten, which I _think_ that hook can do. Apologies. 😅 

I'll still leave this open, since it is functionality that doesn't exist.

BrutalSimplicity:
@asottile I appreciate the quick response. Thank you! Not sure if I'll have time soon to add this, but when I have more time, I'll try to take a look if this is still open.

asottile:
if you're trying to hook into individual "commits" that (for example) an interactive rebase makes unfortunately I don't think there's a way to do that (`git rebase --continue` skips hooks in my experience)

asottile:
(that is, unless you manually call `git commit` with no arguments before continuing)

BrutalSimplicity:
Ahh, thanks for letting me know that. I guess server-side hooks are the way here. Any ideas on how you can implement a server-side hook that will reject a push if a "protected" commit message is modified?

I don't think there is support in `git` for this concept, but some tools like to use git as a more traditional datastore, and if a certain commit is altered, the tools end up in an odd state.

As I was typing this it dawned on me that this could possibly be handled by a tagging strategy.

adamchainz:
I just [blogged about](https://adamj.eu/tech/2022/11/07/pre-commit-run-hooks-rebase/) a solution to run hooks during rebase:

```
$ git rebase -x 'pre-commit run --from-ref HEAD~ --to-ref HEAD' main
```

jawomg:
เพิ่มการรองรับ hook แบบ pre-rebase #2725

PR #2725, opened 2023-01-27
Add pre-rebase hook support

mgaligniana:
Hello!

This PR adds the `pre-rebase` hook

Thank you!

Fixes: https://github.com/pre-commit/pre-commit/issues/2582


pre_commit/commands/hook_impl.py
```diff
@@ -175,6 +177,7 @@ def _pre_push_ns(
     'post-merge': 1,
     'post-rewrite': 1,
     'pre-push': 2,
+    'pre-rebase': 1,
```
mgaligniana, line 180: That it should be `2`? Following the documentation it can receive two params: https://git-scm.com/docs/githooks#_pre_rebase

pre_commit/constants.py
```diff
@@ -14,13 +14,13 @@
...
     'post-commit', 'manual', 'post-checkout', 'push', 'post-merge',
-    'post-rewrite',
+    'post-rewrite', 'rebase',
 )
 
 HOOK_TYPES = (
     'pre-commit', 'pre-merge-commit', 'pre-push', 'prepare-commit-msg',
     'commit-msg', 'post-commit', 'post-checkout', 'post-merge',
-    'post-rewrite',
+    'post-rewrite', 'rebase',
```
asottile, line 23: this should be called `pre-rebase` -- the original `commit` stage is a mistake that's too late to fix now

pre_commit/main.py
```diff
@@ -147,6 +147,13 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
             'the rewrite'
         ),
     )
+    parser.add_argument(
+        '--is-rebase',
+        help=(
+            'During a pre-rebase hook, specifies the command that invoked '
+            'the rebase'
+        ),
+    )
```
asottile, line 156: this isn't right -- hook-impl should collect the 1 or 2 arguments from https://git-scm.com/docs/githooks#_pre_rebase

pre_commit/constants.py
mgaligniana: I see, thank you! Applied!

pre_commit/main.py
mgaligniana: Done!

mgaligniana:
@asottile sorry for the delayed updates. Could I ask you for help one more time? I can't find why the `test_pre_rebase_integration` test is not creating the `pre-rebase.tmp` file

pre_commit/constants.py
```diff
@@ -14,13 +14,13 @@
 STAGES = (
     'commit', 'merge-commit', 'prepare-commit-msg', 'commit-msg',
     'post-commit', 'manual', 'post-checkout', 'push', 'post-merge',
-    'post-rewrite',
+    'post-rewrite', 'rebase',
```
asottile, line 17: the `commit` and `push` stages are a mistake that I'm trying to fix -- can you make this match the hook name `pre-rebase` instead of `rebase`?

pre_commit/commands/hook_impl.py
```diff
@@ -231,6 +232,8 @@ def _run_ns(
         return _ns(hook_type, color, is_squash_merge=args[0])
     elif hook_type == 'post-rewrite':
         return _ns(hook_type, color, rewrite_command=args[0])
+    elif hook_type == 'pre-rebase':
+        return _ns(hook_type, color, from_ref=args[0], to_ref=args[1])
```
asottile, line 236: the docs seem to indicate that _sometimes_ this has 2 args, sometimes it has 1 -- I'm not sure that `from-ref` and `to-ref` arguments are correct here (they're meant for `pre-push`)?

pre_commit/constants.py
mgaligniana: Applied!

pre_commit/commands/hook_impl.py
mgaligniana: Args changed, let me know if naming are OK

mgaligniana:
@asottile Sorry for this taking me too long, I was busy the past month but I'm still trying to solve this issue!

I think I'm not understanding how the pre-commit's flow works because I couldn't fix the `test_pre_rebase_integration` at `install_uninstall_test.py` and I don't understand why 😞 .

Could I ask you for help one more time? In case I'm delaying the fix I can publish it in the discord channel or I would love review it during streaming 👾 !

Thank you!

pre_commit/commands/hook_impl.py
```diff
@@ -73,6 +73,8 @@ def _ns(
         local_branch: str | None = None,
         from_ref: str | None = None,
         to_ref: str | None = None,
+        upstream: str | None = None,
+        branch: str | None = None,
```
asottile, line 77: we'll need corresponding arguments in the `run` subcommand -- and they probably should be namespaced like `--pre-rebase-upstream` and `--pre-rebase-branch`

tests/commands/hook_impl_test.py
```diff
@@ -100,6 +100,8 @@ def call(*_, **__):
         ('commit-msg', ['.git/COMMIT_EDITMSG']),
         ('post-commit', []),
         ('post-merge', ['1']),
+        ('pre-rebase', ['master', 'topic']),
+        ('pre-rebase', ['master']),
```
asottile, line 104: please use `main`

tests/commands/install_uninstall_test.py
```diff
@@ -810,6 +810,46 @@ def test_post_merge_integration(tempdir_factory, store):
         assert os.path.exists('post-merge.tmp')
 
 
+def test_pre_rebase_integration(tempdir_factory, store):
```
asottile, line 813: I suspect this isn't passing due to the `.replace('pre-', '')` bit -- I want to factor that out anyway so maybe 2 birds with one stone? https://github.com/pre-commit/pre-commit/issues/2732

tests/commands/hook_impl_test.py
mgaligniana: Sure!

pre_commit/commands/hook_impl.py
mgaligniana: Done!

tests/commands/install_uninstall_test.py
mgaligniana: Yes! I tested removing that `replace` and the test passed! I leave it as it is.

How I can help with that refactor?

asottile (approved):
good code changed like a ghost

PR #2728, opened 2023-01-29
fix r local hooks

asottile:
`language: r` acts more like `language: script` so we have to *not* append
the prefix when run with `repo: local`


lorenzwalthert:
Thanks a lot @asottile to make this work for the R local hooks with script path too, I see the changes are not trivial ❤️

Issue #776, opened 2018-07-02
git apply error when using git-crypt


ranimufid:
Hi. I am currently using `git-crypt` in tandem with some `pre-commit` hooks

Upon adding a new file for encryption --> a corresponding `.gitattributes` entry --> `git commit -m "blah"` I encounter the following error:
```
An unexpected error has occurred: CalledProcessError: Command: ('/usr/local/Cellar/git/2.18.0/libexec/git-core/git', '-c', 'core.autocrlf=false', 'apply', '--whitespace=nowarn', '/Users/bob/.cache/pre-commit/patch1530518673')
Return code: 128
Expected return code: 0
Output: (none)
Errors:
    error: unrecognized input
```

**git version:** `2.18.0`
**pre-commit version:** `1.10.2`
**.pre-commit-config.yaml**
```
repos:
-   repo: git://github.com/pre-commit/pre-commit-hooks
    rev: v1.3.0
    hooks:
    -   id: trailing-whitespace
    -   id: check-yaml
    -   id: end-of-file-fixer
-   repo: https://github.com/willthames/ansible-lint.git
    rev: v3.4.23
    hooks:
    - id: ansible-lint
      entry: ansible-lint bootstrap.yml
```

**.gitattributes:**
```
hosts/testing/secret-vars.yml filter=git-crypt diff=git-crypt
```

asottile:
Thanks for the report! could you provide some more information (if possible) for me?

it appears from the output that you committed with an unstaged change, to which file was that and if possible could you share the patch file mentioned?

I suspect the `diff=git-crypt` is somehow throwing off the way we generate and then reapply patches and will need to be worked around.

ranimufid:
thanks for your prompt response @asottile. 

**Context:**
I'm trying to encrypt some ansible secrets using git-crypt. I'm also doing `ansible-lint` as a pre-commit hook.

>it appears from the output that you committed with an unstated change

The change which triggered the error is creating an empty file `/hosts/testing/secret-vars.yml` with a corresponding entry in `.gitattributes`

**.gitattributes**
```
hosts/testing/secret-vars.yml filter=git-crypt diff=git-crypt
```

>to which file was that and if possible could you share the patch file mentioned

The patch file corresponding to the above change

```
diff --git a/hosts/testing/secret-vars.yml b/hosts/testing/secret-vars.yml
GIT binary patch
literal 0
HcmV?d00001

literal 22
[...]

PR #2743, opened 2023-02-02
Add `--no-textconv` to `git diff` calls

### LINKED ISSUE #2742: Binary file change by hook not detected due to Git textconv

textconv

### describe your issue

I’ve set up Git to use `exiftool` as a text conversion filter for JPEGs, with this in `~/.config/git/config`:

```ini
[diff "exiftool"]
    textconv = exiftool --composite -x 'Exiftool:*' -x 'File:*' -g0
    cachetextconv = true
    xfuncname = "^-.*$"
```

…and this in `~/.config/git/attributes`:

```
*.jpeg diff=exiftool
```

In my repo, I’m using `jpegoptim` to optimize JPEG's with the below `.pre-commit-config.yaml`.

Take an unoptimized JPEG:

```console
$ git status
On branch main
Changes to be committed:
        new file:   donut.jpeg
```

Running pre-commit, the hook passes:

```console
$ pre-commit run jpegoptim
Optimize JPEGs...........................................................Passed
```

(There’s no option to make `jpegoptim` exit with a non-zero status code when it changes files.)

However, the tool did actually modify the image:

```console
$ git status
On branch main
Changes to be committed:
        new file:   donut.jpeg

Changes not staged for commit:
        modified:   donut.jpeg
```

This behaviour means commits don't include the optimized version of a flie, instead they’re left around unstaged.

pre-commit doesn't detect the change because the text conversions of the binary file before/after are the same, leading to an empty diff:

```
$ git diff donut.jpeg
```

But using [`--no-textconv`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---no-textconv) shows the tool actually made a change:

$ git diff --no-textconv donut.jpeg |cat
diff --git a/donut.jpeg b/donut.jpeg
index 3bd0743..d66a2f4 100644
Binary files a/donut.jpeg and b/donut.jpeg differ
```

The `git diff` call that pre-commit uses to check for changes, in [`_git_diff()`](https://github.com/pre-commit/pre-commit/blob/e846829992a84ce8066e6513a72a357709eec56c/pre_commit/commands/run.py#L273C6-L277) already uses `--no-ext-diff` to make diffs reproducible
I think it should also use `--no-textconv`.
This would also make it a little faster.

### pre-commit --version

3.0.3 / main

### .pre-commit-config.yaml

```yaml
repos:
-   repo: local
    hooks:
    -   id: jpegoptim
        name: Optimize JPEGs
        entry: jpegoptim
        language: system
        types: [jpeg]
```

_No response_

adamchainz:
Fixes #2742.

I fixed the one in `run.py`, fixing the bug. Ithen audited the other `git diff` calls to consider whether text conversion would be suitable there. Those using `--name-only` don't need it, since Git skips text conversion in that case. The diff in `_run_hooks` is for human consumption doesn't need it because it's for the human to read.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -272,7 +272,8 @@ def _all_filenames(args: argparse.Namespace) -> Collection[str]:
 
 def _get_diff() -> bytes:
     _, out, _ = cmd_output_b(
-        'git', 'diff', '--no-ext-diff', '--ignore-submodules', check=False,
+        'git', 'diff', '--no-ext-diff', '--no-textconv', '--ignore-submodules',
+        check=False,
     )
     return out
 
@@ -326,8 +327,7 @@ def _has_unmerged_paths() -> bool:
 
 def _has_unstaged_config(config_file: str) -> bool:
     retcode, _, _ = cmd_output_b(
-        'git', 'diff', '--no-ext-diff', '--exit-code', config_file,
-        check=False,
+        'git', 'diff', '--quiet', '--no-ext-diff', config_file, check=False,
     )
     # be explicit, other git errors don't mean it has an unstaged config.
     return retcode == 1
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -766,6 +766,47 @@ def test_lots_of_files(store, tempdir_factory):
         )
 
 
+def test_no_textconv(cap_out, store, repo_with_passing_hook):
+    # git textconv filters can hide changes from hooks
+    with open('.gitattributes', 'w') as fp:
+        fp.write('*.jpeg diff=empty\n')
+
+    with open('.git/config', 'a') as fp:
+        fp.write('[diff "empty"]\n')
+        fp.write('textconv = "true"\n')
+
+    config = {
+        'repo': 'local',
+        'hooks': [
+            {
+                'id': 'extend-jpeg',
+                'name': 'extend-jpeg',
+                'language': 'system',
+                'entry': (
+                    f'{shlex.quote(sys.executable)} -c "import sys; '
+                    'open(sys.argv[1], \'ab\').write(b\'\\x00\')"'
+                ),
+                'types': ['jpeg'],
+            },
+        ],
+    }
+    add_config_to_repo(repo_with_passing_hook, config)
+
+    stage_a_file('example.jpeg')
+
+    _test_run(
+        cap_out,
+        store,
+        repo_with_passing_hook,
+        {},
+        (
+            b'Failed',
+        ),
+        expected_ret=1,
+        stage=False,
+    )
+
+
 def test_stages(cap_out, store, repo_with_passing_hook):
     config = {
         'repo': 'local',


pre_commit/staged_files_only.py
```diff
@@ -51,8 +51,8 @@ def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
     tree = cmd_output('git', 'write-tree')[1].strip()
     retcode, diff_stdout_binary, _ = cmd_output_b(
         'git', 'diff-index', '--ignore-submodules', '--binary',
```
asottile, line 53: I think `--binary` implies `--no-textconv` -- or that it's unavoidable (one of the two) -- I did some extensive research on this here: https://github.com/pre-commit/pre-commit/issues/776#issuecomment-1299159293

pre_commit/git.py
```diff
@@ -173,7 +173,7 @@ def head_rev(remote: str) -> str:
 
 
 def has_diff(*args: str, repo: str = '.') -> bool:
-    cmd = ('git', 'diff', '--quiet', '--no-ext-diff', *args)
+    cmd = ('git', 'diff', '--quiet', '--no-ext-diff', '--no-textconv', *args)
```
asottile, line 176: I don't think textconv is invoked for `--quiet`?

pre_commit/commands/run.py
```diff
@@ -326,8 +327,8 @@ def _has_unmerged_paths() -> bool:
 
 def _has_unstaged_config(config_file: str) -> bool:
     retcode, _, _ = cmd_output_b(
-        'git', 'diff', '--no-ext-diff', '--exit-code', config_file,
-        check=False,
+        'git', 'diff', '--no-ext-diff', '--no-textconv', '--exit-code',
+        config_file, check=False,
```
asottile, line 331: this should probably be changed to `--quiet` instead

pre_commit/commands/run.py
```diff
@@ -272,7 +272,8 @@ def _all_filenames(args: argparse.Namespace) -> Collection[str]:
 
 def _get_diff() -> bytes:
     _, out, _ = cmd_output_b(
-        'git', 'diff', '--no-ext-diff', '--ignore-submodules', check=False,
+        'git', 'diff', '--no-ext-diff', '--no-textconv', '--ignore-submodules',
+        check=False,
```
asottile, line 276: can you add a test demoing this? otherwise it's just going to be forgotten

pre_commit/staged_files_only.py
adamchainz: I'm not seeing that locally, the `--binary` flag doesn't change results

```
$ git diff --binary

$ git diff --no-textconv
diff --git a/donut.jpeg b/donut.jpeg
index 3bd0743..d66a2f4 100644
Binary files a/donut.jpeg and b/donut.jpeg differ
```

It seems binary diffs are only shown if neither side of the diff is textconv'd: https://github.com/git/git/blob/2fc9e9ca3c7505bc60069f11e7ef09b1aeeee473/diff.c#L3558-L3595

pre_commit/git.py
adamchainz: Good point

pre_commit/staged_files_only.py
asottile: `diff-index` is different than `diff`

pre_commit/commands/run.py
adamchainz: Added a test, let me know if it's okay

pre_commit/staged_files_only.py
adamchainz: Right you are, missed that, reverting this change.

tests/commands/run_test.py
```diff
@@ -766,6 +766,47 @@ def test_lots_of_files(store, tempdir_factory):
...
 
 
+def test_no_textconv(cap_out, store, repo_with_passing_hook):
+    # git textconv filters can hide changes from hooks
+    with open('.gitattributes', 'w') as fp:
+        fp.write('*.jpeg diff=empty\n')
+
+    with open('.git/config', 'a') as fp:
+        fp.write('[diff "empty"]\n')
+        fp.write('textconv = "python -c print()"\n')
```
asottile, line 776: this should use `sys.executable` (or maybe even bash) so it doesn't fail on systems with only `python3`

adamchainz: Switched to `sys.executable`

asottile: eheh I think it might have quoting issues now /o\  we'll see what CI says

asottile: I would just use `echo` probably

adamchainz: `echo` is only a shell builtin on my system; `printf` is binary, but it will require printing its args, which can cause a textconv difference since `git diff` uses a temporary filename for comparison

I added `shlex.quote`...

asottile: I'm surprised you don't have /bin/echo, it's posix specced iirc

adamchainz: `/bin/echo` exists yes, I was just being slow and only checked `/usr/bin/`.

I have changed the test to use `true`.

asottile (approved):

PR #2796, opened 2023-03-04
prefer `sys.platform` over `os.name` when checking for windows OS

m-rsha:


pre_commit/languages/conda.py
```diff
@@ -26,7 +27,7 @@ def get_env_patch(env: str) -> PatchesT:
     # $CONDA_PREFIX/Scripts and $CONDA_PREFIX. Whereas the latter only
     # seems to be used for python.exe.
     path: SubstitutionT = (os.path.join(env, 'bin'), os.pathsep, Var('PATH'))
-    if os.name == 'nt':  # pragma: no cover (platform specific)
+    if sys.platform == 'win32':  # pragma: no cover (platform specific)
```
asottile, line 30: this can be `# pragma: win32 cover`

pre_commit/languages/python.py
```diff
@@ -137,7 +137,7 @@ def norm_version(version: str) -> str | None:
     elif _sys_executable_matches(version):  # virtualenv defaults to our exe
         return None
 
-    if os.name == 'nt':  # pragma: no cover (windows)
+    if sys.platform == 'win32':  # pragma: no cover (windows)
```
asottile, line 140: this can be `# pragma: win32 cover`

m-rsha: It borked the tests, so I changed it back :D. I guess this conversation still counts as "resolved"?

asottile (approved):

PR #2820, opened 2023-03-17
don't set CARGO_HOME in rust

asottile:
this adds a 270 MB registry cache in the output

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -50,7 +50,6 @@ def _rust_toolchain(language_version: str) -> str:
 
 def get_env_patch(target_dir: str, version: str) -> PatchesT:
     return (
-        ('CARGO_HOME', target_dir),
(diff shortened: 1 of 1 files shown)


daschuer:
At Mixxx we have an issue that the qml-fomatter installation is broken since the last pre-commit update. 
Could this be related? 
https://github.com/mixxxdj/mixxx/issues/11386

```
[INFO] Installing environment for https://github.com/qarmin/qml_formatter.git.
An unexpected error has occurred: CalledProcessError: command: ('rustup', 'toolchain', 'install', '--no-self-update', '1.64.0')
return code: 1
stdout:
    Executable `rustup` not found
stderr: (none)
Check the log at /github/home/.cache/pre-commit/pre-commit.log
Error: Process completed with exit code 3.
```

asottile:
it's possible? (but seems a little unlikely)  can you open a more detailed issue with the full log / reproduction?

Issue #1140, opened 2019-09-15
Support monorepos


dpinol:
Hi,
several people have asked about monorepos, which is a very popular trend:
https://github.com/pre-commit/pre-commit/issues?utf8=%E2%9C%93&q=is%3Aissue+monorepo+

Without explicit support, I need to specify the project path 3 times for pylint:
```
    -   id: pylint
        name: pylint
        entry: env PYTHONPATH=hubtype_babel/deploy_app python -m pylint.__main__ --rcfile=hubtype_babel/deploy_app/pylintrc
        language: system
        types: [python]
        files: hubtype_babel/deploy_app/.*/.*py
```
Maybe adding a new "working_folder" hook field which just performs a "cd" before running the hook could be a quick win?

thanks!

asottile:
please continue discussion in an existing issue, thanks

Issue #466, opened 2017-01-12
Guidance on pre-commit + monorepos


snakescott:
Any thoughts on using pre-commit in a monorepo (http://danluu.com/monorepo/) setting, with each project having its own `.pre-commit-config.yaml` file? For example, will pre-commit be happy/sad with

- project root not being repo root
- multiple invocations of `pre-commit install`
- etc

I expect to play around and experiment but any early guidance (e.g., "I have seen people try and fail at this, beware" or "Yelp actually does exactly this with steps X, Y, Z") is welcome!

asottile:
Aww man I typed up a response to this and never pressed reply.  I guess I'll try and remember what I said the first time!

Though it's not explicitly designed for a monorepo I can imagine one way you could use the existing configuration to support this goal with a single configuration file.

```yaml
# configure shared hooks that you want to run on all repositories here
-   repo: https://github.com/pre-commit/pre-commit-hooks.git
    sha: v0.6.1
    hooks:
    -   id: trailing-whitespace
    -   id: end-of-file-fixer

# For hooks that only apply to specific repositories, list them below
-   repo: https://github.com/pre-commit/pre-commit-hooks.git
    sha: v0.6.1
    hooks:
    -   id: flake8
        name: 'flake8 (for repo1)'
        # Only run on repo1 files
        files: '^repo1/.*\.py'
    -   id: flake8
        name: 'flake8 (for repo2)'
        # Only run on repo2 files
        files: '^repo2/.*\.py'
        # maybe repo2 wants specific argument configuration
        args: [--max-line-length=131]
    # maybe repo3 is python3 only
    -   id: flake8
        name: 'flake8 (for repo3)'
        files: '^repo3/.*\.py'
        language_version: python3

    # etc.
```

Note that the same `repo` can be configured multiple times and within the repo the same hook id can also be repeated

asottile:
@snakescott any additional thoughts on this, does this seem reasonable?

asottile:
Going to close this as answered, haven't heard back in a few months.  If this isn't sufficient, please comment and I'll reopen :)

alkuzad:
@asottile wouldn't that beneficial to create "folders" array which is compiled into turns into file regexp equal to `^(folder1|folder2|folder3){os.sep}*` ?

asottile:
I see no reason to have two ways to do things

dpinol:
The problem with this approach is that you need to specify the path for each project multiple times. 

[...]

PR #2833, opened 2023-04-01
feature: add monorepo multiple configurations

zp4rand0miz31:
Not to be merged, this is a POC to explore the monorepo feature.

This MR proposes a way to use multiple configuration, found recursively in the code tree so that multiple hooks can be applied. This is particularly useful when working with a monorepo, which is the case at my company.

This is currently a POC, only tested on Linux. No unit test were added at this stage of the proof of concept.

This relates to https://github.com/pre-commit/pre-commit/issues/466 and https://github.com/pre-commit/pre-commit/issues/1140 which were closed a long time ago. It seems the monorepo feature is not considered. Not sure where the discussion stands on this.

# current limitations

The current implementation does not use the subdirectory of staged files to run hooks. 
This creates several problems:
- any local configuration file (for instance pyproject.toml, .flake8 etc...) is taken from the root directory because hook are taken from root directory.
- hooks from multiple files can interfere, because they all share the same virtual environment 
- This also means a number of problem can happen when committing files from several directories in a single commit: every hook from all the impacted directories are ran on all the committed files.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -9,8 +9,9 @@ import re
 import subprocess
 import time
 import unicodedata
-from typing import Any, List
+from typing import Any
 from typing import Collection
+from typing import List
 from typing import MutableMapping
 from typing import Sequence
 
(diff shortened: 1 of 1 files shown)


asottile:
no means no

PR #2834, opened 2023-04-01
Fix double shebang in Windows hook template

mataha:
In its current form, `_install_hook_script` appends Windows-specific shebang on top of the previous one when it should replace it instead (probably). Here's what it looks like now:

```sh
#!/bin/sh
#!/usr/bin/env bash
# File generated by pre-commit: https://pre-commit.com
# ID: 138fd403232d2ddd5efb44317e38bf03

...
```

I got triggered by this, so I went ahead and wrote a fix.


asottile:
it is harmless

Issue #2835, opened 2023-04-03
SwiftLint doesn't compile on Swift 5.8


kgrobelny-intive:
### search you tried in the issue tracker

unknown option

### describe your issue

I've got pre-commit configured with SwiftLint.
Executing `pre-commit run --all-files` fails due to wrong `-C` option passed to `swift build`.

### pre-commit --version

3.2.1

### .pre-commit-config.yaml

```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.3.0
    hooks:
    -   id: end-of-file-fixer
    -   id: trailing-whitespace
-   repo: https://github.com/nicklockwood/SwiftFormat
    rev: 0.51.2
    hooks:
    -   id: swiftformat
-   repo: https://github.com/realm/SwiftLint
    rev: 0.50.3
    hooks:
    -   id: swiftlint
```

### version information

```
pre-commit version: 3.2.1
git --version: git version 2.39.2 (Apple Git-143)
sys.version:
    3.10.9 | packaged by conda-forge | (main, Feb  2 2023, 20:26:08) [Clang 14.0.6 ]
sys.executable: /opt/homebrew/Caskroom/mambaforge/base/envs/torch2/bin/python
os.name: posix
sys.platform: darwin
```

### error information

```
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/swift', 'build', '-C', '/Users/[REDACTED]/.cache/pre-commit/repoc7zisw50', '-c', 'release', '--build-path', '/Users/[REDACTED]/.cache/pre-commit/repoc7zisw50/swift_env-default/.build')
return code: 64
stdout: (none)
stderr:
    error: Unknown option '-C'
    Usage: swift build <options>
      See 'build -help' for more information.
```

```
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/mambaforge/base/envs/torch2/lib/python3.10/site-packages/pre_commit/error_handler.py", line 73, in error_handler
    yield
  File "/opt/homebrew/Caskroom/mambaforge/base/envs/torch2/lib/python3.10/site-packages/pre_commit/main.py", line 409, in main
    return run(args.config, store, args)
  File "/opt/homebrew/Caskroom/mambaforge/base/envs/torch2/lib/python3.10/site-packages/pre_commit/commands/run.py", line 442, in run
    install_hook_envs(to_install, store)
[...]

PR #2836, opened 2023-04-03
Change deprecated `swift build` `-C` command line option to replacement `--package-path`

### LINKED ISSUE #2235: Alternative to stashing files for testing
Are there any plans to implement alternatives to stashing the worktree?

Ideally this would be hook/scriptable, like some 'prepare-worktree' and 'restore-worktree' options (which default to the current stash behavior) but can also yield some new directory where the tests are run. The rationale here is that my editor reverts files changed on disk and I'd like to add notes to  source files while the commit is in progress.

In my own pre-commit hooks I use something like:

     git archive "$(git write-tree)" --prefix="$test_dir/" | tar xf -

To create a pristine source tree (actually, I also prime it with `cp -rl` with build artifacts from the previous build to speed up incremental builds). 'git-worktree' and other tools could be used as well...

Eventually I have the idea to run some (more expensive) pre-commit checks in the background while one types the  commit message. Then in the commit-msg hook wait for the background results and abort the commit there. This should reduce the turn around times significantly.

edelabar:
Based on this comment: https://github.com/pre-commit/pre-commit/issues/2835#issuecomment-1494873806

Sorry, I'm not a python dev and I don't see how to run unit tests or anything, or even how to test this locally, but looking at the swift docs it seems like this makes sense:

```
$ swift build -help
OVERVIEW: Build sources into binary products

SEE ALSO: swift run, swift package, swift test

USAGE: swift build <options>

OPTIONS:
  --package-path <package-path>
                          Specify the package path to operate on (default current directory). This changes the working directory before any other operation
...
```

### DIFF
--- a/pre_commit/languages/swift.py
+++ b/pre_commit/languages/swift.py
@@ -44,7 +44,7 @@ def install_environment(
     os.mkdir(envdir)
     cmd_output_b(
         'swift', 'build',
-        '-C', prefix.prefix_dir,
+        '--package-path', prefix.prefix_dir,
         '-c', BUILD_CONFIG,
         '--build-path', os.path.join(envdir, BUILD_DIR),
     )


asottile:
can you use a more descriptive commit message and/or PR title?

edelabar:
@asottile sorry, rushing... how's that?

PR #2845, opened 2023-04-17
some updates and fixes for newer python versions

marksmayo:
Didn't get around to it in my previous attempt, so doing it now.
 - updated deprecated .warn() to .warning()
 - unnecessary f-strings
 - imported sys to allow sys.exit calls instead of exit
 - simplified some not checks
 - replaced type with isinstance


asottile:
this improves nothing. don't send unrequested changes

PR #2858, opened 2023-04-29
Add ability to skip hooks by branches

vanya909:
Me and my team have faced with a problem when we want to run some hooks only on certain branches (like tests which is pretty long-run hook) so I've decided to provide such ability.


asottile:
please discuss features before wasting time on them

PR #2885, opened 2023-05-15
More robust R hooks for Windows

### LINKED ISSUE #2870: Avoid unquoting weirdness of Windows for  `language: r`

never, r, found

### describe your issue

Multiple reports in https://github.com/lorenzwalthert/precommit (https://github.com/lorenzwalthert/precommit/issues/441, https://github.com/lorenzwalthert/precommit/issues/473) were raised and describe a problem with (un)quoting the long string that runs when `language: r` is setup in `Rscript -e 'xxx'` where `'xxx'` contains [multiple levels of quotes](https://github.com/pre-commit/pre-commit/blob/6896025288691aafd015a4681c59dc105e61b614/pre_commit/languages/r.py#L101). For the readers convenience, the output looks like:
```
[INFO] Installing environment for https://github.com/lorenzwalthert/precommit.
[INFO] Restored changes from C:\Users\USER\.cache\pre-commit\patch1678401203-36472.
An unexpected error has occurred: CalledProcessError: command: ('C:/PROGRA~1/R/R-41~1.0\\bin\\Rscript.exe', '--vanilla', '-e', '    options(install.packages.compile.from.source = "never", pkgType = "binary")\n            prefix_dir <- \'C:\\\\Users\\\\USER\\\\.cache\\\\pre-commit\\\\repovawmpj_r\'\n        options(\n            repos = c(CRAN = "https://cran.rstudio.com"),\n            renv.consent = TRUE\n        )\n        source("renv/activate.R")\n        renv::restore()\n        activate_statement <- paste0(\n          \'suppressWarnings({\',\n          \'old <- setwd("\', getwd(), \'"); \',\n          \'source("renv/activate.R"); \',\n          \'setwd(old); \',\n          \'renv::load("\', getwd(), \'");})\'\n        )\n        writeLines(activate_statement, \'activate.R\')\n        is_package <- tryCatch(\n          {\n              path_desc <- file.path(prefix_dir, \'DESCRIPTION\')\n              suppressWarnings(desc <- read.dcf(path_desc))\n              "Package" %in% colnames(desc)\n          },\n          error = function(...) FALSE\n        )\n        if (is_package) {\n            renv::install(prefix_dir)\n        }\n        \n    ')
return code: 1
stdout: (none)
stderr:
    During startup - Warning messages:
    1: Setting LC_COLLATE=en_US.UTF-8 failed 
    2: Setting LC_CTYPE=en_US.UTF-8 failed 
    3: Setting LC_MONETARY=en_US.UTF-8 failed 
    4: Setting LC_TIME=en_US.UTF-8 failed 
    Error in options(install.packages.compile.from.source = never, pkgType = binary) : 
      object 'never' not found
    Execution halted
Check the log at C:\Users\USER\.cache\pre-commit\pre-commit.log
```

The solution described by @asottile in https://github.com/lorenzwalthert/precommit/issues/473#issuecomment-1511498032 is to probably write the contents to a temporary file and avoid unquoting within the expression (i.e. the term after `-e`). This should be quite straight forward.

Question is if we can create a good test first to reproduce the offending behavior and whether or not there are tools already in pre-commit to deal with temp files etc. that we could use.

### pre-commit --version

precommit 3.1.1

### .pre-commit-config.yaml

```yaml
repos:
-   repo: https://github.com/lorenzwalthert/precommit
    rev: v0.3.2.9007
    hooks:
    -   id: style-files
```

_No response_

lorenzwalthert:
Closes #2870. I also fixed a dedention issue in `_inline_r_setup` because if `code` has a different indention than the string we define there. R is indention insensitive but still good to have correct indention.

### DIFF
--- a/pre_commit/languages/r.py
+++ b/pre_commit/languages/r.py
@@ -4,6 +4,8 @@ import contextlib
 import os
 import shlex
 import shutil
+import tempfile
+import textwrap
 from typing import Generator
 from typing import Sequence
 
@@ -19,6 +21,19 @@ ENVIRONMENT_DIR = 'renv'
 RSCRIPT_OPTS = ('--no-save', '--no-restore', '--no-site-file', '--no-environ')
 get_default_version = lang_base.basic_get_default_version
 health_check = lang_base.basic_health_check
+
+
+@contextlib.contextmanager
+def _r_code_in_tempfile(code: str) -> Generator[str, None, None]:
+    """
+    To avoid quoting and escaping issues, avoid `Rscript [options] -e {expr}`
+    but use `Rscript [options] path/to/file_with_expr.R`
+    """
+    with tempfile.TemporaryDirectory() as tmpdir:
+        fname = os.path.join(tmpdir, 'script.R')
+        with open(fname, 'w') as f:
+            f.write(_inline_r_setup(textwrap.dedent(code)))
+        yield fname
 
 
 def get_env_patch(venv: str) -> PatchesT:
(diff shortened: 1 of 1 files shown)


lorenzwalthert:
uff, some windows file issues. Here is what' ChatGPT says:

> on windows, I am writing to a temp file with Python and then, I get 3221225477.

The error code `3221225477` corresponds to the hexadecimal value `0xC0000005`, which is a common Windows error code known as the "Access Violation" or "Segmentation Fault" error. It indicates that your program attempted to access memory that it does not have permission to access.

There can be various reasons why you might encounter this error when writing to a temporary file in Python on Windows. Here are a few possible causes and solutions:

1. Insufficient permissions: Ensure that the user running the Python script has appropriate permissions to write to the specified temporary directory. Make sure the directory exists and the user has write access to it.

2. Antivirus or security software interference: Some antivirus or security software might restrict write access to certain locations, including temporary directories. Temporarily disabling such software or configuring it to allow the Python script to write to the temporary directory could resolve the issue.

3. File handle leakage: If you repeatedly open and write to temporary files without properly closing them, it can result in too many open file handles, causing resource exhaustion. Make sure to close the file handle explicitly using the `close()` method or by utilizing the `with` statement, as you have shown in your code snippet.

4. File path issues: Double-check that the path you are using for the temporary file is correct and valid. If the path contains any special characters or non-standard characters, it might cause issues.

If none of these solutions resolve the issue, please provide more details about your code and the specific steps you are taking that lead to the error. Additionally, if there are any relevant error messages or stack traces accompanying the error code, including those would be helpful in further diagnosing the problem.

asottile:
it looks like R is segfaulting on windows -- no idea why -- there's no useful output to debug it.  I think it might work better with a TemporaryDirectory so there aren't two things holding file handles?  going to try that

asottile (approved):

lorenzwalthert:
Perfect thank you.

Issue #2795, opened 2023-03-03
Pre-commit causing the index to report deleted files when core.fsmonitor enabled


jarshwah:
### search you tried in the issue tracker

fsmonitor stashed

### describe your issue

After committing a file, when pre-commit runs, it reports thousands of files as deleted and then stashes and unstashes them.

This has happened to several members of our team in the last 24 hours.

It only happens when git is configured to run with fsmonitor:

```
git config core.fsmonitor true
```

Git version: git version 2.39.2 (also reproducible with 2.39.1)

This isn't caused by pre-commit, but is a PSA for others using pre-commit and can't figure out why this might be happening.

When the issue is presenting you can get git to show that thousands of files are being deleted by running:

```
git diff-index --ignore-submodules --exit-code --no-color --no-ext-diff $(git write-tree) --
```

Which is what pre-commit effectively runs when checking for unstaged files: https://github.com/pre-commit/pre-commit/blob/2822de9aa6284f2de1c5ff8d0884b38bc553afa5/pre_commit/staged_files_only.py#L52-L56

I'm unsure if  running `git update-index` before `git write-tree` helps, [but the docs seem to suggest](https://git-scm.com/docs/git-write-tree) that the index should always be updated before using write-tree:

> In order to have that match what is actually in your directory right now, you need to have done a git update-index phase before you did the git write-tree

Disabling fsmonitor resolves the issue. Passing `--no-fsmonitor` also seems to help but I haven't re-confirmed this step.

## Reproducing the issue

```
$ git config core.fsmonitor true
$ git checkout -b test-branch
Switched to a new branch 'test-branch'

$ echo "test 3" >> newfile.txt 
$ git add . 
$ git status
On branch test-branch
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	new file:   newfile.txt

$ git commit -a -m "will it fail?"
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/josh/.cache/pre-commit/patch1677888420-10129.
trim trailing whitespace.................................................Passed
check python ast.....................................(no files to check)Skipped
check for case conflicts.................................................Passed
[INFO] Restored changes from /Users/josh/.cache/pre-commit/patch1677888420-10129.
[test-branch 7c899582c19] will it fail?
 1 file changed, 1 insertion(+)
 create mode 100644 newfile.txt

$ git status
On branch test-branch
Changes not staged for commit:
[...]

PR #2890, opened 2023-05-18
Work around a bug in fsmonitor daemon by stopping it

LenaWil:
#2795 is still a thing and [someone claimed](https://lore.kernel.org/git/8afb1988-df49-d759-ead8-1bd384f09c82@gmx.de/) that commands like `git update-index --[really]-refresh` should work. 

However, they didn’t, so I am resorting to temporally stopping the daemon all together to work around it. If a future version fixes it, it’s possible re-enable the daemon again. Worse perfomance is better than deleting random files.

### DIFF
--- a/pre_commit/staged_files_only.py
+++ b/pre_commit/staged_files_only.py
@@ -49,6 +49,8 @@ def _intent_to_add_cleared() -> Generator[None, None, None]:
 
 @contextlib.contextmanager
 def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
+    # Work around a bug in fsmonitor daemon by stopping it.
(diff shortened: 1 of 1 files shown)


asottile:
we can't really make this decision for our users. this would be very surprising

LenaWil:
Thanks for the quick reply.

> we can't really make this decision for our users. this would be very surprising

But so is randomly deleting files! This daemon only caches files, so nothing is permanently lost by stopping it. I will look future into it and see if I can send if I can fill something upstream, but would you accept a pull request with a warning when it’s running? Or `git update-index --no-fsmonitor` which does work around the bug but doesn’t stop the daemon?

PR #2891, opened 2023-05-18
Warn if the fsmonitor-daemon is running

LenaWil:
Less extreme than #2890.

### DIFF
--- a/pre_commit/staged_files_only.py
+++ b/pre_commit/staged_files_only.py
@@ -49,6 +49,8 @@ def _intent_to_add_cleared() -> Generator[None, None, None]:
 
 @contextlib.contextmanager
 def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
+    # Work around a bug in fsmonitor daemon by stopping it.
+    cmd_output('git', 'fsmonitor--daemon', 'stop')
     tree = cmd_output('git', 'write-tree')[1].strip()
     diff_cmd = (
         'git', 'diff-index', '--ignore-submodules', '--binary',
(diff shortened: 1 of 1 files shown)


pre_commit/staged_files_only.py
```diff
@@ -49,6 +49,9 @@ def _intent_to_add_cleared() -> Generator[None, None, None]:
 
 @contextlib.contextmanager
 def _unstaged_changes_cleared(patch_dir: str) -> Generator[None, None, None]:
+    if cmd_output('git', 'fsmonitor--daemon', 'status')[1].startswith('fsmonitor-daemon is watching'):
+        logger.warning('The fsmonitor-daemon is running; a bug in the fsmonitor-daemon sometimes causes data loss when run with pre-commit.')
```
asottile, line 53: this will break on old versions of git

this will also add this subprocess overhead to every execution -- ~20ms on my machine

LenaWil: > this will break on old versions of git

Good point, switched it to checking for the return code.

> this will also add this subprocess overhead to every execution -- ~20ms on my machine

I added two extra checks, a platform check, since it’s only available on Windows & macOS, and a check for the ipc file that should [by default](https://github.com/git/git/blob/main/compat/fsmonitor/fsm-ipc-win32.c) be there, although it’s [seems](https://github.com/git/git/blob/main/compat/fsmonitor/fsm-ipc-darwin.c) to be possible to change it on macOS. Does this speed it up enough?

LenaWil:
@asottile Is this okay? Do I need to write tests?

asottile:
thanks for the PR but I don't think I'm ok with adding this overhead to every command -- I also think that fsmonitor might've fixed some of their issues since then?

LenaWil:
Thanks for finally answering.

PR #2913, opened 2023-06-26
Remove trailing dot after repo URL

eumiro:
When running on GitHub, pre-commit prints log entries with dots at the end of line (as period):

```
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
```

GitHub interprets them as part of URL and breaks the links. I'd suggest removing these trailing dots.


asottile:
there's an exact duplicate PR

PR #2932, opened 2023-07-18
Add haskell language support to pre-commit.

alunduil:
I'm attempting to follow the other languages and documentation in CONTRIBUTING (which seems outdated) to add Haskell support to pre-commit so I can add hooks for several useful Haskell tools.

I'm running into issues when trying to run tests of my new language, and I'm trying to figure out where I'm going wrong. Can someone provide guidance?

I'm assuming the system has Haskell installed already, the same as Python and attempting to follow the documented interface in CONTRIBUTING as well as I can without adding too many things for a first pass. I'm hitting strange errors where my mock project can be listed, but it seems to be something other than the CWD for the installation (no project file found to do the install of '.').

Please let me know if there is more up-to-date documentation or something I should try. Once I pass the tests, I'll be ready for deeper integration with some tools to see how things go.

### DIFF
--- a/pre_commit/languages/haskell.py
+++ b/pre_commit/languages/haskell.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+import contextlib
+import os.path
+from typing import Generator
+from typing import Sequence
+
+from pre_commit import lang_base
+from pre_commit.envcontext import envcontext
+from pre_commit.envcontext import PatchesT
+from pre_commit.envcontext import Var
+from pre_commit.errors import FatalError
+from pre_commit.prefix import Prefix
+
+ENVIRONMENT_DIR = 'hs_env'
+get_default_version = lang_base.basic_get_default_version
+health_check = lang_base.basic_health_check
+run_hook = lang_base.basic_run_hook
+
+
+def get_env_patch(target_dir: str) -> PatchesT:
+    bin_path = os.path.join(target_dir, 'bin')
+    return (('PATH', (bin_path, os.pathsep, Var('PATH'))),)
+
+
+@contextlib.contextmanager
+def in_env(prefix: Prefix, version: str) -> Generator[None, None, None]:
+    envdir = lang_base.environment_dir(prefix, ENVIRONMENT_DIR, version)
+    with envcontext(get_env_patch(envdir)):
+        yield
+
+
+def install_environment(
+    prefix: Prefix,
+    version: str,
+    additional_dependencies: Sequence[str],
+) -> None:
+    lang_base.assert_version_default('haskell', version)
+    envdir = lang_base.environment_dir(prefix, ENVIRONMENT_DIR, version)
+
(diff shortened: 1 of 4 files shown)


pre_commit/languages/haskell.py
```diff
@@ -0,0 +1,74 @@
...
+                additional_dependencies,
+                into=environment_dir,
+            ),
+            check=True,
+        )
+
+
+def _cabal_install_command(
+    packages: typing.Iterable[str],
+    into: pathlib.Path | None = None,
```
asottile, line 69: this should not be optional -- it should *always* be installing into a target directory

pre_commit/languages/haskell.py
```diff
@@ -0,0 +1,74 @@
...
+    _LOGGER.debug(f'environment_dir: {environment_dir}')
+    print(f'pathlib.Path.cwd(): {pathlib.Path.cwd()}')
+    _LOGGER.debug(f'pathlib.Path.cwd(): {pathlib.Path.cwd()}')
+
+    cabal_files = [str(path) for path in pathlib.Path.cwd().glob('*.cabal')]
+    assert len(cabal_files) == 1, (
+        f"didn't find a unique cabal file: {', '.join(cabal_files)}"
+    )
+
+    subprocess.run(
```
asottile, line 53: there shouldn't be any usage of `subprocess` directly -- these should go through `setup_cmd` so they have the proper location -- this is probably most of your issues with getting this to work

pre_commit/languages/haskell.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import annotations
+
+import contextlib
+import logging
+import pathlib
+import subprocess
+import typing
```
asottile, line 7: names should be from-imported from `typing`

tests/languages/haskell_test.py
```diff
@@ -0,0 +1,58 @@
...
+            'init',
+            '--non-interactive',
+            '--quiet',
+            '--minimal',
+            '--package-name=example',
+            '--simple',
+        ],
+        cwd=tmp_path,
+        check=True,
+    )
```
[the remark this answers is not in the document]
alunduil: Yep, that was my next planned step due to the differences between CI and my local runs.

pre_commit/languages/haskell.py
alunduil: That sounds like the winner.  I'll give that a shot and see what happens.  Thanks.

alunduil:
@asottile, thanks for the pointers. Let me know if you have any other stylistic or architectural guidance. This is now passing, and I am confident I do what I need to start playing with other repositories to add hooks for this language.

asottile:
I reworked it a little bit, picked a smaller test dependency, and updated it to match the structure of the rest of the languages and language tests

asottile (approved):

PR #2940, opened 2023-07-24
Create Python virtual environments with `--system-site-packages`

sedrubal:
#### My use case

I have an ansible repository which I want to lint using `ansible-lint` in `pre-commit`. Ansible is installed system wide but `ansible-lint` is installed in the virtual environment of `pre-commit` without access to the system site packages. Thus `ansible-lint` fails to resolve some ansible modules.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -203,7 +203,7 @@ def install_environment(
         additional_dependencies: Sequence[str],
 ) -> None:
     envdir = lang_base.environment_dir(prefix, ENVIRONMENT_DIR, version)
-    venv_cmd = [sys.executable, '-mvirtualenv', envdir]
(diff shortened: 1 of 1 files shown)


asottile:
the whole point is that pre-commit provisions and maintains the whole environment in an isolated manner

sedrubal:
But why? The virtual environment ensures that all packages are available in the required version and other system packages can also be loaded. I don't see any problem here.

Do you have any other solution how I can configure `ansible-lint` to be able to resolve system site packages?

asottile:
the point of pre-commit is any contributor can clone your repo, run` pre-commit install` and all the linting tools work.  `--system-site-packages` means whatever (often messed up) state of your system is involved and not reproducible.

if you need specific dependencies, that's what `additional_dependencies` is for

PR #2943, opened 2023-07-28
Fix for cases where commands output UTF8 in Windows.

shaperilio:
In Windows, `subprocess.PIPE` seems to default to the 1252 codepage, so a hook which outputs UTF8 characters will fail.

The solution [I found](https://stackoverflow.com/a/74607949/149506) is to set the `PYTHONIOENCODING` environment variable.

I've written a unit test which fails in windows if that variable is not set.


asottile:
this is not correct. there's also a duplicate please search the issue tracker

shaperilio:
lol

Issue #2970, opened 2023-08-21
pre-commit treats hook commands dying due to POSIX signal as successful


chriskuehl:
### search you tried in the issue tracker

signal

### describe your issue

When a hook command dies due to an unhandled signal, pre-commit treats the hook as successful.

You can see this for example with a node process that crashes due to hitting a memory limit (it signals itself with SIGTRACE):

```
$ pre-commit run --all-files --verbose node-crash
node-crash...............................................................
Passed
- hook id: node-crash
- duration: 0.02s

<--- Last few GCs --->


<--- JS stacktrace --->


#
# Fatal javascript OOM in GC during deserialization
#

$ echo $?
0
```

I've also included a simple `kill-with-term` hook in the attached config that reproduces it using only bash.

This happens because [`CompletedProcess.returncode` will be a negative integer](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.returncode) (e.g. `-15` for SIGTERM) but pre-commit's returncode merging uses `max()` to collate the returncodes from the processes it spawns:
https://github.com/pre-commit/pre-commit/blob/a1f1d1915646865be2fe84d04633ba964feb0ba0/pre_commit/xargs.py#L172-L173

A draft fix like this resolves the issue for me and correctly shows the hooks as failing:

```diff
--- venv/lib/python3.10/site-packages/pre_commit/xargs.py.orig  2023-08-21 16:19:25.351609037 -0700
+++ venv/lib/python3.10/site-packages/pre_commit/xargs.py       2023-08-21 16:19:31.183503472 -0700
@@ -170,7 +170,7 @@
         results = thread_map(run_cmd_partition, partitions)

         for proc_retcode, proc_out, _ in results:
-            retcode = max(retcode, proc_retcode)
+            retcode = retcode or proc_retcode
             stdout += proc_out

     return retcode, stdout
```

### pre-commit --version

pre-commit 3.3.3

### .pre-commit-config.yaml

```yaml
repos:
-   repo: local
    hooks:
    -   id: kill-with-term
        name: kill-with-term
        entry: bash -c 'kill -TERM $$; while :; do :; done'
        language: system
        always_run: true
    -   id: node-crash
        name: node-crash
        entry: node --max-old-space-size=1 -p "console.log('ohai')"
        language: system
        always_run: true
```

asottile:
yep makes sense, `max(0, -9)` is going to be `0` -- good catch!

PR #2971, opened 2023-08-22
Fix exit code for commands terminated by signals

chriskuehl:
Fixes https://github.com/pre-commit/pre-commit/issues/2970

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -170,7 +170,8 @@ def xargs(
         results = thread_map(run_cmd_partition, partitions)
 
         for proc_retcode, proc_out, _ in results:
-            retcode = max(retcode, proc_retcode)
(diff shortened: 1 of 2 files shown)


pre_commit/xargs.py
```diff
@@ -170,7 +170,8 @@ def run_cmd_partition(
         results = thread_map(run_cmd_partition, partitions)
 
         for proc_retcode, proc_out, _ in results:
-            retcode = max(retcode, proc_retcode)
+            if abs(proc_retcode) > abs(retcode):
+                retcode = proc_retcode
```
chriskuehl, line 174: I used this over something like `retcode = retcode or proc_retcode` just so that it would consistently pick the same exit status from a set of exit statuses. If you prefer some other way of selecting which to use, I can change it to that.

asottile (approved):

Issue #2978, opened 2023-08-28
CPU count logic does not distinguish "usable" CPUs inside cgroup'd containers (e.g. inside k8s)


jdb8:
### search you tried in the issue tracker

multiprocessing, k8s, usable, cpu

### describe your issue

When invoking pre-commit from within a container environment with cgroup cpu enforcement, pre-commit fans out more than expected (based on the total number of CPUs rather than the number of _usable_ CPUs). This ends up causing unexpected behaviour and makes it hard to optimise for performance when running hooks in CI inside containers (since [pre-commit's parallelism is non-configurable](https://github.com/pre-commit/pre-commit/issues/1710)), which leaves us with either "shard using too many CPUs" or `require_serial: true`, both of which aren't ideal.

The cause is seemingly due to the implementation in xargs.py: https://github.com/pre-commit/pre-commit/blob/bde292b51078357384602ebe6b9b27b1906987e5/pre_commit/xargs.py#L28 which uses `multiprocessing.cpu_count()`, which is documented to [not [be] equivalent to the number of CPUs the current process can use](https://docs.python.org/3/library/multiprocessing.html#:~:text=This%20number%20is%20not%20equivalent%20to%20the%20number%20of%20CPUs%20the%20current%20process%20can%20use.%20The%20number%20of%20usable%20CPUs%20can%20be%20obtained%20with%20len(os.sched_getaffinity(0))). This is confirmed by running `python3 -c 'import multiprocessing; print(multiprocessing.cpu_count())'` inside the container environment: the number is higher than expected.

From the docs, it looks like a cgroup-compatible way of grabbing the number of usable CPUs would be to use `len(os.sched_getaffinity(0))`, but I don't know if that has undesirable downsides. I also don't know if this would be a disruptive breaking change to anyone relying on the old behaviour, so I wanted to make this issue first to get your thoughts.

### pre-commit --version

pre-commit 2.19.0

### .pre-commit-config.yaml

```yaml
n/a - can provide a dummy config invoking `python3 -c 'import multiprocessing; print(multiprocessing.cpu_count())'` if more info needed, but this isn't specific to any one hook or config option
```

asottile:
sched_getaffinity isn't available on other platforms but could probably be used if available

PR #2979, opened 2023-08-29
Use os.sched_getaffinity for cpu counts when available

jdb8:
Attempt at fixing/improving https://github.com/pre-commit/pre-commit/issues/2978 for systems that support the `os.sched_getaffinity` method.

When unavailable (i.e. on mac), we fall back to the old behaviour.

I didn't know how much to mess with the existing tests so went with a new fixture to simulate when this is available - I don't have a windows machine to test on so let me know if there's a better way to avoid this change being annoying!


tests/lang_base_test.py
```diff
@@ -30,6 +30,23 @@ def fake_expanduser(pth):
         yield
 
 
+@pytest.fixture
+def no_sched_getaffinity():
+    # Create the attribute if it doesn't exist, so we can reliably patch it
+    if not hasattr(os, 'sched_getaffinity'):
+        os.sched_getaffinity = None  # type: ignore
```
asottile, line 37: can probably use `create=not hasattr(os, 'sched_getaffinity')` below

asottile:
pre-commit.ci run

asottile (approved):

deronnax:
shimming in to tell you - in case you don't alreadu know it - that [`os.process_cpu_count`](https://docs.python.org/fr/3/library/os.html#os.process_cpu_count) was introduced in python 3.13 for this purpose.

PR #2994, opened 2023-09-06
Allow hook installation even when global core.hooksPath is set via --allow-global-hooks

lyoung-confluent:
In corporate environments it is common to have global git hooks (such as a `pre-push` check) installed via `core.hooksPath` (set in the global gitconfig). When global hooks are installed per-repository hooks will not be executed so right now `pre-commit` reasonably checks for this condition and refuses to install:
```
$ pre-commit install
[ERROR] Cowardly refusing to install hooks with `core.hooksPath` set.
hint: `git config --unset-all core.hooksPath`
```
However it's possible/common to setup any global hooks to still explicitly execute any per-repository hooks such as via the following:
```bash
if [ -e "$GIT_DIR/.git/hooks/pre-commit" ]; then
  $GIT_DIR/.git/hooks/pre-commit "$@"
fi
```
In this circumstance users may still wish to install per-repository hooks but `pre-commit` will refuse to allow this. This PR permits the `pre-commit install` command even if a global `core.hooksPath` has been configured if the user provides the new (opt-in) `--allow-global-hooks` flag.

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -118,9 +118,14 @@ def install(
         overwrite: bool = False,
         hooks: bool = False,
         skip_on_missing_config: bool = False,
+        allow_global_hooks: bool = False,
         git_dir: str | None = None,
 ) -> int:
-    if git_dir is None and git.has_core_hookpaths_set():
+    if (
+            git_dir is None and
+            git.has_core_hookpaths_set() and
+            not allow_global_hooks
+    ):
         logger.error(
             'Cowardly refusing to install hooks with `core.hooksPath` set.\n'
             'hint: `git config --unset-all core.hooksPath`',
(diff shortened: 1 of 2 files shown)


asottile:
as you can tell from the code this is intentional

jkao97:
@asottile curious, any reason why it's made intentional?

lyoung-confluent:
@asottile Yes I see it's intentional and agree with the logic behind that as a default but is there any reason users shouldn't be able to intentional disable the check when they are aware of the behavior/prerequisites?

asottile:
you're free to maintain your own fork or use `init-templatedir` but I'm not going to actively maintain and support a footgun for free

in general it is best to search the issue tracker and start a discussion before dumping a PR as well -- you would have found several discussions indicating that this is not going to be accepted and saved us all the trouble

I'd also recommend your company sponsor if you're deriving value from my free labor

PR #2999, opened 2023-09-10
Fix mypy on MacOS

mxr:
Without this change, `mypy` fails with

```
pre_commit/xargs.py:31: error: Module has no attribute "sched_getaffinity"  [attr-defined]
```

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -24,13 +24,11 @@ TRet = TypeVar('TRet')
 
 
 def cpu_count() -> int:
-    try:
-        # On systems that support it, this will return a more accurate count of
-        # usable CPUs for the current process, which will take into account
-        # cgroup limits
+    # On systems that support it, this will return a more accurate count of
+    # usable CPUs for the current process, which will take into account
+    # cgroup limits
+    if hasattr(os, 'sched_getaffinity'):
         return len(os.sched_getaffinity(0))
-    except AttributeError:
-        pass
 
     try:
         return multiprocessing.cpu_count()
--- a/tests/xargs_test.py
+++ b/tests/xargs_test.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import concurrent.futures
+import contextlib
 import multiprocessing
 import os
 import sys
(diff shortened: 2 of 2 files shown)


asottile:
I don't use macos, I'd recommend Linux for development

pre_commit/xargs.py
```diff
@@ -24,13 +24,11 @@
...
 def cpu_count() -> int:
-    try:
-        # On systems that support it, this will return a more accurate count of
-        # usable CPUs for the current process, which will take into account
-        # cgroup limits
+    # On systems that support it, this will return a more accurate count of
+    # usable CPUs for the current process, which will take into account
+    # cgroup limits
+    if hasattr(os, 'sched_getaffinity'):
         return len(os.sched_getaffinity(0))
```
asottile, line 31: this makes this ever-so-slightly worse in that it grabs the attribute twice in the "normal" path

tests/xargs_test.py
```diff
@@ -20,30 +21,30 @@ def test_cpu_count_sched_getaffinity_exists():
...
-            create=True,
-            side_effect=AttributeError,
-    ):
+    if hasattr(os, 'sched_getaffinity'):  # pragma: no cover (win32 py38)
+        func = os.sched_getaffinity
+        del os.sched_getaffinity
         yield
+        os.sched_getaffinity = func
+    else:
+        yield  # pragma: no cover (darwin)
```
asottile, line 34: you're looking for `# pragma: posix cover` and `# pragma: posix no cover` which uses `covdefaults` to mark these appropriately

asottile (commented):
left a review anyway -- but I don't support type checking on other platforms -- windows also doesn't "pass" mypy due to the platform-specific code iirc

tests/xargs_test.py
mxr: TIL these tokens have semantic meaning

mxr: TIL

asottile: it would need try + finally

mxr: Oh I meant TIL that purest fixtures are exception-safe. I realize without try/finally my code isn't exception-safe

PR #3011, opened 2023-09-25
Bump Node.js version to 18.14.0 and Go to 1.21.1

hack3ric:
On riscv64, nodeenv will pull binary from unofficial-builds [1], and unfortunately 18.13.0 seems to be the only version above 18 that is missing riscv64 builds. Shifting the version slightly to make test work.

Go's binary now ships with linux/riscv64 binary since 1.21.

### DIFF
--- a/tests/languages/golang_test.py
+++ b/tests/languages/golang_test.py
@@ -111,11 +111,11 @@ def test_golang_versioned(tmp_path):
         tmp_path,
         golang,
         'go version',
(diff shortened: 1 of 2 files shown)


asottile:
the `[1]` doesn't go anywhere

hack3ric:
> the `[1]` doesn't go anywhere

My bad. https://unofficial-builds.nodejs.org/download/release/

Issue #1453, opened 2020-05-12
additional_dependencies for system language hooks


marcjay:
I'm aware of the requirement that additional dependencies for system hooks - the 'escape hatch' for hooks that don't fit into the python/node/go etc. managed environments list, such as shellcheck, must be installed outside of pre-commit in order to work.

Also aware of the python module they can be used to 'mirror' shellcheck in particular, though that feels a little hacky.

I'm sure you've given this thought before, and there's a very good reason why this hasn't been done, but, is it conceivable, with a PR I'd be willing to write, that additional dependencies could be supported for non-python etc. hooks?

I'm thinking along the lines of being able to specify a URL to a ZIP, which pre-commit would `wget` into its cache directory, sorting a checksum so it could 'manage' when to re-download it, and then adding the directory to the front of `$PATH` when running the hook.

I'm sure it's not possible, but would be keen to know why, as there are a few hooks I use that this applies too and would feel like a useful addition for others.

asottile:
I don't think that would be a feature to `language: system` -- I think that would be a separate `language` itself (maybe `language: download` or something).  in the ideal world, pre-commit would learn how to install environments of all major programming languages and then you'd be able to install the tool you're looking for in that manner.  there's [directions for adding a new language here](https://github.com/pre-commit/pre-commit/blob/master/CONTRIBUTING.md#adding-support-for-a-new-hook-language) and several languages have been contributed by external contributors already!

also `python` is not the only language which supports `additional_dependencies`, `go`, `ruby`, `node`, and a few others also support the setting (and presumably languages which are added can also support that as well).

for shellcheck in particular, [shellcheck-py](https://github.com/shellcheck-py/shellcheck-py) is the suggested tool for pre-commit

note that a url alone does not help you here since you need to work about platform and architecture and likely checksumming, etc.

for `system` / `script` hooks in particular, they do not get any "environment" to install things into (they're run directly from configuration-only)

marcjay:
> I think that would be a separate `language` itself (maybe `language: download` or something)

That's a good point, yup

> in the ideal world, pre-commit would learn how to install environments of all major programming languages and then you'd be able to install the tool you're looking for in that manner.

[...]

PR #3020, opened 2023-10-06
Add a 'download' language for fetching pre-built binaries

fangfufu:
This commit implements https://github.com/pre-commit/pre-commit/issues/1453#issuecomment-1702880253

### DIFF
--- a/pre_commit/all_languages.py
+++ b/pre_commit/all_languages.py
@@ -7,6 +7,7 @@ from pre_commit.languages import dart
 from pre_commit.languages import docker
 from pre_commit.languages import docker_image
 from pre_commit.languages import dotnet
+from pre_commit.languages import download
 from pre_commit.languages import fail
 from pre_commit.languages import golang
 from pre_commit.languages import haskell
@@ -30,6 +31,7 @@ languages: dict[str, Language] = {
     'docker': docker,
     'docker_image': docker_image,
     'dotnet': dotnet,
+    'download': download,
     'fail': fail,
     'golang': golang,
     'haskell': haskell,
--- a/pre_commit/languages/download.py
+++ b/pre_commit/languages/download.py
@@ -1,0 +1,328 @@
+from __future__ import annotations
+
+import contextlib
+import hashlib
+import os.path
+import platform
+import stat
+from base64 import standard_b64decode as b64decode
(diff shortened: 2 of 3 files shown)


asottile:
please no dataclasses or backslashes -- really you probably don't need any classes at all.  I'm also not sure why the strange sha256 -- use hexdigest which is way more standardized.

there's a lot of code here and almost zero tests so it's not in a reviewable state -- please convert your PR to draft until you're ready for a first round

fangfufu:
What do you mean by 
> you probably don't need any classes at all

The " the strange sha256" is [subresource integrity](https://www.w3.org/TR/SRI/#integrity-metadata).

What do you mean by "no backslashes"?

asottile:
> What do you mean by
> 
> > you probably don't need any classes at all
> 

your code has a bunch of classes with nontrivial methods -- simple plain-old-data classes is preferred -- check the approaches used by most of the rest of the codebase

> The " the strange sha256" is [subresource integrity](https://www.w3.org/TR/SRI/#integrity-metadata).

just because it is standardized doesn't mean it is good -- we aren't optimizing for http headers where characters are at a minimum so use normal hexdigest which is far more common and much easier to compute
 
> What do you mean by "no backslashes"?

your source code contains many very very ugly backslashes

fangfufu:
> so use normal hexdigest which is far more common

@asottile, normal checksum, i.e. the output from ``sha256sum`` or ``md5sum`` are not in hex. If we use hexdigest here, then you also need something to generate those hexdigest, right?

asottile:
> > so use normal hexdigest which is far more common
> 
> @asottile, normal checksum, i.e. the output from `sha256sum` or `md5sum` are not in hex. If we use hexdigest here, then you also need something to generate those hexdigest, right?

what do you mean yes they are

fangfufu:
Ah never mind, for some reason I thought ``md5sum`` and ``sha256sum`` outputs string in base64 format.

fangfufu:
@asottile , do you have a preferred way of handling the different OS and architecture?

asottile:
setuptools-download handles that with `markers` -- essentially filters on `sys.platform`, `platform.machine()` and `os.name` (though `os.name` is perhaps redundant)

DrummyFloyd:
any update regarding this ?

PR #3029, opened 2023-10-13
Improve hook duration timing

adamchainz:
Use `time.monotonic()` to prevent clock slippage error ~, and always output two significant figures to avoid ambiguity~.

### DIFF
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -293,7 +293,7 @@ def test_verbose_duration(cap_out, store, in_git_dir, t1, t2, expected):
     write_config('.', {'repo': 'meta', 'hooks': [{'id': 'identity'}]})
     cmd_output('git', 'add', '.')
     opts = run_opts(verbose=True)
-    with mock.patch.object(time, 'time', side_effect=(t1, t2)):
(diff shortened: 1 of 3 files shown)


tests/commands/run_test.py
```diff
@@ -286,14 +286,14 @@ def test_global_files(cap_out, store, in_git_dir):
     ('t1', 't2', 'expected'),
     (
         (1.234, 2., b'\n- duration: 0.77s\n'),
-        (1., 1., b'\n- duration: 0s\n'),
+        (1., 1., b'\n- duration: 0.00s\n'),
```
asottile, line 289: I prefer the previous display here

adamchainz: For 0s I get it. But if you got 5.00s, it would be reported as 5s, which it wouldn't clear is faster than another measurement of 5.49s.

asottile: that's fine, I don't want that part of this change though

adamchainz: Okay I’ve undone the fixed point.

asottile (approved):

PR #3042, opened 2023-10-28
python3.9+

asottile:
my hands are tied -- if you want to complain do it here: https://github.com/python/cpython/pull/106532


EwoutH:
I don't know if you want to do at this stage, but if you want you could support Python 3.8 again:
- https://github.com/python/cpython/issues/116608

asottile:
I'm not, but it's good to see that I was right all along

PR #3044, opened 2023-10-29
Update hooks version in the example config

kaplanlior:
Update to version v4.5.0


asottile:
this is not meant to be kept up to date. you can use autoupdate

Issue #3059, opened 2023-11-09
Add option exclude_file_path to allow passing a file path with a list of excludes.


cesan3:
### search you tried in the issue tracker

provide option to pass external file with excludes files or patterns

### describe your actual problem

I need to exclude a huge list of paths from several hooks.  I'd like the ability to provide an option with a file containing the files/patterns.

I'll provide PR

### pre-commit --version

pre-commit 3.5.0

asottile:
no thank you -- pre-commit only reads from its configuration file and I'm not changing that

asottile:
fwiw you probably want https://pre-commit.com/#top_level-exclude or to use yaml anchors of some sort

PR #3060, opened 2023-11-09
fix: add ability to provide a file with excludes

cesan3:
This resolves: #3059

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -110,6 +110,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
 
     cfgv.Optional('files', check_string_regex, ''),
     cfgv.Optional('exclude', check_string_regex, '^$'),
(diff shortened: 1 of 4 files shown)


asottile:
feature was never agreed upon -- don't waste time implementing features until they've been agreed upon

cesan3:
Hmm, just a thought, having the ability to extend the configuration using external files could help a lot with integration with other tools and frameworks.

asottile:
no it would not.  it would create a bunch of maintenance work for me for little to no gain

DrMarkEy:
Just my 2 cents: My project already has a CI pipeline set up, which uses files with excludes for some linting jobs.
If I now want to use pre-commit for my local setup, I would have to copy all these excludes to the config file, translate them from glob patterns to regular expressions and from now on maintain both lists in parallel. This is a huge pain and currently a showstopper for me. I would really appreciate this feature a lot!

asottile:
you don't though

pre-commit only runs on checked in files and you can run it in ci

PR #3072, opened 2023-11-18
Add missing help to --verbose param

touilleMan:
Hi again,

I noticed `--verbose` is missing its help ;-)

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -59,7 +59,10 @@ def _add_hook_type_option(parser: argparse.ArgumentParser) -> None:
 
 def _add_run_options(parser: argparse.ArgumentParser) -> None:
     parser.add_argument('hook', nargs='?', help='A single hook-id to run')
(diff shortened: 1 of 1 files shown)


asottile:
it does a little bit more than that but also "verbose" is usually self explanatory

touilleMan:
> it does a little bit more than that but also "verbose"

FYI I took the verbatim text from the [pre-commit hooks documentation](https://pre-commit.com/#pre-commit-configyaml---hooks), so maybe it should be updated then (or removed entirely given it is self explanatory :clown_face: )

image

asottile:
that's a hook configuration not the commandline 🤡

touilleMan:
> that's a hook configuration not the commandline 🤡

My bad, I was [pointing the wrong place](https://pre-commit.com/#pre-commit-run)

image

anyway, I won't be wasting any more of your item

PR #3079, opened 2023-12-01
Address deprecation warning in `shutil.rmtree(onerror=...)`

edgarrmondragon:
<blockquote>
<p><a href="https://docs.python.org/3.12/library/shutil.html#shutil.rmtree"><code>shutil.rmtree()</code></a> now accepts a new argument <em>onexc</em> which is an
error handler like <em>onerror</em> but which expects an exception instance
rather than a <em>(typ, val, tb)</em> triplet. <em>onerror</em> is deprecated and
will be removed in Python 3.14.
(Contributed by Irit Katriel in <a class="reference external" href="https://github.com/python/cpython/issues/102828">gh-102828</a>.)</p>
</blockquote>

https://docs.python.org/3.12/whatsnew/3.12.html#shutil

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,5 +19,5 @@ jobs:
   main-linux:
     uses: asottile/workflows/.github/workflows/tox.yml@v1.6.0
     with:
-      env: '["py39", "py310", "py311"]'
+      env: '["py39", "py310", "py311", "py312"]'
       os: ubuntu-latest
--- a/pre_commit/util.py
+++ b/pre_commit/util.py
@@ -202,24 +202,36 @@ else:  # pragma: no cover
     cmd_output_p = cmd_output_b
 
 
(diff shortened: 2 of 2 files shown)


asottile:
cast should only be used for bugs in the type system

I don't think we need this function any more sooner than the deprecation matters so I'm probably going to decline this anyway

asottile:
> > cast should only be used for bugs in the type system
> > I don't think we need this function any more sooner than the deprecation matters so I'm probably going to decline this anyway
> 
> No problem. And I prefer avoiding `cast` but Mypy was complaining that `"tuple[type[OSError], OSError, TracebackType] | BaseException" is not indexable`.

yeah it's not -- the way you've written it is not typesafe

edgarrmondragon:
FWIW I added a guard to check if the exception is an `OSError` so that allowed to remove the `cast` call.

PR #3083, opened 2023-12-04
Update validate_config.py

msmall318:
I was looking to improve the validate config function. I was initially using without passing any .yaml and it didn't alert me to pass a config.

I do not know where the rest of this functionality is used so second and third order affects were not tested, but I didn't want to continue developing further if it wasn't of interest. Thanks for the package!


msmall318:
It might be breaking because I inserted a raise FileNotFoundError rather than simply using a print statement.

msmall318:
You intentionally want to provide the user no feedback? Interesting strategy.

asottile:
I don't need your attitude

PR #3099, opened 2023-12-13
add docker support for cgroup v2 implementations

niall-byrne:
## The problem:

The current implementation of "_is_in_docker" unfortunately only covers cases where the container engine is running `version 1` of cgroups.  This is only a problem due to `version 2` now becoming more widely adopted.  (ie. Docker Desktop for Mac)

I’d like to make use of pre-commit inside containers for various projects I’m working on but the lack of compatibility with `version 2` makes this infeasible, so I did some reading and then some testing…

### Sources:

- [baeldung.com](https://www.baeldung.com/linux/is-process-running-inside-container)
- [is-docker](https://www.npmjs.com/package/is-docker)
- [toradex.com](https://community.toradex.com/t/python-nullresource-error-when-running-torizoncore-builder-build/15240/3)
- [kubernetes.io](https://kubernetes.io/docs/concepts/architecture/cgroups/)
- [docs.kernel.org](https://docs.kernel.org/admin-guide/cgroup-v2.html)

### Main takeaways on detecting version 2 containerization:

- You cannot identify anything related to containerization by looking at the cgroup entry for process 1 anymore… this is what “_is_in_docker” relies on.
- Although you can OFTEN still get the container ID through network-related files (hosts and hostname for example) that are derived from the host machine (`/proc/self/mountinfo`) ([Source](https://www.baeldung.com/linux/is-process-running-inside-container)) it's not consistent across all docker implementations and container use cases.
- Likewise, although LXC has been removed from docker, there are STILL a lot of folks using the '.dockerenv' file to determine if the current environment is a container- but this is NOT a standard.
 
## PR:

This PR went through two cycles:
1. Very complex, I was parsing the `/proc/self/mountinfo` file to extract the container ID, before doing more testing and realizing it's NOT always available in `version 2` cgroups.  There are exceptions to this.
2. I learned from my mistaken assumption and opted to go an arguably simpler route of exposing an environment variable to allow the user to override the default behaviour.

I have named the variable "PRE_COMMIT_OVERRIDE_DOCKER_HOST_PATH", which is a bit wordy.
Please feel free to request additional changes...

### Example Implementation of PR

```yaml
---
version: "3.8"

services:
  my_python_service:
    build:
      args:
        BUILD_ARG_CONTAINER_UID: "${DOCKER_USER_UID:-1000}"
        BUILD_ARG_CONTAINER_GID: "${DOCKER_USER_UID:-1000}"
        BUILD_ARG_PIP_INDEX_URL: "https://pypi.org/simple"
        BUILD_ARG_PLATFORM: "amd64"
      context: .
      dockerfile: assets/Dockerfile
      target: development
    platform: linux/amd64
    env_file:
      - assets/local.env
    environment:
      ENVIRONMENT: "DEVELOPMENT"
      PRE_COMMIT_OVERRIDE_DOCKER_HOST_PATH: "${PWD}"
    volumes:
      - ./:/app
      - /var/run/docker.sock:/var/run/docker.sock
```

I'm hoping this simpler approach (although potentially falling victim to bad user input) would allow support for a wider range of docker implementations.

Thanks again for all your work supporting this incredibly useful tool.

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -10,6 +10,7 @@ from pre_commit.prefix import Prefix
 from pre_commit.util import CalledProcessError
 from pre_commit.util import cmd_output_b
 
+ENV_VAR_OVERRIDE_HOST_PATH = 'PRE_COMMIT_OVERRIDE_DOCKER_HOST_PATH'
 ENVIRONMENT_DIR = 'docker'
 PRE_COMMIT_LABEL = 'PRE_COMMIT'
 get_default_version = lang_base.basic_get_default_version
@@ -37,6 +38,8 @@ def _get_container_id() -> str:
 
 
 def _get_docker_path(path: str) -> str:
+    if ENV_VAR_OVERRIDE_HOST_PATH in os.environ:
+        return os.environ[ENV_VAR_OVERRIDE_HOST_PATH]
     if not _is_in_docker():
         return path
 
--- a/tests/languages/docker_test.py
+++ b/tests/languages/docker_test.py
@@ -33,6 +33,8 @@ DOCKER_CGROUP_EXAMPLE = b'''\
 # The ID should match the above cgroup example.
 CONTAINER_ID = 'c33988ec7651ebc867cb24755eaf637a6734088bc7eef59d5799293a9e5450f7'  # noqa: E501
 
+HOST_PATH_ENV_VAR_OVERRIDE_EXAMPLE = '/host/path/override'
+
 NON_DOCKER_CGROUP_EXAMPLE = b'''\
 12:perf_event:/
 11:hugetlb:/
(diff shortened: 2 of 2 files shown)


niall-byrne:
It looks like [these](https://github.com/pre-commit/pre-commit/blob/9c9983dba00bf67d1b2625f1f0e9112afc063849/pre_commit/util.py#L229C5-L230C79) lines are not executing with python 3.12.1 (your last release used 3.12.0)- I can replicate locally.

niall-byrne:
For the window build, the only difference I can see is that the version of distlab was incremented.

```
$ diff a b
14c14
< py39: cfgv==3.4.0,colorama==0.4.6,covdefaults==2.3.0,coverage==7.3.2,distlib==0.3.7,exceptiongroup==1.2.0,filelock==3.13.1,identify==2.5.33,iniconfig==2.0.0,nodeenv==1.8.0,packaging==23.2,pip==23.3.1,platformdirs==4.1.0,pluggy==1.3.0,pre-commit @ file:///D:/a/pre-commit/pre-commit/.tox/.tmp/package/1/pre_commit-3.6.0.tar.gz#sha256=a2eb5928471cf15d0da5451c13b46acd439af96914449ff058d2a356eeb3a579,pytest==7.4.3,pytest-env==1.1.3,PyYAML==6.0.1,re-assert==1.1.0,regex==2023.10.3,setuptools==69.0.2,tomli==2.0.1,virtualenv==20.25.0,wheel==0.42.0
---
> py39: cfgv==3.4.0,colorama==0.4.6,covdefaults==2.3.0,coverage==7.3.3,distlib==0.3.8,exceptiongroup==1.2.0,filelock==3.13.1,identify==2.5.33,iniconfig==2.0.0,nodeenv==1.8.0,packaging==23.2,pip==23.3.1,platformdirs==4.1.0,pluggy==1.3.0,pre-commit @ file:///D:/a/pre-commit/pre-commit/.tox/.tmp/package/1/pre_commit-3.6.0.tar.gz#sha256=599c3d617ed3be25663a354006835a954da822fcb94249dd257b6b3789845146,pytest==7.4.3,pytest-env==1.1.3,PyYAML==6.0.1,re-assert==1.1.0,regex==2023.10.3,setuptools==69.0.2,tomli==2.0.1,virtualenv==20.25.0,wheel==0.42.0
16a17
>
```

asottile:
the coverage difference looks like a regression in `coverage` itself -- going to retry the build now that coverage 7.3.4 is out -- if not then we'll need to report there

pre_commit/languages/docker.py
```diff
@@ -10,6 +10,7 @@
 from pre_commit.util import CalledProcessError
 from pre_commit.util import cmd_output_b
 
+ENV_VAR_OVERRIDE_HOST_PATH = 'PRE_COMMIT_OVERRIDE_DOCKER_HOST_PATH'
```
asottile, line 13: this doesn't actually implement what the title of this PR says -- I don't really like sprinkling environment variable hacks so I can't accept this as it is.  I would accept a PR which improves the detection -- but as it is right now this is just a workaround that I'd have to document and maintain forever rather than an actual solution

niall-byrne: That was the intent of my original PR yes. 

Unfortunately, after reading and testing I realized there isn't a consistent way of obtaining the container ID that works on all platforms for version 2 of cgroups...  I opted for the simpler solutions of an env var as that has the benefit on consistent functionality.

PR #3101, opened 2023-12-20
Support GN build language

KindDragon:
GN is a meta-build system that generates build files for Ninja

See also https://gn.googlesource.com/gn/

I create it based on golang language, but simplify it a bit. Main purpose of this language to create hook for `gn format`

### DIFF
--- a/pre_commit/languages/gn.py
+++ b/pre_commit/languages/gn.py
@@ -1,0 +1,101 @@
+from __future__ import annotations
+
+import contextlib
+import functools
+import os.path
+import platform
+import shutil
+import tempfile
+import urllib.error
+import urllib.request
+import zipfile
+from collections.abc import Generator
+from collections.abc import Sequence
+from typing import ContextManager
+from typing import IO
+
+import pre_commit.constants as C
+from pre_commit import lang_base
+from pre_commit.envcontext import envcontext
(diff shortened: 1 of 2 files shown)


asottile:
if your goal is to just make a hook for `gn` -- it probably doesn't make sense as a language

you can build something similar to https://github.com/shellcheck-py/shellcheck-py -- basically you'd be writing a small python package around it and use [setuptools-download](https://github.com/asottile/setuptools-download) to ensure the binary is available

optionally you can publish that to pypi as well -- that'd make it super simple to `pip install gn-py` as well!

KindDragon:
BTW why you not allowing `language: docker`, or `language: docker_image` hooks at [pre-commit.com](https://github.com/pre-commit/pre-commit.com/tree/main)
/all-repos.yaml?

asottile:
native installation provides a better user experience. installation at runtime is unsupported and can corrupt the cache

KindDragon:
Ok, thanks. But I think it's better in the future Implement for something like `entry_install` and `entry_uninstall` for `language: script` to reduce such unnecessary python packages when you just need to download binaries

asottile:
if you search the issue tracker you'll find such a suggestion already

PR #3113, opened 2024-01-18
add check on pre-push stage: exit if staged files not committed yet

### LINKED ISSUE #2486: dotnet format pre-push hook allow push when correct file is in staged changes

yes

### describe your issue

I configured **dotnet format** as push hook in my **.pre-commit-config.yaml**
```
default_stages: [commit]
repos:
    rev: v4.1.0
    hooks:
      - id: check-added-large-files
      - id: check-case-conflict
  - repo: local
    hooks:
    - id: dotnet-format
      name: dotnet-format
      language: system
      entry: dotnet format --include
      types_or: ["c#"]
      stages : [push]
```
I installed them using
```
pre-commit install --install-hooks --hook-type pre-commit --hook-type pre-push
```
I committed a modification with dotnet format not following correct rules, it passes
I try to push, it is rejected with error
```
 error: failed to push some refs to
```
And it corrects my file after dotnet format

When I try to push again, I have the same but when I add the file in staged changes without commit it, I can push.

### pre-commit --version

pre-commit 2.20.0

### .pre-commit-config.yaml

```yaml
default_stages: [commit]
repos:
    rev: v4.1.0
    hooks:
      - id: check-added-large-files
      - id: check-case-conflict
  - repo: local
    hooks:
    - id: dotnet-format
      name: dotnet-format
      language: system
      entry: dotnet format --include
      types_or: ["c#"]
      stages : [push]
```

_No response_

aless10:
This PR aims to solve the issue #2486. The idea is to follow the same solution used for #1418 and add the check on the hook name and if there are staged files not committed yet.

**About the error message**
I did not want to print the list of the staged files, because it can too long and not very useful. I used `logger.error` to print the message, following what is done in the other checks above. 

**Alternatives**:
- instead of using the function `git.get_staged_files`, we can just check for the number of staged files and fail if it is > 0, something like `git diff --cached --numstat | wc -l`

fixes #2486


tests/commands/run_test.py
```diff
@@ -1217,3 +1222,36 @@ def test_pre_commit_env_variable_set(cap_out, store, repo_with_passing_hook):
         cap_out, store, repo_with_passing_hook, args, environ,
     )
     assert environ['PRE_COMMIT'] == '1'
+
+
+def test_pre_push_fails_if_staged_files(
+        cap_out, store, repo_with_passing_hook,
+):
```
asottile, line 1229: this test is probably sufficient -- then you don't need the additional one above

pre_commit/commands/run.py
```diff
@@ -370,7 +370,14 @@ def run(
             environ.get('_PRE_COMMIT_SKIP_POST_CHECKOUT')
     ):
         return 0
-
+    # prevent pushing staged files not committed (#2486)
+    if (
+            args.hook_stage == 'pre-push' and git.get_staged_files()
+    ):
```
asottile, line 376: the extra parens here do nothing

pre_commit/commands/run.py
```diff
@@ -370,7 +370,14 @@ def run(
...
+        logger.error(
+            'Staged files found. Please commit before pushing',
+        )
```
asottile, line 379: this easily fits on one line

.gitignore
```diff
@@ -4,3 +4,4 @@
 /.tox
 /dist
 .vscode/
+.idea
```
asottile, line 7: please don't touch gitignore files in projects you don't own -- please rebase this out thanks!

PR #3125, opened 2024-02-08
Fixes xargs partition length on Windows for EXE hooks

### LINKED ISSUE #3124: In Windows, `pre-commit -a` silently skips files as xargs.py partitions are too long for command line

Windows all-files ignored pretty_format_json

### describe your issue

I have a large Git repo with about 6k files, including 1.7k JSON files. Some relative filepaths are long (256 characters). I tried to lint all the JSON files using

```sh
pre-commit run pretty-format-json --all-files
```

but I noticed that it only modified about a third of the JSON files. I thought it was an issue with Windows programs sometimes not supporting long filepaths (UTC). I searched the issues and thought this was related to #589 and opened #1007. I then realized that I could lint the individual files, e.g.

```sh
pre-commit run pretty-format-json --files="my/really/long/filepath/file.json"
```

so thought it wasn't an issue with long filepaths.

I used a debugger to confirm that all of the 1.7k JSON filepaths are passed as `varargs` to `pre_commit.xargs.xargs()`. In there, there is a partitioner that should shorten the commands to Windows' 8192 character limit (developed in #1686 to resolve #1604). Unfortunately, that only works if the hook is a `.bat` or `.cmd` file, whereas the `pretty_format_json` hook is a `.exe`. I think either:

1. the `pre_commit.xargs.xargs()` partition logic should be expanded also to check if the hook is `.exe`
2. or simplify the partition logic just to check `sys.platform == 'win32'`

I tried both options separately, and either would resolve my issue (all JSONs linted).

### pre-commit --version

3.2.0

### .pre-commit-config.yaml

```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.5.0
    hooks:
    -   id: pretty-format-json
        args: ['--autofix', '--indent=2']
```

_No response_

amarvin:
Closes #3124

### DIFF
--- a/pre_commit/xargs.py
+++ b/pre_commit/xargs.py
@@ -155,7 +155,7 @@ def xargs(
     # on windows, batch files have a separate length limit than windows itself
     if (
             sys.platform == 'win32' and
-            cmd[0].lower().endswith(('.bat', '.cmd'))
+            cmd[0].lower().endswith(('.bat', '.cmd', '.exe'))
     ):  # pragma: win32 cover
         # this is implementation details but the command gets translated into
         # full/path/to/cmd.exe /c *cmd
--- a/tests/xargs_test.py
+++ b/tests/xargs_test.py
@@ -243,7 +243,7 @@ def test_xargs_color_true_makes_tty():
 
 
 @pytest.mark.xfail(os.name == 'posix', reason='nt only')
-@pytest.mark.parametrize('filename', ('t.bat', 't.cmd', 'T.CMD'))
+@pytest.mark.parametrize('filename', ('t.bat', 't.cmd', 'T.CMD', 't.exe'))
 def test_xargs_with_batch_files(tmpdir, filename):
     f = tmpdir.join(filename)
     f.write('echo it works\n')


asottile:
I don't think this is the right fix or really the problem

I've verified the length computation is correct on my windows machine so something is different about yours (and it's not this)

you'll need to do some more debugging on your end too figure out what's happening

and usually when the length is exceeded there's an error.  if that's not happening there's likely something else at play!

amarvin:
Can you confirm that in your system the `pretty-format-json` hook in `pre_commit.xargs.xargs()` gets shebang normalized to a `.bat` or `.cmd` file? For me:

```py
> cmd[0]
'C:\\Users\\alexm\\.cache\\pre-commit\\repodli143ao\\py_env-python3.12\\Scripts\\pretty-format-json.EXE'
```

so then `_max_length` stays at the default of 30720, which is too long for Windows batch file lines.

asottile:
batch is not in play on my system, no

amarvin:
Huh? All I'm saying is that on my system Windows 11, this if-statement is never entered:

image

amarvin:
The length of the first command sent to `cmd_fn` was 15440 characters. I can't even paste that long of a string into `cmd.exe`.

amarvin:
You're right that neither batch files nor `cmd.exe` are being used in this case / on my system by the `subprocess.Popen()` in `pre_commit.util`. Not sure where this differing behavior is coming or why shorting the partitioning max length fixes it.

asottile:
yes that's why I'm saying you need to do some debugging to figure out why.  I'm not blindly changing the code

PR #3146, opened 2024-03-08
Advise user to unset `core.hooksPath` git config globally as well as locally

wircho:
This is a very simple PR that should not need a test of an issue (but feel free to advise otherwise).

Global `core.hooksPath` configurations are rare but not unheard of. I just happened to work on a project that set this config globally on my machine and this broke my ability to install pre-commit. This simple change would have saved me a bit of time.

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -123,7 +123,7 @@ def install(
     if git_dir is None and git.has_core_hookpaths_set():
         logger.error(
             'Cowardly refusing to install hooks with `core.hooksPath` set.\n'
-            'hint: `git config --unset-all core.hooksPath`',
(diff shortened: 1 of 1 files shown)


wircho:
Thanks for the quick response @asottile. To be clear though, the hint _does not_ unset the global config when present. Do you mean that users can simply find out on their own that `--global` may be needed to fully resolve the problem?

asottile:
if you're blindly running commands and not understanding what they are doing a hint is not going to help you understand what you're doing and whether it is the correct thing to do

it's not possible to solve everyone's situation and so the hint is provided as a hint.  your adjustment of the hint still doesn't necessarily clear it for everyone (since there's other ways to set that through environment, `/etc/gitconfig`, etc.)

wircho:
Fair enough! Thanks so much. I really appreciate the quick and thoughtful response.

PR #3152, opened 2024-03-14
Add guard against the hooks directory being a symbolic link

SimonSegerblomRex:
`pre-commit install` shouldn't mess with directories residing outside of the git repository.

I think you agree with this and at least it should be relevant to add the test even you don't like the implementation.

### DIFF
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -67,7 +67,7 @@ def _install_hook_script(
         overwrite: bool = False,
         skip_on_missing_config: bool = False,
         git_dir: str | None = None,
-) -> None:
+) -> int:
     hook_path, legacy_path = _hook_paths(hook_type, git_dir=git_dir)
 
     # If the hooks directory is a symlink we need to be careful
--- a/pre_commit/commands/install_uninstall.py
+++ b/pre_commit/commands/install_uninstall.py
@@ -70,6 +70,22 @@ def _install_hook_script(
 ) -> None:
     hook_path, legacy_path = _hook_paths(hook_type, git_dir=git_dir)
 
+    # If the hooks directory is a symlink we need to be careful
+    if os.path.islink(os.path.dirname(hook_path)):
+        git_dir = git_dir if git_dir is not None else git.get_git_common_dir()
+        # If the hooks directory links to a directory outside the
+        # git repo we shouldn't try to mess with it
+        if os.path.commonpath(
+            [os.path.realpath(git_dir), os.path.realpath(hook_path)]
+            ) != os.path.realpath(git_dir):
+            logger.error(
+                'Cowardly refusing to install hook script to a directory '
+                'outside of the git repo.\n'
+                f'hint: {os.path.dirname(hook_path)} is a symbolic link '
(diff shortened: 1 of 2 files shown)


pre_commit/commands/install_uninstall.py
```diff
@@ -70,6 +70,22 @@ def _install_hook_script(
 ) -> None:
     hook_path, legacy_path = _hook_paths(hook_type, git_dir=git_dir)
 
+    # If the hooks directory is a symlink we need to be careful
+    if os.path.islink(os.path.dirname(hook_path)):
+        git_dir = git_dir if git_dir is not None else git.get_git_common_dir()
+        # If the hooks directory links to a directory outside the
+        # git repo we shouldn't try to mess with it
+        if os.path.commonpath(
```
SimonSegerblomRex, line 78: This would be a bit cleaner using pathlib:
```python3
Path.absolute(Path(hook_path)).is_relative_to(git_dir)
```
but I'm sure you have good reasons not to use pathlib.

SimonSegerblomRex:
I can reproduce the coverage failure locally:
```sh
coverage report '--omit=pre_commit/languages/*,tests/languages/*'
Name                                       Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------------------------
pre_commit/commands/install_uninstall.py      94      0     36      1    99%   78->89
--------------------------------------------------------------------------------------
TOTAL                                       6257      0   1662      1    99%
```
even though the code is covered by the new test... I'll try to dig into what's happening.

asottile:
what are you trying to solve? I don't think this improves anything and just adds a bunch of complexity

asottile:
please in the future start with an issue to describe your goals before wasting time on implementation

SimonSegerblomRex:
> what are you trying to solve? I don't think this improves anything and just adds a bunch of complexity

A common problem in corporate environments is the use of centralized git hooks. Often the case is that `.git/hooks` is a symbolic link to a directory outside of the git repo, e.g., `/usr/share/<somewhere>/hooks`. A problem that many of my colleagues face is that when they run `pre-commit install` for some repo they either get permission problems, or they end up messing up the corporate pre-commit hook residing in `/usr/share/...` (that is shared with all other repos not using `pre-commit`). I know they don't think this is a stupid setup, but I still think the added test cases are relevant. If you believe that `pre-commit` should be allowed to mess with files outside the git repo I can remove the check in `install_uninstall.py` and change the test.

asottile:
if it's already failing there's nothing needed here then.  the problem is self inflicted

SimonSegerblomRex:
> if it's already failing there's nothing needed here then. the problem is self inflicted

Hmm, what do you mean is failing? I don't understand why you think it's a good thing that `pre-commit install` messes up files outside of the git repo. If you really believe that this should be the case you should still add the test cases to make sure that you don't accidentally removes this "feature".

asottile:
please don't tell me how to think:

> I think you agree with this 

or how to run my project:

> you should still add the test cases to make sure that you don't accidentally remove this "feature".

___

your case is firmly in the "play stupid games win stupid prizes" category.  defining against this is not going to be possible as there are all sorts of other ways this can happen (mounts, symlinks, junctions, etc.) and defending against them doesn't break the hooks system like hooksPath does

PR #3163, opened 2024-03-21
Use `--locked` when installing tools via Cargo

### LINKED ISSUE #3162: Install Rust tools with `--locked` by default or add a parameter to allow this

rust locked

Many (most?) Rust projects come with lock files that specify the exact versions of dependencies with which they are tested. Sometimes, they don't work with newer versions, which is why Cargo offers the `--locked` parameter to build with the dependency versions specified in the lock files.

Currently, pre-commit installs Rust tools without the `--locked` parameter. I suggest to use this parameter by default or offer a way to specify that a hook should be built using that parameter.

### pre-commit --version

pre-commit 3.6.2

mlegner:
Closes #3162

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -155,6 +155,6 @@ def install_environment(
 
         for args in packages_to_install:
             cmd_output_b(
-                'cargo', 'install', '--bins', '--root', envdir, *args,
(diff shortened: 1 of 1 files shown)


asottile:
no I didn't say start working on this

PR #3173, opened 2024-03-28
feat: add options for disabling venv in python

twmht:
In many cases, the dependencies we need are not available on PyPI but are bundled within an image, such as the Arm64 version of Nvidia/TensorRT, which does not offer a wheel file. When using virtualenv, it becomes impossible to locate this package. This PR introduces an option allowing users to use the current Python Environment directly. This is particularly beneficial in CI/CD scenarios.


asottile:
absolutely not.  this bypasses a core promise of pre-commit that it will never touch your system packages

twmht:
@asottile 

how about "cloning" the current virtual env (or system) to the newly venv, this ensures that it does not affect the system level packages and also resolove the dependency problems.

asottile:
no, pre-commit only installs from it's config file otherwise caching is intractable

PR #3188, opened 2024-04-24
feat: utilize cpanminus for perl dependencies if available

Stealthii:
[cpanminus](https://metacpan.org/pod/App::cpanminus) is broadly compatible with our syntax use of [cpan](https://metacpan.org/pod/CPAN), but can offer additional benefits such as `Pkg::Name@Version` definitions, as well as built-in modules to reduce the environment requirements on the system when resolving dependencies.

This PR updates the perl dependency resolver to use `cpanm` if available in the environment. Tests have been updated to ensure operation with either utility, and the Github Actions workflow updated to make the utility available on Ubuntu runners (being already provided in the Windows Perl distribution).


asottile:
why would one want to use this?  I'm inclined to decline unless there's a very good reason and that should start with an issue first rather than a pull request

Issue #3189, opened 2024-04-25
Show additional information when a hook fails


helly25:
### search you tried in the issue tracker

manual

### describe your actual problem

When developers run the pre-commit, they don't always understand what to or where to find more information. The hook description field is already present and often used to provide more descriptive information. For brevity this is not shown in normal cases, but this PR changes the tool behavior to show the description in verbose mode (`-v`). So most users won't see any difference but this can easily be enabled for people who want it.

Alternatives considered:
* Always show the description: Some people may not like this though as it can be spammy.
* Provide a new field: Possible but seems to be just more work with description already present and not otherwise used.
* Show the `repo` link. This is at the wrong level and has a slightly different purpose. Though description can just be set to the same url where that is correct.

### pre-commit --version

pre-commit 3.7.0

asottile:
no thanks -- I'd rather keep it the way it is

PR #3190, opened 2024-04-25
Show additional information when a hook fails

helly25:
Possible implementation for https://github.com/pre-commit/pre-commit/issues/3189

When `run` is called with `-v`, then show `hook.description` if available.

When developers run the pre-commit, they don't always understand what to or where to find more information. The hook description field is already present and often used to provide more descriptive information. For brevity this is not shown in normal cases, but this PR changes the tool behavior to show the description in verbose mode (`-v`). So most users won't see any difference but this can easily be enabled for people who want it.

Alternatives considered:
* Always show the description: Some people may not like this though as it can be spammy.
* Provide a new field: Possible but seems to be just more work with description already present and not otherwise used.
* Show the `repo` link. This is at the wrong level and has a slightly different purpose. Though description can just be set to the same url where that is correct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.egg-info
 *.py[co]
+*.pytest_cache
 /.coverage
 /.tox
+/build
 /dist
 .vscode/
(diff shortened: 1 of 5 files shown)


asottile:
please don't waste time on implementation until a feature is agreed on

PR #3232, opened 2024-06-13
fix: prevent loading the caller's gems from ruby plugins

### LINKED ISSUE #3231: disable loading Gemfiles in the enclosing project in ruby applications

is:issue bundler

I want to use `rubocop` with pre-commit. Internally, rubocop's extension mechanism first uses `Bundler.load` to try to load extensions from the current project (this recurses up through the parent directories of the `cwd` to find the first file named `Gemfile` and then eagerly loads everything in it); this isn't really desirable when it's running through pre-commit, because the rbenv used by pre-commit doesn't have all of the local dependencies installed, so it ends up failing pretty miserably. We pass any extensions as `additional_dependencies` in the `pre-commit-config.yml`; these are also discovered by rubocop through another mechanism.

This isn't a rubocop-specific issue; any Ruby program which calls `Bundle.load` at some point will end up trying to eagerly install all of the dependencies in the current project.

I think in an ideal world, when installing ruby dependencies into an rbenv, pre-commit would create a new `Gemfile` and actually install them with bundler rather than by running `gem install`, but that seems like a larger project.

A change that does work is to simply make an //empty// Gemfile in the rbenv root and point `BUNDLER_GEMFILE` to that. I'll send over a PR which does so.

### pre-commit --version

pre-commit 3.7.1

Roguelazer:
This changes pre-commit to set `BUNDLE_GEMFILE` to point to an empty Gemfile; this prevents ruby-language hooks from attempting to load the gemfile in the invoking application. It's super-hacky, but I'm not immediately sure of a better way to prevent leaking the gems of the invoking application into the isolated environment for ruby hooks.

Fixes #3231

### DIFF
--- a/pre_commit/languages/ruby.py
+++ b/pre_commit/languages/ruby.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import functools
 import importlib.resources
+import os
 import os.path
 import shutil
 import tarfile
@@ -45,6 +46,7 @@ def get_env_patch(
         ('GEM_HOME', os.path.join(venv, 'gems')),
         ('GEM_PATH', UNSET),
         ('BUNDLE_IGNORE_CONFIG', '1'),
+        ('BUNDLE_GEMFILE', os.path.join(venv, 'Gemfile'))
     )
     if language_version == 'system':
         patches += (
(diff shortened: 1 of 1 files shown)


asottile:
https://github.com/pre-commit/pre-commit/issues/3231#issuecomment-2164202868

PR #3235, opened 2024-06-15
feat: allow rust hooks to specify crate features

### LINKED ISSUE #3230: Allow to pass features to cargo install in rust hooks

rust featurres

The rust hooks execute `cargo install --bins --root <root-dir> <packages>` to install the hook, however if a hook wants to use a different set of features than its default features it can't because there's no way to pass either additional flags to the install command or a `--features <features>` to cargo command.

See https://github.com/JohnnyMorganz/StyLua/issues/867 for a real world example where this has become a limitation.

It would be great to have a way to configure a hook to pass cargo additional features when building the hook's package.

### pre-commit --version

3.7.1

Terseus:
Now the rust hooks can use the parameter `additional_dependencies` to receive build-time features for the crate.

The features must start with the string "--feature=" following the name of the feature; e.g. "--feature=full".

Fixes #3230

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -22,6 +22,7 @@ from pre_commit.util import make_executable
 from pre_commit.util import win_exe
 
 ENVIRONMENT_DIR = 'rustenv'
+FEATURE_PREFIX = '--feature='
 health_check = lang_base.basic_health_check
 run_hook = lang_base.basic_run_hook
 
@@ -55,7 +56,8 @@ def get_env_patch(target_dir: str, version: str) -> PatchesT:
         # toolchain
         *(
             (('RUSTUP_TOOLCHAIN', _rust_toolchain(version)),)
-            if version != 'system' else ()
+            if version != 'system'
+            else ()
         ),
     )
 
@@ -68,11 +70,13 @@ def in_env(prefix: Prefix, version: str) -> Generator[None, None, None]:
 
 
 def _add_dependencies(
-        prefix: Prefix,
-        additional_dependencies: set[str],
+    prefix: Prefix,
+    additional_dependencies: set[str],
 ) -> None:
     crates = []
     for dep in additional_dependencies:
+        if dep.startswith(FEATURE_PREFIX):
+            continue
         name, _, spec = dep.partition(':')
         crate = f'{name}@{spec or "*"}'
         crates.append(crate)
@@ -100,20 +104,27 @@ def install_rust_with_toolchain(toolchain: str, envdir: str) -> None:
 
                 # install rustup into `$CARGO_HOME/bin`
                 cmd_output_b(
-                    rustup_init, '-y', '--quiet', '--no-modify-path',
-                    '--default-toolchain', 'none',
+                    rustup_init,
+                    '-y',
+                    '--quiet',
+                    '--no-modify-path',
+                    '--default-toolchain',
+                    'none',
                 )
 
             cmd_output_b(
-                'rustup', 'toolchain', 'install', '--no-self-update',
+                'rustup',
+                'toolchain',
+                'install',
+                '--no-self-update',
                 toolchain,
             )
 
 
 def install_environment(
-        prefix: Prefix,
-        version: str,
-        additional_dependencies: Sequence[str],
+    prefix: Prefix,
+    version: str,
+    additional_dependencies: Sequence[str],
 ) -> None:
     envdir = lang_base.environment_dir(prefix, ENVIRONMENT_DIR, version)
 
@@ -129,6 +140,12 @@ def install_environment(
     # with 'cli:'.
     cli_deps = {
         dep for dep in additional_dependencies if dep.startswith('cli:')
+    }
+    # Features starts with "--feature="
+    features = {
+        dep.replace(FEATURE_PREFIX, '')
+        for dep in additional_dependencies
+        if dep.startswith(FEATURE_PREFIX)
     }
     lib_deps = set(additional_dependencies) - cli_deps
 
@@ -153,8 +170,17 @@ def install_environment(
         if len(lib_deps) > 0:
             _add_dependencies(prefix, lib_deps)
 
+        features_params = []
+        if features:
+            features_params = ['--features', *features]
         for args in packages_to_install:
             cmd_output_b(
-                'cargo', 'install', '--bins', '--root', envdir, *args,
+                'cargo',
+                'install',
+                '--bins',
+                '--root',
+                envdir,
+                *args,
+                *features_params,
                 cwd=prefix.prefix_dir,
             )
--- a/tests/languages/rust_test.py
+++ b/tests/languages/rust_test.py
@@ -44,13 +44,19 @@ def _make_hello_world(tmp_path):
     src_dir.joinpath('main.rs').write_text(
         'fn main() {\n'
         '    println!("Hello, world!");\n'
+        '    if cfg!(feature = "foo") {\n'
+        '        println!("With feature foo");\n'
+        '    }\n'
         '}\n',
     )
     tmp_path.joinpath('Cargo.toml').write_text(
         '[package]\n'
         'name = "hello_world"\n'
         'version = "0.1.0"\n'
-        'edition = "2021"\n',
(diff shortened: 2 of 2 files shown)


pre_commit/languages/rust.py
```diff
@@ -130,6 +141,12 @@ def install_environment(
     cli_deps = {
         dep for dep in additional_dependencies if dep.startswith('cli:')
     }
+    # Features starts with "--feature="
+    features = {
+        dep.replace(FEATURE_PREFIX, '')
+        for dep in additional_dependencies
+        if dep.startswith(FEATURE_PREFIX)
+    }
```
asottile, line 149: I think we can probably just filter out anything that starts with `--` or doesn't have `:`?

asottile (commented):
please rebase out the unrelated formatting changes -- those are not welcome

asottile:
we don't do conventional commits either, please just use a normal commit message

pre_commit/languages/rust.py
```diff
@@ -73,6 +74,8 @@ def _add_dependencies(
 ) -> None:
     crates = []
     for dep in additional_dependencies:
+        if dep.startswith(FEATURE_PREFIX):
+            continue
```
asottile, line 78: I don't think this should be here at all?

pre_commit/languages/rust.py
```diff
@@ -130,6 +133,12 @@ def install_environment(
     cli_deps = {
         dep for dep in additional_dependencies if dep.startswith('cli:')
     }
+    # Features starts with "--feature="
+    features = {
+        dep.replace(FEATURE_PREFIX, '')
```
asottile, line 138: I don't think you should need a `replace(...)` -- we should be able to pass along arguments verbatim I would hope?

asottile:
ideally I'd also like this solution to solve #3162

pre_commit/languages/rust.py
Terseus: What I've implemented uses a custom syntax, so e.g. `--feature=foo --feature=bar` is translated as `--features foo bar` when invoking Cargo.
If I understood you correctly you want to receive arbitrary Cargo arguments in `additional_dependencies` so we could pass, for example, `--features foo bar --locked`.

Is that what you're suggesting?

In that case, every value in `additional_dependencies` after a flag will be interpreted as a Cargo argument, because after a `--features` flag we cannot distinguish between a feature name or a dependency name.

pre_commit/languages/rust.py
Terseus: You're right, the features never reach this function.

Terseus:
@asottile I've pushed a new version that addresses #3162 too.

Now instead of using the syntax "--feature=" in "additional_dependencies" we take all the arguments after an "--" as arbitrary arguments for "cargo install".

Please check it out when you can.

asottile:
hmmm can it work if we just treat things starting with `--` as options instead? or are there things in cargo that explicitly don't work with `--foo=...` ?

Terseus:
> or are there things in cargo that explicitly don't work with `--foo=...` ?

Features, they have the syntax "--features feat1 feat2 feat3".

asottile:
> @asottile
> 
> > or are there things in cargo that explicitly don't work with `--foo=...` ?
> 
> Features, they have the syntax "--features feat1 feat2 feat3".

afaict that's not valid:

```
$ cargo install exa --features git vendored-openssl
    Updating crates.io index
  Downloaded exa v0.10.1
  Downloaded 1 crate (136.6 KB) in 0.10s
    Updating crates.io index
error: could not find `vendored-openssl` in registry `crates-io` with version `*`
  Installing exa v0.10.1
    Updating crates.io index
...
```

it'd have to be `--features 'foo bar'` or `--features foo,bar` -- which maybe we can just require specifically `--features=foo,bar` ?

Terseus:
@asottile 
You're right, it's `--features "git vendored-openssl"` or `--features git,vendored-openssl`, however I don't see what can we gain by accepting only `--features=foo,bar`.

The current implementation can accept any `cargo install` param, long or short; I think that's okay.

2bndy5:
> The current implementation can accept any cargo install param, long or short;

What about `--manifest-path <path/to/Cargo.toml>` or `--bin <bin-name>`? The current command does seem well suited to cargo workspaces:
```
An unexpected error has occurred: CalledProcessError: command: ('\\.cargo\\bin\\cargo.EXE', 'install', '--bins', '--root', '\\AppData\\Local\\Temp\\tmpyqboytse\\repoczr9yu0q\\rustenv-system', '--path', '.')
return code: 101
stdout: (none)
stderr:
    error: found a virtual manifest at `\AppData\Local\Temp\tmpyqboytse\repoczr9yu0q\Cargo.toml` instead of a package manifest
```

2bndy5:
I tried `pre-commit try-repo` using changes in this PR on my rust project (a [cargo "virtual" workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace)), but it didn't work because all `additional-dependencies` are passed to `cargo add`. I need a way to change the args to `cargo install`.

## Why `cargo add`?
Personally speaking, it doesn't make sense to pass features to the [`cargo add`](https://doc.rust-lang.org/cargo/commands/cargo-add.html) command because a well-formed Cargo.toml can adjust the dependency tree (of a binary build) according to the enabled/disabled features passed to a [`cargo install`](https://doc.rust-lang.org/cargo/commands/cargo-install.html) command. AFAIK, the `cargo add` command is strictly for project _development_, I don't think it has any usage in project _deployment_.

If a rust-based binary executable needs to make shell calls to other cargo-built binaries, then the `cargo install` is more suitable for that (or maybe a [custom build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html) that performs `cargo install` when `PRE_COMMIT` env var is set). Adding libs to a project's manifest doesn't seem to achieve anything for deployment. Am I missing a use case or something?

2bndy5:
https://github.com/pre-commit/pre-commit/blob/0f8f383d5371f7b2474a3929d9e157a2d6979ee4/pre_commit/languages/rust.py#L124-L129

This is exactly what the [feature selection for `cargo install`](https://doc.rust-lang.org/cargo/commands/cargo-install.html#feature-selection) aims to resolve. I _really_ think there is a misconception that led to using `cargo add`. `cargo install` build binaries. `cargo add` simply adds dependency specs to a manifest (Cargo.toml), it doesn't build binaries.

asottile:
> You're right, it's --features "git vendored-openssl" or --features git,vendored-openssl, however I don't see what can we gain by accepting only --features=foo,bar.

personally I'd rather be _more_ constrained such that if we need to do something later we have the opportunity to do so rather than being stuck with "allow arbitrary arguments" decision today.  not to mention completely arbitrary arguments likely allows one to break out of the intended isolation

chriskuehl:
> https://github.com/pre-commit/pre-commit/blob/0f8f383d5371f7b2474a3929d9e157a2d6979ee4/pre_commit/languages/rust.py#L124-L129
> 
> This is exactly what the [feature selection for `cargo install`](https://doc.rust-lang.org/cargo/commands/cargo-install.html#feature-selection) aims to resolve. I _really_ think there is a misconception that led to using `cargo add`. `cargo install` builds binaries. `cargo add` simply adds dependency specs to a manifest (Cargo.toml), it doesn't build binaries.

The reason behind this originally was to allow you to override specific library versions used to build the hook binary if needed, similar to how you can use `additional_dependencies` to pin a transitive library version for Python. It wasn't really intended for feature selection. That said, I certainly haven't kept up with Cargo in the six years since this was merged (and maybe it was even wrong in the first place) so I don't object to any changes.

2bndy5:
I opened #3278. Thanks for the historic rationale. The cargo docs don't really preserve historic info like that, so I was completely unaware. I've only been using rust for about a year, but you could probably tell that already 😜

Terseus:
@asottile Should we close this PR in favor of https://github.com/pre-commit/pre-commit/issues/3278? If so, please feel free to do it.

asottile:
> @asottile Should we close this PR in favor of #3278? If so, please feel free to do it.

I think they're a little unrelated to each other

PR #3247, opened 2024-07-04
Introduce `use_filesnames_file` to aid with tools with expensive per-run overhead

thejcannon:
This PR works around a performance issue observed when running pre-commit on large repositories with tools that may have expensive overhead when invoked concurrently (such as tools that load configs remotely, operate on whole directories, have semi-expensive imports, or don't run well concurrently)

(I don't consider this a "problem" with pre-commit, but rather a "quirk" of tools being called in ways they weren't expecting)

I've added unit tests for the new functionality. I'm open to feedback and happy to refine the approach as needed.

(Example benchmark before/after)
```
% time pre-commit run -a infer-deps
Infer dependencies.......................................................Passed
pre-commit run -a infer-deps  4.87s user 6.94s system 100% cpu 11.572 total
% time pre-commit run -a infer-deps
Infer dependencies.......................................................Passed
pre-commit run -a infer-deps  2.79s user 6.71s system 121% cpu 7.75 total
```

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -128,6 +128,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
     cfgv.Optional('always_run', cfgv.check_bool, False),
     cfgv.Optional('fail_fast', cfgv.check_bool, False),
     cfgv.Optional('pass_filenames', cfgv.check_bool, True),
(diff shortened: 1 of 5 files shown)


asottile:
this was rejected, please don't waste my time

PR #3253, opened 2024-07-15
Adjust permission only if needed

gzzi:
We are using pre-commit in a environment where we are not the owner of the repository but we are part of the group having it. In this scenario pre_commit was failing with error:

```
An unexpected error has occurred: PermissionError: [Errno 1] Operation not permitted: '.git/hooks/pre-commit'
```

This change adjust permission only if needed.

### DIFF
--- a/pre_commit/util.py
+++ b/pre_commit/util.py
@@ -42,8 +42,10 @@ def resource_text(filename: str) -> str:
 
 def make_executable(filename: str) -> None:
     original_mode = os.stat(filename).st_mode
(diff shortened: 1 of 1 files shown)


asottile:
I think your scenario should be unsupported.  git is not designed to be used in a shared manner

gzzi:
Ok but the change is not related to only our scenario, it is calling `os.chmod` only if needed. It will not hurt.

asottile:
if pre-commit cannot change the permissions of a file it creates something else has already gone horribly wrong

PR #3257, opened 2024-07-17
[Bugfix][Python] Cache virtual environments by full python path

Lunderberg:
Prior to this commit, a cached virtualenv could be erroneously re-used, causing failures in pre-commit hooks.  This is caused by incorrect normalization in `get_default_version`, resulting in a cache key of `"python3"`, regardless of the minor version used.

This commit updates the implementation of `get_default_version` for python-based hooks to return `os.path.realpath(sys.executable)`.  This ensures that the correct python version is used for running pre-commit hooks.

### DIFF
--- a/pre_commit/languages/python.py
+++ b/pre_commit/languages/python.py
@@ -75,46 +75,9 @@ def _find_by_py_launcher(
     return None
 
 
-def _find_by_sys_executable() -> str | None:
-    def _norm(path: str) -> str | None:
-        _, exe = os.path.split(path.lower())
-        exe, _, _ = exe.partition('.exe')
-        if exe not in {'python', 'pythonw'} and find_executable(exe):
-            return exe
-        return None
-
-    # On linux, I see these common sys.executables:
-    #
-    # system `python`: /usr/bin/python -> python2.7
-    # system `python2`: /usr/bin/python2 -> python2.7
-    # virtualenv v: v/bin/python (will not return from this loop)
-    # virtualenv v -ppython2: v/bin/python -> python2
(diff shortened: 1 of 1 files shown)


Lunderberg:
The steps to reproduce this bug are as follows, though I'm not sure the best way to capture this in a test case.

1. Define a pre-commit hook that depends on the python version.  For example:

   ```
   # In pyproject.toml
   [tool.poetry.dependencies]
   python = "^3.11"

   # In .pre-commit-config.yaml
   ---
   repos:
   - repo: https://github.com/python-poetry/poetry
     rev: '1.7.1'
     hooks:
     - id: poetry-lock
       args: ["--no-update", "--no-cache"]
   ```

2. Run the pre-commit hook in an environment that doesn't satisfy the requirement on the python version.
   
3. Call `pre-commit run`, which fails.  This generates a virtualenv in which the hook is run, using `"python3"` as the key for the cache.

4. Update `PATH` (e.g. by explicitly activating a virtual environment) so that `python3` resolves to a python version that does satisfy the python version.
   
5.  Call `pre-commit run`.  This call should succeed, because `python3` now resolves to a compatible version of python.  However, the cache entry for `"python3"` points to the virtualenv made in step (2), which causes the test to fail.

asottile:
fwiw, if you absolutely need a more specific python version -- use `language_version` -- it's what it is for

Lunderberg:
> please discuss major changes in an issue as well

Apologies, as this seemed like a sufficiently straight-forward bug that it was easier to fix, and didn't seem to be a major change.  I'll open an issue for discussing the bug.

PR #3265, opened 2024-07-23
Support health check for `language: r`

### LINKED ISSUE #3206: Global R upgrade breaks pre-commit

break class update / 2, second upgrade

### describe your issue

R is currently implemented as a second class language, i.e. global interpreter required but packages are bootstrapped by pre-commit in virtual environment with {renv}. R packages are installed for a minor version, e.g. R 4.3 and R is not referenced version-specific on the `$PATH` usually. Hence, updating R means you have to install all these packages again, otherwise your code just won't work (as `R` on the `$PATH` now points to the updated R version). However, pre-commit does not know you just updated R and will fail (unless you already used {renv} to update all required dependencies (e.g. for another project) and have not opted out of the {renv} cache). Currently, the clean solution seems to run `pre-commit clean`  (for all hooks since https://github.com/pre-commit/pre-commit/issues/2338 got rejected). I wonder if other second class languages have similar characteristics. Can pre-commit recognise that the global R version changed and then just re-install that specific environment or at least give helpful error messages? 

Downstream reference: https://github.com/lorenzwalthert/precommit/issues/567

### pre-commit --version

pre-commit 3.3.3

### .pre-commit-config.yaml

```yaml
repo: https://github.com/lorenzwalthert/precommit
```

spell-check..............................................................Failed
- hook id: spell-check
- exit code: 1

ℹ Using R 4.4.0 (lockfile was generated with R 4.3.3)
ℹ Using R 4.4.0 (lockfile was generated with R 4.3.3)
Error in loadNamespace(x) : there is no package called ‘precommit’
Calls: loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted

lorenzwalthert:
Closes #3206.

### DIFF
--- a/pre_commit/languages/r.py
+++ b/pre_commit/languages/r.py
@@ -14,13 +14,74 @@ from pre_commit.envcontext import envcontext
 from pre_commit.envcontext import PatchesT
 from pre_commit.envcontext import UNSET
 from pre_commit.prefix import Prefix
+from pre_commit.util import cmd_output
 from pre_commit.util import cmd_output_b
 from pre_commit.util import win_exe
 
 ENVIRONMENT_DIR = 'renv'
 RSCRIPT_OPTS = ('--no-save', '--no-restore', '--no-site-file', '--no-environ')
 get_default_version = lang_base.basic_get_default_version
-health_check = lang_base.basic_health_check
+
+
+def _execute_vanilla_r_code_as_script(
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
@asottile this is ready for a first review I think.

asottile (approved):

Issue #2732, opened 2023-01-30
make `stages` and `hook-type` match


asottile:
this is mostly a confusion point / slightly annoying and a mistake I made when first introducing new hook types

currently all of the `--hook-type` and `stages` values match except for these three:

- `pre-commit`: `commit`
- `pre-merge-commit`: `merge-commit`
- `pre-push`: `push`

I'd like to deprecate the mismatched names!

this is going to need a long deprecation period since it's sorta subtle and usually not the *user*'s fault and will need hook authors to (potentially) make updates

the plan is to do the following:

1. introduce the following in a minor release
    - no deprecation notice in this initial release (give authors / users some time to update without warnings)
    - the new names are available in addition to the old names
    - `migrate-config` will autofix `.pre-commit-config.yaml` usages of `stages` with the old values (there isn't an equivalent `migrate-manifest` -- though *users* outnumber *hook authors* by several orders of magnitude)
3. introduce the following in a minor release
   - a warning is shown for *configuration* using the old `stages` names
   - a warning is shown *once* for repos using the old `stages` names (as to not be super annoying for users who have no control)
   - a recommendation for hook authors to also set `minimum_pre_commit_version` to this version
4. a long time passes (typically my deprecation period has been 12-18+ months)
5. introduce the following in a major release
    - removal of the old names

archana-1209:
is someone working on this , or can I pick this up ? @asottile

asottile:
@archana-1209 we're sort of in the "wait out the deprecation period" phase of this so at the time there isn't much to do

PR #3302, opened 2024-09-17
migrate-config rewrites deprecated stages

asottile:
part of #2732


francastagna:
Is there a way to avoid this rewriting when running `pre-commit autoupdate`? @asottile 
If an enterprise was parsing the pre-commit-config.yaml file in the hook itself, then this could be a breaking change for them.

asottile:
you willingly upgraded to 4.0 you tell me?

PR #3304, opened 2024-09-18
disable automatic toolchain switching for golang hooks

### LINKED ISSUE #3300: Disable automatic toolchain switching

toolchain

### describe your issue

In version 1.21 go introduced [automatic toolchain switching](https://go.dev/doc/toolchain), which basically means downloading the go version specified in the `go.mod` file if it's newer than the one being used.

This is the cause of #3149. In that case the command failed only because the helm docs set `go 1.22` in `go.mod`. Had they set `1.22.0` that version would have been downloaded and the build would work.

This may not be bad if `language_version` isn't specified, because autoupdate would only break in case the full version isn't specified like the above. However, this breaks the `language_version` if it is specified. Here's an example:

```yaml
default_language_version:
  golang: "1.21.0"

repos:
- repo: https://github.com/golangci/golangci-lint
  rev: v1.61.0  # uses go 1.22.1
  hooks:
    - id: golangci-lint-config-verify
 ```

```console
$ ~/.cache/pre-commit/repoka9x_fj6/golangenv-1.21.0/bin/golangci-lint version
golangci-lint has version (devel) built with go1.22.1 from (a1d6c560de1a193a0c68ffed68cd5928ef39e884, modified: true, mod sum: "") on 2024-09-09T14:33:19Z
```

### pre-commit --version

pre-commit 3.8.0

### .pre-commit-config.yaml

```yaml
default_language_version:
  golang: "1.21.0"

repos:
- repo: https://github.com/golangci/golangci-lint
  rev: v1.61.0  # uses go 1.22.1
  hooks:
    - id: golangci-lint-config-verify
```

_No response_

AleksaC:
Resolves #3300 

I only disabled the toolchain switching if the version is not `system`. I did this because `system` is the default if go is installed globally and no `language_version` is explicitly specified.  In this case users likely doesn't care which version of go is used as long as their hooks are working, so I think it makes more sense to allow the toolchain switching since it doesn't break user expectations and other than a small performance hit on hook installation doesn't have negative impact.

If we were to disable toolchain switching regardless of the version, many hooks would break since many environments lag with their go version. For example, in GitHub-hosted runners for GitHub Actions the pre-installed go version is `1.21` which is two versions behind the current version, so it's likely that a lot of hooks work thanks to automatic toolchain switching without even knowing.

### DIFF
--- a/pre_commit/languages/golang.py
+++ b/pre_commit/languages/golang.py
@@ -75,6 +75,7 @@ def get_env_patch(venv: str, version: str) -> PatchesT:
 
     return (
         ('GOROOT', os.path.join(venv, '.go')),
+        ('GOTOOLCHAIN', 'local'),
         (
             'PATH', (
                 os.path.join(venv, 'bin'), os.pathsep,
@@ -145,6 +146,7 @@ def install_environment(
     env = no_git_env(dict(os.environ, GOPATH=gopath))
     env.pop('GOBIN', None)
     if version != 'system':
+        env['GOTOOLCHAIN'] = 'local'
         env['GOROOT'] = os.path.join(env_dir, '.go')
         env['PATH'] = os.pathsep.join((
             os.path.join(env_dir, '.go', 'bin'), os.environ['PATH'],
--- a/tests/languages/golang_test.py
+++ b/tests/languages/golang_test.py
@@ -11,11 +11,13 @@ from pre_commit.commands.install_uninstall import install
 from pre_commit.envcontext import envcontext
 from pre_commit.languages import golang
 from pre_commit.store import _make_local_repo
+from pre_commit.util import CalledProcessError
 from pre_commit.util import cmd_output
 from testing.fixtures import add_config_to_repo
 from testing.fixtures import make_config_from_repo
 from testing.language_helpers import run_language
 from testing.util import cmd_output_mocked_pre_commit_home
+from testing.util import cwd
 from testing.util import git_commit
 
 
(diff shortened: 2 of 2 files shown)


AleksaC:
Unrelated to this PR, but mypy fails with `Module has no attribute "sched_getaffinity"` likely because it's not available on MacOS:
https://github.com/pre-commit/pre-commit/blob/de8590064e181c0ad45d318a0c80db605bf62a60/pre_commit/xargs.py#L31

Replacing try/except with `hasattr` would fix it, but I didn't want to change it because it's outside the scope of the current PR.

pre_commit/languages/golang.py
```diff
@@ -145,6 +145,7 @@ def install_environment(
     env = no_git_env(dict(os.environ, GOPATH=gopath))
     env.pop('GOBIN', None)
     if version != 'system':
+        env['GOTOOLCHAIN'] = 'local'
```
asottile, line 148: I think (?) this should be part of `get_env_patch` ?

AleksaC: From what I can see `get_env_patch` is only called from `in_env` which is used when running the hook and not during the installation. This should only be relevant during installation because that's when the build happens. 

I ran the test to make sure it works this way and it breaks when I move it.

asottile: hmmm I guess? but what about for hooks that just call `gofmt` -- I know the rust toolchain gets grumbly when the versions mismatch -- does something like this survive?

```yaml
-   repo: local
    hooks:
    -   id: gofmt
        name: gofmt
        language: golang
        entry: gofmt -l -w
        types: [go]
```

AleksaC: `gofmt` doesn't seem to care about this option, but `go fmt` performs the toolchain switch. Looking at the code it seems that any `go` command will try to perform a [toolchain switch](https://github.com/golang/go/blob/bae2e968e2daadd39c1bdb1221648361d7277ddc/src/cmd/go/main.go#L105).

I guess the variable should be set in both places then.

tests/languages/golang_test.py
```diff
@@ -165,3 +166,72 @@ def test_during_commit_all(tmp_path, tempdir_factory, store, in_git_dir):
...
+
+    with pytest.raises(CalledProcessError) as excinfo:
+        run_language(
+            path=tmp_path,
+            language=golang,
+            version='1.22.0',
+            exe='golang-version-test',
+        )
+
+        assert 'go.mod requires go >= 1.23.1' in excinfo.value
```
asottile, line 196: this line is never run because it is indented

tests/languages/golang_test.py
```diff
@@ -165,3 +166,72 @@ def test_during_commit_all(tmp_path, tempdir_factory, store, in_git_dir):
...
+
+    with monkeypatch.context() as m:
+        m.chdir(str(test_dir))
+
+        ret, out = run_language(
+            path=hook_dir,
+            language=golang,
+            version='1.22.0',
+            exe='go fmt',
+            file_args=(str(main_file)),
```
asottile, line 233: this is just a parenthesized string and should be a tuple I believe -- I'm kinda surprised the type checker was ok with this? (oh duh, `str` is a `Sequence[str]` so it just kinda lets it happen)

pre_commit/languages/golang.py
```diff
@@ -75,6 +75,7 @@ def get_env_patch(venv: str, version: str) -> PatchesT:
 
     return (
         ('GOROOT', os.path.join(venv, '.go')),
+        ('GOTOOLCHAIN', 'local'),
```
asottile, line 78: does this break with `language_version: system`? this seems different than the install code below

tests/languages/golang_test.py
```diff
@@ -165,3 +166,72 @@ def test_during_commit_all(tmp_path, tempdir_factory, store, in_git_dir):
...
+    hook_dir.joinpath('go.mod').write_text(go_mod_hook)
+
+    test_dir = tmp_path.joinpath('test')
+    test_dir.mkdir()
+    test_dir.joinpath('go.mod').write_text(go_mod)
+    main_file = test_dir.joinpath('main.go')
+    main_file.write_text(main_go)
+
+    with monkeypatch.context() as m:
+        m.chdir(str(test_dir))
```
asottile, line 226: there's a helper for this I believe (I think it's called `with chdir(...)`? or `with cwd(...)`)

pre_commit/languages/golang.py
AleksaC: If the `language_version` is `system` the function is going to return before reaching that line, so it's consistent with the install code.

asottile (approved):

PR #3310, opened 2024-09-28
prevent mypy from failing on systems without os.sched_getaffinity

AleksaC:
mypy fails on MacOs with `error: Module has no attribute "sched_getaffinity"  [attr-defined]` because it isn't available there. Since the `AttributeError` is ignored here, there's no need for mypy to check for it either.


asottile:
I only run mypy on Linux and have no intention of supporting it elsewhere

PR #3322, opened 2024-10-07
Fix typos all around

szepeviktor:
Found very few misspellings.

https://github.com/crate-ci/typos


asottile:
no thanks this doesn't improve things

PR #3344, opened 2024-10-28
Bump pre-commit-hooks rev in sample config

agrueneberg:
v3.2.0 -> v5.0.0


vishalnandagopal:
Hi @asottile ! I was just about to create another PR and noticed you've closed many similar ones.

The current sample config causes a warning right after installing. While I did read that this is not meant to be up to date, I feel that it should not throw warnings and maybe run with as little hassles as possible.

And the docs should be up to date atleast, for people copying from the home page.

asottile:
as you can see I am not interested in updating the value there.  it's actually better if people learn `pre-commit autoupdate` anyway

PR #3348, opened 2024-11-01
Add support for julia hooks

### LINKED ISSUE #2689: 2nd-class support Julia
Would `pre-commit` be open to supporting Julia as a first-class language? There's at least one project that Julia seem keen to have `pre-commit` support ([`JuliaFormatter.jl`][jlfmt-support]). I realized in attempting to provide `pre-commit` support that there's no first-class support for Julia.

I believe I'd be open to making a PR to provide said support, but wanted to ask if it's something `pre-commit` would be open to supporting, in the first place. 🙂 

[jlfmt-support]: https://github.com/domluna/JuliaFormatter.jl/issues/498

fredrikekre:
This patch adds 2nd class support for hooks using julia as the language. pre-commit will install any dependencies defined in the hooks repo `Project.toml` file, with support for `additional_dependencies` as well. Julia doesn't (yet) have a way to install binaries/scripts so for julia hooks the `entry` value is a (relative) path to a julia script within the hooks repository. When executing a julia hook the (globally installed) julia interpreter is prepended to the entry.

Example `.pre-commit-hooks.yaml`:

```yaml
- id: foo
  name: ...
  language: julia
  entry: bin/foo.jl --arg1
```

Example hooks repo: https://github.com/fredrikekre/runic-pre-commit/tree/fe/julia
Accompanying pre-commit.com PR: https://github.com/pre-commit/pre-commit.com/pull/998

Fixes #2689.

### DIFF
--- a/pre_commit/languages/julia.py
+++ b/pre_commit/languages/julia.py
@@ -1,0 +1,132 @@
+from __future__ import annotations
+
+import contextlib
+import os
+import shutil
+from collections.abc import Generator
+from collections.abc import Sequence
+
+from pre_commit import lang_base
+from pre_commit.envcontext import envcontext
+from pre_commit.envcontext import PatchesT
+from pre_commit.envcontext import UNSET
+from pre_commit.prefix import Prefix
+from pre_commit.util import cmd_output_b
+
+ENVIRONMENT_DIR = 'juliaenv'
+health_check = lang_base.basic_health_check
+get_default_version = lang_base.basic_get_default_version
+
+
+def run_hook(
+        prefix: Prefix,
+        entry: str,
+        args: Sequence[str],
+        file_args: Sequence[str],
+        *,
+        is_local: bool,
+        require_serial: bool,
+        color: bool,
+) -> tuple[int, bytes]:
+    # `entry` is a (hook-repo relative) file followed by (optional) args, e.g.
+    # `bin/id.jl` or `bin/hook.jl --arg1 --arg2` so we
+    # 1) shell parse it and join with args with hook_cmd
+    # 2) prepend the hooks prefix path to the first argument (the file), unless
+    #    it is a local script
+    # 3) prepend `julia` as the interpreter
+
+    cmd = lang_base.hook_cmd(entry, args)
+    script = cmd[0] if is_local else prefix.path(cmd[0])
+    cmd = ('julia', script, *cmd[1:])
+    return lang_base.run_xargs(
+        cmd,
+        file_args,
+        require_serial=require_serial,
+        color=color,
+    )
+
+
+def get_env_patch(target_dir: str, version: str) -> PatchesT:
+    return (
+        ('JULIA_LOAD_PATH', target_dir),
+        # May be set, remove it to not interfer with LOAD_PATH
+        ('JULIA_PROJECT', UNSET),
+    )
+
+
+@contextlib.contextmanager
+def in_env(prefix: Prefix, version: str) -> Generator[None]:
+    envdir = lang_base.environment_dir(prefix, ENVIRONMENT_DIR, version)
+    with envcontext(get_env_patch(envdir, version)):
+        yield
+
+
+def install_environment(
+        prefix: Prefix,
+        version: str,
+        additional_dependencies: Sequence[str],
+) -> None:
+    envdir = lang_base.environment_dir(prefix, ENVIRONMENT_DIR, version)
+    with in_env(prefix, version):
+        # TODO: Support language_version with juliaup similar to rust via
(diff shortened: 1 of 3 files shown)


pre_commit/languages/julia.py
```diff
@@ -0,0 +1,127 @@
...
+        require_serial: bool,
+        color: bool,
+) -> tuple[int, bytes]:
+    # `entry` is a (hook-repo relative) file followed by (optional) args, e.g.
+    # `bin/id.jl` or `bin/hook.jl --arg1 --arg2` so we
+    # 1) shell parse it and join with args with hook_cmd
+    # 2) prepend the hooks prefix path to the first argument (the file)
+    # 3) prepend `julia` as the interpreter
+    cmd = lang_base.hook_cmd(entry, args)
+    cmd = ('julia', prefix.path(cmd[0]), *cmd[1:])
```
asottile, line 34: I suspect this will need special handling for `is_local: True` -- otherwise hooks won't be able to use `julia` with local scripts -- `language: r` is very similar to this so perhaps some inspiration can be drawn from that? (although `r` has a lot more complicated setup so I guess just look at run part for inspiration)

pre_commit/languages/julia.py
```diff
@@ -0,0 +1,127 @@
...
+        color=color,
+    )
+
+
+def get_env_patch(target_dir: str, version: str) -> PatchesT:
+    return (
+        # Single entry pointing to the hook env
+        ('JULIA_LOAD_PATH', target_dir),
+        # May be set, remove it to not interfer with LOAD_PATH
+        ('JULIA_PROJECT', ''),
```
asottile, line 48: there's an `Unset` thing which I believe is better than setting to an empty string

pre_commit/languages/julia.py
```diff
@@ -0,0 +1,127 @@
...
+        cmd,
+        file_args,
+        require_serial=require_serial,
+        color=color,
+    )
+
+
+def get_env_patch(target_dir: str, version: str) -> PatchesT:
+    return (
+        # Single entry pointing to the hook env
```
asottile, line 45: idk that this comment is helpful

pre_commit/languages/julia.py
```diff
@@ -0,0 +1,127 @@
...
+            error("No (Julia)Project.toml found in hooks repository")
+        end
+
+        # Copy Manifest.toml to hook env (not mandatory)
+        manifest_names = ("JuliaManifest.toml", "Manifest.toml")
+        for manifest_name in manifest_names
+            isfile(manifest_name) || continue
+            cp(manifest_name, joinpath(hook_env, manifest_name))
+            break
+        end
```
asottile, line 101: I think (?) it would be preferable to do this logic in python rather than in julia

tests/languages/julia_test.py
```diff
@@ -0,0 +1,61 @@
...
+        end
+    end
+    main()
+    """
+    _make_hook(tmp_path, code)
+    deps = ('TOML=fa267f1f-6049-4f14-aa54-33bafae1ed76',)
+    ret, out = run_language(tmp_path, julia, 'src/main.jl', deps=deps)
+    assert ret == 0
+    assert b'Example = 7876af07-990d-54b4-ab0e-23690620f79a' in out
+    assert b'TOML = fa267f1f-6049-4f14-aa54-33bafae1ed76' in out
```
asottile, line 61: maybe I just don't know enough about julia but from a quick search it seems like TOML is a stdlib library? does it still need to be installed?

pre_commit/languages/julia.py
fredrikekre: Yes, thanks for the pointer. Added a test.

tests/languages/julia_test.py
fredrikekre: It is a standard library, and it can be loaded at top level (e.g. in the Julia REPL) by default, since the `@stdlib` entry is included in the default `LOAD_PATH`. However, for Julia packages it is required (so people are used to explicitly adding also stdlibs), and for full reproducibility it is recommended. There is also ongoing work to separate stdlibs to make them less tied to the Julia version, so then they would behave more like normal packages that can be upgraded separately etc.

We can of course make the choice to include `@stdlib` in the load path for julia hooks, but I would argue that it is better to be explicit with the dependencies in this case.

fredrikekre:
Thanks for the review and merge! I have updated https://github.com/pre-commit/pre-commit.com/pull/998 to match the final implementation here.

fredrikekre:
@asottile just out of curiosity, is there a timeline for a new release (4.1.0?) that includes this feature? Thanks!

PR #3358, opened 2024-11-29
[RFC] Add possibility to specify paths relative to .pre-commit-config.yaml in the configuration

chrta-wmi:
I open this as a Draft RFC PR to discuss if something like this could be merged or if there is a better solution.

**Usecase**

We use pre-commit hooks in a yocto project were multiple git repositories are used. To prevent copy-and-paste only one of these repositories contains the `.pre-commit-config.yaml`. When executing pre-commit, this config file is always used: `pre-commit run --config <main project>/.pre-commit-config.yaml`. The path `<main project>` differs in different environments and is not fixed.

This works fine if every hook is completely configured via the command line, without using any config files. But if we want to use a config file for a hook and this file is checked into the same repository as the `.pre-commit-config.yaml` file, it is not possible to specify the correct path in the `.pre-commit-config.yaml`. This is the issue this PR tries to solve.

**Proposed Solution**

The implementation consists of two steps:
1. Provide the environment variable `PRE_COMMIT_CONFIG_PATH` which contains the abolute path of the directory of the used pre-commit configuration file.
2. Add the possibility to use environment variable in `.pre-commit-config.yaml`. The variable is replaced when the configuration is parsed.

**Open Issues**

1. The unit test coverage is not 100% anymore, because i implemented one test in `tests/commands/validate_config_test.py` which is ignored when the coverage is evaluated.
2. How should it be handled if the specified environment variable is not available? Currently this is silently ignored and an empty string is used.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -412,6 +412,10 @@ def run(
     if args.rewrite_command:
         environ['PRE_COMMIT_REWRITE_COMMAND'] = args.rewrite_command
 
+    environ['PRE_COMMIT_CONFIG_PATH'] = (
+        os.path.dirname(os.path.abspath(config_file))
+    )
+
     # Set pre_commit flag
     environ['PRE_COMMIT'] = '1'
 
--- a/pre_commit/yaml.py
+++ b/pre_commit/yaml.py
@@ -1,11 +1,26 @@
 from __future__ import annotations
 
 import functools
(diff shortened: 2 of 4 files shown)


asottile:
this has been rejected before

chrta-wmi:
> this has been rejected before

Can you point me to the discussion? I was not able to find it.

PR #3368, opened 2024-12-18
feat(config): read config path from env

alexanderilyin:
I'm cleaning top level of https://github.com/partcad/partcad/ repo and this will help me to move one more file from the root folder.


asottile:
you can use `--config`.  and environment variable has been rejected in the past already

PR #3372, opened 2024-12-23
fix: failing post-checkout on single file checkout when solving merge conflicts

nekrich:
### Current behavior 
pre-commit doesn't allow to solve merge conflicts as `git checkout --theirs -- <paths>`.

Reason: git status returns unmerged files 🤷.

But this is wrong, since there are a lot of unmerged paths I'm trying to resolve with `git checkout`. Which triggers the post-checkout hook.

### New behavior
Do not run post-checkout hooks if there are unmerged files.
 
This handles cases when there are unmerged paths and:
 * resolving conflict with `git checkout --ours/--theirs -- <paths>`
 * or checking out single files with `git checkout <branch> -- <paths>`

### Afterthoughts
I understand it kinda breaks expectations on pre-commit behavior, but at least it doesn't break git operations.

As I understand this is the reason:
```python
    with contextlib.ExitStack() as exit_stack:
        if stash:
            exit_stack.enter_context(staged_files_only(store.directory))
```
But I don't have enough knowledge about how to pre-commit works with those stashes to fix it correctly. And especially cover the case with unmerged pre-commit config.

That's why I suggest to have this little limitation in behavior, but working git checkout with unmerged files and installed post-checkout hook.

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -344,8 +344,20 @@ def run(
 
     # Check if we have unresolved merge conflict files and fail fast.
     if stash and _has_unmerged_paths():
-        logger.error('Unmerged files.  Resolve before committing.')
-        return 1
+        # Do not fail on unmerged files if this is a post-checkout hooks.
+        # This handles cases when there are unmerged paths and:
+        #   * resolving conflict with `git checkout --ours/--theirs -- <paths>`
+        #   * or checking out single files with `git checkout <b> -- <paths>`.
+        if args.hook_stage == 'post-checkout':
+            output.write_line(
+                f'Skipping `{args.hook_stage}` hooks since '
+                f'it\'s a file checkout or same head ref',
+            )
+            return 0
+        else:
+            logger.error('Unmerged files.  Resolve before committing.')
+            return 1
+
     if bool(args.from_ref) != bool(args.to_ref):
         logger.error('Specify both --from-ref and --to-ref.')
         return 1
--- a/tests/commands/run_test.py
+++ b/tests/commands/run_test.py
@@ -536,6 +536,22 @@ def test_merge_conflict(cap_out, store, in_merge_conflict):
     assert b'Unmerged files.  Resolve before committing.' in printed
 
 
+def test_merge_conflict_post_checkout(cap_out, store, in_merge_conflict):
+    ret, printed = _do_run(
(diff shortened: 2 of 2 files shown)


asottile:
I don't think this is correct. this blindly turns off the hook

asottile:
note that post-checkout does not prevent checkout.  so I suspect you might just be confused by the output?

nekrich:
@asottile , this turns of the hook only if there are unmerged changes. Please don't ruch to close the PR.

Here is my output:
```sh
> git write-tree                                                                             
Scripts/bootstrap/spm.sh: unmerged (734b5ea642b6d59bc9c5a4f6ce71d6a6d3bac421)
Scripts/bootstrap/spm.sh: unmerged (cb960fa7e94ae4680d61e12e56e893edced5e23b)
```

I've added some log to the hook:
```
hook-impl --config=.pre-commit-config.yaml --hook-type=post-checkout --hook-dir /Users/..../.git/hooks -- da8bc09b69d77a10e68a4a85576f6beb640cf887 da8bc09b69d77a10e68a4a85576f6beb640cf887 0
```

And I get error when I:
```
> git checkout --theirs -- Scripts/bootstrap/spm.sh                                              
[ERROR] Unmerged files.  Resolve before committing.
```

The error comes directly form the pre-commit.
An yes, pre-commit post-checkout prevents checkout if there are merge conflicts.

asottile:
no it doesn't

nekrich:
@asottile , it does. And you check it here: https://github.com/nekrich/test-repo

## Breaking pre-commit

1. Clone the repo https://github.com/nekrich/test-repo.git.

2. Run `cd test-repo`.

4. Fetch all branches `git fetch --all`.

5. Install pre-commit hooks by running `pre-commit install -f -t post-checkout` or `configure.sh`

6. Run `break-me.sh` or two commands:
  * `git cherry-pick --edit 98cd69f8c9b34d6b1dbc552ff1dd5db1cb1fc1d0`
  * `git checkout new-main -- configure.sh`

7. Enjoy [ERROR] Unmerged files.  Resolve before committing.

asottile:
and yet, `configure.sh` is successfully checked out from `new-main`

nekrich:
and yet, returned code is not 0.

asottile:
I don't have time for your attitude

PR #3398, opened 2025-02-04
Add pipenv language support to pre-commit

macieyng:
- Implement pipenv language module in pre-commit
- Update workflows to include pipenv language tests
- Add pipenv language to allowed languages in testing script
- Create comprehensive test suite for pipenv language support


asottile:
this won't work as written:  it installs from the repository under test which breaks caching completely

Issue #3089, opened 2023-12-08
Allow using bun instead of node


gaborbernat:
### search you tried in the issue tracker

bun

### describe your actual problem

Documentation about bun: https://bun.sh.

The primary benefit would be much faster startup/run on larger projects.

### pre-commit --version

latest

gaborbernat:
I'll try if I get to it, no promises 😆 would it be possible to convince https://github.com/pre-commit/mirrors-prettier#prettier-mirror to use bun without needing changes to that repo, though?

asottile:
yeah in theory you'd set `language: bun` and it should just work assuming we make `language: bun` as compatible with `language: node` as possible

gaborbernat:
So it would be:
```
-   repo: https://github.com/pre-commit/mirrors-prettier
    language: bun
    rev: ''  # Use the sha / tag you want to point at
    hooks:
    -   id: prettier
```
?

matanshavit:
I gave it a shot [#3568](https://github.com/pre-commit/pre-commit/pull/3568)

gaborbernat:
There does not seem to be enough interest in this, so I am closing it. Thanks for taking a look.

PR #3403, opened 2025-02-16
Add bun as a supported language.

mrswats:
This PR implements `bun` as a new language.

Closes https://github.com/pre-commit/pre-commit/issues/3089

### DIFF
--- a/pre_commit/all_languages.py
+++ b/pre_commit/all_languages.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from pre_commit.lang_base import Language
+from pre_commit.languages import bun
 from pre_commit.languages import conda
 from pre_commit.languages import coursier
 from pre_commit.languages import dart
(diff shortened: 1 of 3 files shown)


mrswats:
Still a WIP. Working on the tests. I pushed before realizing I was missing a couple of things.

mrswats:
Okay, I need some help. I am not sure what tests to write because I feel like `node` and `npm` are a bit quirky and some of the tests are meant to reflect that but with bun I am not super sure what to do.

One thing I noticed is that I do not know how to run binary executables defined in `packages.json`. The docs say you need to run `bun x` but I think this is not for a local repo but rather a remote repo.

So I am a bit lost, here and any help would be very appreciated.

asottile:
yeah admittedly I'm not super familiar with how bun works but I'm pretty sure as written this won't work generally

somehow need to have bun install the dependencies into the `bun_env` directory and then execute from there.  I'm not sure (?) if there needs to be nodeenv involved (since I thought bun was its own node executable?) -- and I'm even less familiar with _executing_ things with bun

mrswats:
Afaik, bun does not __yet__ support all the commands npm uses to install stuff so I will close this for now.

PR #3418, opened 2025-03-03
Use plain sh for the pre-commit script

WhyNotHugo:
Bash is currently a dependency of pre-commit. It is only used in the pre-commit script, and the only bash-specific feature that is used is a single array.

Use the `args` array instead, which works on plain sh (e.g.: POSIX sh). This change drops bash as a runtime dependency.

Tested with busybox sh(ash) and OpenBSD's sh(ksh).


asottile:
this introduces a shell injection.  bash is ubiquitous on any modern system

PR #3431, opened 2025-03-19
Add freeze config option for autoupdate command

dbast:
This is an idea that starts with a PR to see how much implementation effort this is and how much acceptable it is.

It enables users to activate hook freezing/pinning by setting 'freeze: true' in their pre-commit configuration file.

The intend are two effects, when setting `freeze: true`:
* Enables freeze automatically when running `pre-commit autoupdate` (no need then anymore to specify --freeze for repeated `autoupdate` calls).
* Enables https://pre-commit.ci/ to freeze hooks as part of update PRs. (this is an assumption: does pre-commit.ci actually use the `autoupdate` subcommand during update PR generation?)

This is done in context of e.g. https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/ where a gh account got compromised and somebody mutated a git tag to point a malicious commit. This could also happen to pre-commit hooks, where then a local hook execution can do all kind of unwanted things. Pinning to sha can prevent this or at least makes such happenings visible (in case of sha changes while semver remains unchanged).

Happy to also propose a doc update PR to https://github.com/pre-commit/pre-commit.com or implement this in a different way in case the idea is acceptable.

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -466,6 +466,7 @@ CONFIG_SCHEMA = cfgv.Map(
     cfgv.Optional('files', check_string_regex, ''),
     cfgv.Optional('exclude', check_string_regex, '^$'),
     cfgv.Optional('fail_fast', cfgv.check_bool, False),
+    cfgv.Optional('freeze', cfgv.check_bool, False),
     cfgv.WarnAdditionalKeys(
         (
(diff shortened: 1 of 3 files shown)


asottile:
this isn't needed.  use the argument and pre-commit ci will already preserve it

PR #3435, opened 2025-03-25
update workflows for python 3.13 release

shenxianpeng:


asottile:
please don't touch infra in projects you don't own

Issue #3440, opened 2025-03-31
try-tepo failure message useless if no commits are in repo


jaredkoontz:
### search you tried in the issue tracker

https://github.com/pre-commit/pre-commit/issues?q=is%3Aissue%20%22try-repo%22%20(empty%20OR%20no%20commits)

### describe your issue

If one runs `try-repo` on a repo without a commit, pre-commit will silently fail. 

There is no indication on how to fix it, without going into the source. Upon going into the source, I have found out that the repo requires a commit in order to be used in `try-repo`.

## To Replicate
try to run `try-repo` on a repo that does not have a commit.

```bash
❯ git init ./some-path

❯ cat <<EOF > ./some-path/.pre-commit-hooks.yaml
- id: some-id
  name: some-id
  description: "do stuff"
  entry: echo 'hi'
  language: system
  types_or: []
  args: []
  require_serial: true
  additional_dependencies: []
  minimum_pre_commit_version: "2.9.2"
EOF
❯ pre-commit try-repo ./some-path some-id --verbose

# output of final command
An unexpected error has occurred: CalledProcessError: command: ('C:\\Program Files\\Git\\cmd\\git.EXE', 'ls-remote', '--exit-code', '..\\..\\some-path', 'HEAD')
return code: 2
stdout: (none)
stderr: (none)
Check the log at C:\Users\myname\.cache\pre-commit\pre-commit.log
```

Because of https://git-scm.com/docs/git-ls-remote#Documentation/git-ls-remote.txt---exit-code we only get a return code of 2 with no stderr or stdout.

```
    raise CalledProcessError(returncode, cmd, stdout_b, stderr_b)
pre_commit.util.CalledProcessError: command: ('C:\\Program Files\\Git\\cmd\\git.EXE', 'ls-remote', '--exit-code', '..\\..\\some-path', 'HEAD')
```

```bash 
git ls-remote ./some-path HEAD
```
but running this we get nothing other than the error code (notice the prompt color change):

![Image](https://github.com/user-attachments/assets/1dfff6e3-93c3-4f13-92be-a2ba3695e01d)
![Image](https://github.com/user-attachments/assets/10f60764-8b55-47eb-b471-85575a317188)

Unless I am missing in the docs, this was not obvious to me that we needed a commit in the repo `try-repo`.

To fix:
```bash
❯ cd some-path
❯ git add .pre-commit-hooks.yaml
❯ git commit -m 'my message'
[...]

PR #3441, opened 2025-03-31
feat: create error message if try-repo fails because there are no commits in the repo

jaredkoontz:
Make it more obvious to the end user that their garbage_in_repo (a bare repo without commits) is indeed garbage by letting them know the reason pre-commit is failing (there are no commits on the repo).

follow up from:
https://github.com/pre-commit/pre-commit/issues/3440

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -168,7 +168,19 @@ def get_changed_files(old: str, new: str) -> list[str]:
 
 
 def head_rev(remote: str) -> str:
-    _, out, _ = cmd_output('git', 'ls-remote', '--exit-code', remote, 'HEAD')
+    try:
+        _, out, _ = cmd_output(
+            'git', 'ls-remote', '--exit-code',
+            remote, 'HEAD',
+        )
+    except CalledProcessError as e:
+        if e.returncode == 2:
+            raise FatalError(
+                f'repo {remote} found but appears to have no commits. '
+                f'Make a commit in {remote} to use it',
+            )
+        raise e
+
     return out.split()[0]
 
 
--- a/tests/commands/try_repo_test.py
+++ b/tests/commands/try_repo_test.py
@@ -100,19 +100,28 @@ def test_try_repo_relative_path(cap_out, tempdir_factory):
         assert not try_repo(try_repo_opts(relative_repo, hook='bash_hook'))
 
 
-def test_try_repo_no_commits(cap_out, tempdir_factory):
+@pytest.mark.parametrize(
+    'has_commits', [
+        False,
+        True,
+    ],
+)
+def test_try_repo_no_commits(cap_out, tempdir_factory, has_commits):
     repo = make_repo(
         tempdir_factory,
         'modified_file_returns_zero_repo',
-        commits=False,
+        commits=has_commits,
     )
 
     with cwd(git_dir(tempdir_factory)):
         bare_repo = os.path.join(repo, '.git')
         # previously crashed attempting modification changes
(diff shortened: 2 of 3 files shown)


pre_commit/git.py
```diff
@@ -168,7 +168,19 @@ def get_changed_files(old: str, new: str) -> list[str]:
...
+        _, out, _ = cmd_output(
+            'git', 'ls-remote', '--exit-code',
+            remote, 'HEAD',
+        )
+    except CalledProcessError as e:
+        if e.returncode == 2:
+            raise FatalError(
+                f'repo {remote} found but appears to have no commits. '
+                f'Make a commit in {remote} to use it',
+            )
```
asottile, line 181: error code 2 could mean other things -- it's better to just let this raise and then at that point there's no reason for this patch

tests/commands/try_repo_test.py
```diff
@@ -98,6 +101,37 @@ def test_try_repo_relative_path(cap_out, tempdir_factory):
...
+
+    with cwd(git_dir(tempdir_factory)):
+        bare_repo = os.path.join(repo, '.git')
+        # previously crashed attempting modification changes
+        if not has_commits:
+            with pytest.raises(FatalError) as e:
+                try_repo(try_repo_opts(bare_repo, hook='bash_hook'))
+            assert 'appears to have no commits' in str(e.value)
+        else:
+            assert not try_repo(try_repo_opts(bare_repo, hook='bash_hook'))
```
asottile, line 132: a few rules of thumb to follow:
- don't use booleans in parametrize (this is a special type of [boolean trap](https://youtu.be/CnRkXO_a5mI)
- don't mix success and failure in the same test
- don't [write logic in tests](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html) -- your tests should be dead simple obvious what's going on and if you have to put `if` statements or loops in them then your tests likely aren't doing a good job at that

any of those -> these should be two disparate tests and not attempting to batch them together

tests/commands/try_repo_test.py
```diff
@@ -98,6 +101,37 @@ def test_try_repo_relative_path(cap_out, tempdir_factory):
         assert not try_repo(try_repo_opts(relative_repo, hook='bash_hook'))
 
 
+def test_try_repo_no_commits_no_git(cap_out, tempdir_factory):
+    repo = tempdir_factory.get()
+    with pytest.raises(CalledProcessError) as e:
+        try_repo(try_repo_opts(repo, hook='bash_hook'))
+    assert 'appears to have no commits' not in str(e.value)
```
asottile, line 108: I don't really see how this is a useful test -- if you're going to catch an exception you should assert what it actually outputs rather than what it _doesn't_ output -- there's an infinite number of things it doesn't output but ideally only one that it actually does

asottile:
given https://github.com/pre-commit/pre-commit/pull/3441#discussion_r2021957615 I think this is DOA

pre_commit/git.py
jaredkoontz: for this given command `git-ls-remote` i doubt that it will return 2 for anything else other than `when no matching refs are found in the remote repository`, given that it is explicitly called out in the [docs as this being the behavior](https://git-scm.com/docs/git-ls-remote#Documentation/git-ls-remote.txt---exit-code), but that is definitely an assumption. 

thanks for the feedback! I think the best thing is to just update the docs to make that phrase more explicit so it doesn't confuse anyone else.

https://github.com/pre-commit/pre-commit.com/pull/1020

PR #3446, opened 2025-04-13
Fix permission errors for mounts in rootless docker

matthewhughes934:
By running containers in a rootless docker context as root. This is because user and group IDs are remapped in the user namespaces uses by rootless docker, and it's unlikely that the current user ID will map to the same ID under this remap (see docs[1] for some more details). Specifically, it means ownership of mounted volumes will not be for the current user and trying to write can result in permission errors.

This change borrows heavily from an existing PR[2].

The output format of `docker system info` I don't think is documented/guaranteed anywhere, but it should corresponding to the format of a `/info` API request to Docker[3]

The added test _hopes_ to avoid regressions in this behaviour, but since tests aren't run in a rootless docker context on the PR checks (and I couldn't find an easy way to make it the case) there's still a risk of regressions sneaking in.

Link: https://docs.docker.com/engine/security/rootless/ [1]
Link: https://github.com/pre-commit/pre-commit/pull/1484/ [2]
Link: https://docs.docker.com/reference/api/engine/version/v1.48/#tag/System/operation/SystemAuth [3]


tests/languages/docker_test.py
```diff
@@ -62,6 +62,25 @@ def invalid_attribute():
...
+@pytest.mark.parametrize(
+    ('info_ret', 'expect_root'),
+    (
+        ((0, b'["name=rootless","name=cgroupns"]', b''), True),
+        ((0, b'["name=cgroupns"]', b''), False),
+        ((1, b'', b''), False),
+    ),
+)
+def test_docker_user_rootless_docker(info_ret, expect_root):
+    docker._is_rootless_docker.cache_clear()
```
matthewhughes934, line 74: :thinking: maybe we should clear this before _and_ after the test run: otherwise we risk polluting following tests with whatever value we last injected

asottile: usually a better approach is something like this:

```python
@pytest.fixture(autouse=True)
def _avoid_cache():
    with mock.patch.object(docker, '_is_rootless_docker', docker._is_rootless_docker.__wrapped__):
        yield
```

this bypasses the cache for the duration of the test

pre_commit/languages/docker.py
```diff
@@ -101,7 +102,23 @@ def install_environment(
...
+def _is_rootless_docker() -> bool:  # pragma: win32 no cover
+    retcode, out, _ = cmd_output_b(
+        'docker', 'system', 'info', '--format', '{{ json .SecurityOptions }}',
+    )
+    # some failures are to be expected, e.g. for 'podman' aliased as 'docker'
+    if retcode != 0:
+        return False
+
+    info = json.loads(out)
+    return any(opt == 'name=rootless' for opt in info)
```
asottile, line 115: this doesn't seem to work for podman :(

```console
$ readlink -f $(which docker)
/usr/bin/podman
$ docker system  info --format '{{ json .SecurityOptions }}'
Error: template: info:1:8: executing "info" at <.SecurityOptions>: can't evaluate field SecurityOptions in type *define.Info
```

tests/languages/docker_test.py
```diff
@@ -62,6 +62,25 @@ def invalid_attribute():
...
+    with mock.patch.object(
+            docker, 'cmd_output_b', return_value=info_ret,
+    ):
+        if expect_root:
+            assert docker.get_docker_user() == ()
+        else:
+            assert docker.get_docker_user() != ()
```
asottile, line 81: don't write logic in tests -- these are two separate disparate behaviours and should be tested separately

pre_commit/languages/docker.py
matthewhughes934: > this doesn't seem to work for podman :(
> 
> ```
> $ readlink -f $(which docker)
> /usr/bin/podman
> $ docker system  info --format '{{ json .SecurityOptions }}'
> Error: template: info:1:8: executing "info" at <.SecurityOptions>: can't evaluate field SecurityOptions in type *define.Info
> ```

The corresponding invocation for `podman` seems to be: `podman system info --format '{{ json .Host.Security.Rootless }}'`. 

> this doesn't seem to work for podman :(
> 
> ```
> $ readlink -f $(which docker)
> /usr/bin/podman
> $ docker system  info --format '{{ json .SecurityOptions }}'
> Error: template: info:1:8: executing "info" at <.SecurityOptions>: can't evaluate field SecurityOptions in type *define.Info
> ```

It looks like the invocation we need for `podman` is: `podman system info --format '{{ json .Host.Security.Rootless }}'`, I guess we could either:

* Run the docker command and on failure try the podman one, or
* Run a separate command to figure out if `docker` invokes docker or podman and then run the corresponding command (ignoring any other container engines for now)

tests/languages/docker_test.py
matthewhughes934: > don't write logic in tests -- these are two separate disparate behaviours and should be tested separately

[909c165](https://github.com/pre-commit/pre-commit/pull/3446/commits/909c165194672b5e998c6bbd608277f14f26b08e) also included your cache suggestion from above

pre_commit/languages/docker.py
matthewhughes934: Parsing the entire info response seemed the most robust: [9bc412d](https://github.com/pre-commit/pre-commit/pull/3446/commits/9bc412d0bc160b63fe3fba4700b2773f82204c0f)

Issue #3448, opened 2025-04-15
wrong lookup path scheme for `python` executable in MSYS2 (MinGW)


ByteB4rb1e:
### search you tried in the issue tracker

msys2

### describe your issue

In an MSYS2 (Cygwin/MinGW) environment with CPython compiled against Microsoft Visual C++ Runtime (MSVCRT) with MinGW, `pre-commit run` is unable to lookup the correct path of the Python executable. It is looking in `{base}/Scripts`, instead of `{base}/bin`. The mentioned setup prefers a POSIX path scheme, though it identifies as a `win32`, `nt` platform.

```
> cat /proc/version
MSYS_NT-10.0-26100 version 3.5.7-2644508f.x86_64 (runneradmin@fv-az1853-986) (gcc version 13.3.0 (GCC) ) 2025-01-30 09:08 UTC
>  pre-commit run --all-files --verbose --show-diff-on-failure
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: AssertionError: BUG: expected environment for python to be healthy immediately after install, please open an issue describing your environment

more info:

virtualenv python version did not match created version:
- actual version: <<error retrieving version from C:/Users/me/.cache/pre-commit/repoqgpcjyrq/py_env-python3.12/Scripts/python.exe>>
- expected version: 3.12.9.final.0

Check the log at C:/Users/me/.cache/pre-commit/pre-commit.log
```

In reference to issue [#1605](https://github.com/pre-commit/pre-commit/issues/1605), I require POSIX emulation.

### pre-commit --version

pre-commit 4.2.0

### .pre-commit-config.yaml

```yaml
exclude: '^(pipenv/patched/|pipenv/vendor/|tests/|pipenv/pipenv.1)'

repos:
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v4.6.0
  hooks:
  - id: check-builtin-literals
  - id: check-added-large-files
  - id: check-case-conflict
  - id: check-ast
  - id: check-toml
  - id: check-yaml
  - id: debug-statements
  - id: end-of-file-fixer
    exclude: WHEEL
  - id: forbid-new-submodules
  - id: trailing-whitespace
    exclude: .patch

- repo: https://github.com/astral-sh/ruff-pre-commit
  rev: v0.9.3
  hooks:
  - id: ruff
[...]

PR #3449, opened 2025-04-15
Refactor cross-platform `bin/` / `Scripts/` path lookup

ByteB4rb1e:
Cygwin isn't the only edge-case for a differing `bin/` path. In addition there also is MSYS2, which behaves differently than Cygwin when using the MinGW POSIX emulation layer. CPython compiled against Microsoft Visual C++ Runtime (MSVCRT) through MinGW, will correctly identify as an 'nt' platform, yet will prefer a POSIX path scheme. In order to avoid having to patch every edge-case, we can spy on the system configurations install paths through the `sysconfig` module and get the basenam of the global scripts path, which will be returned in accordance with the preferred install path scheme of the system/platform.

This is a fix for [issue #3448](https://github.com/pre-commit/pre-commit/issues/3448)

### DIFF
--- a/pre_commit/languages/conda.py
+++ b/pre_commit/languages/conda.py
@@ -12,6 +12,7 @@ from pre_commit.envcontext import PatchesT
 from pre_commit.envcontext import SubstitutionT
 from pre_commit.envcontext import UNSET
 from pre_commit.envcontext import Var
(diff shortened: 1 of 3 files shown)


asottile:
sorry but no thanks.  your platform combination is intentionally unsupported

PR #3473, opened 2025-05-30
improve Docker env detection heuristic

### LINKED ISSUE #3360: Pre-commit unable to determine if it is in a container from /proc/1/cgroup for Docker-in-Docker setups

is:issue docker is:open 

### describe your issue

Developing in a Fedora 40 container on a Fedora 41 host, using Docker-in-Docker such that pre-commit can make use of existing containers for lint tests. It appears that the contents of `/proc/1/cgroup` has changed:

```
$ cat /proc/1/cgroup
0::/
```

This is in contrast to pre-commit's checks for [_is_in_docker()](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/languages/docker.py#L22) and [_get_container_id](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/languages/docker.py#L32):

```
def _is_in_docker() -> bool:
    try:
        with open('/proc/1/cgroup', 'rb') as f:
            return b'docker' in f.read()
    ...

def _get_container_id() -> str:
    ...
    with open('/proc/1/cgroup', 'rb') as f:
        for line in f.readlines():
            if line.split(b':')[1] == b'cpuset':
                return os.path.basename(line.split(b':')[2]).strip().decode()
    ...
```

It's not clear yet what has changed with `/proc/1/cgroup` nor what the fix should be.

Host:
```
$ rpm -qa |grep docker
docker-ce-cli-27.3.1-1.fc41.x86_64
docker-ce-27.3.1-1.fc41.x86_64
docker-ce-rootless-extras-27.3.1-1.fc41.x86_64
docker-compose-plugin-2.29.7-1.fc41.x86_64
docker-buildx-plugin-0.17.1-1.fc41.x86_64
```

Guest:
```
$ rpm -qa |grep docker
docker-compose-plugin-2.29.7-1.fc40.x86_64
docker-buildx-plugin-0.17.1-1.fc40.x86_64
docker-ce-cli-27.3.1-1.fc40.x86_64
```

### pre-commit --version

pre-commit 3.8.0

### .pre-commit-config.yaml

```yaml
repos:
  - repo: local
    hooks:
      - id: shfmt
        name: shfmt
        # shfmt versions can be found here:
        #   https://hub.docker.com/r/mvdan/shfmt/tags
        entry: mvdan/shfmt:v3.8.0
        language: docker_image
        types: [shell]
        args: ["-w"]
```

_No response_

jouve:
With cgroupsv2, `/proc/1/cgroup` contains only `0::/`. 

Use `/.dockerenv` to detect Docker environment
and hostname to detect the current container id.

fix #3360

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -4,6 +4,8 @@ import functools
 import hashlib
 import json
 import os
+import re
+import socket
 from collections.abc import Sequence
 
 from pre_commit import lang_base
@@ -21,9 +23,11 @@ in_env = lang_base.no_env  # no special environment for docker
 def _is_in_docker() -> bool:
     try:
         with open('/proc/1/cgroup', 'rb') as f:
-            return b'docker' in f.read()
+            if b'docker' in f.read():
+                return True
     except FileNotFoundError:
-        return False
+        pass
+    return os.path.exists('/.dockerenv')
 
 
 def _get_container_id() -> str:
@@ -34,6 +38,9 @@ def _get_container_id() -> str:
         for line in f.readlines():
             if line.split(b':')[1] == b'cpuset':
                 return os.path.basename(line.split(b':')[2]).strip().decode()
+    hostname = socket.gethostname()
+    if re.match(r'^[0-9a-f]{12}$', hostname):
+        return hostname
     raise RuntimeError('Failed to find the container ID in /proc/1/cgroup.')
 
 
--- a/tests/languages/docker_test.py
+++ b/tests/languages/docker_test.py
@@ -30,8 +30,13 @@ DOCKER_CGROUP_EXAMPLE = b'''\
 0::/system.slice/containerd.service
 '''  # noqa: E501
 
+DOCKER_CGROUPV2_EXAMPLE = b"""\
+0::/
+"""
+
 # The ID should match the above cgroup example.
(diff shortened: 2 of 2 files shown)


pre_commit/languages/docker.py
```diff
@@ -34,6 +38,9 @@ def _get_container_id() -> str:
         for line in f.readlines():
             if line.split(b':')[1] == b'cpuset':
                 return os.path.basename(line.split(b':')[2]).strip().decode()
+    hostname = socket.gethostname()
+    if re.match(r'^[0-9a-f]{12}$', hostname):
```
asottile, line 42: this isn't reliable.  hostname can be set arbitrarily both inside and outside of docker 

a similar approach has already been rejected

jouve: yes, that why I tried to at least ensure it looks like some container ID.

if not in a container, then this code won't be executed, so not problem here.

if in a container, at best, the hostname is the container ID and it works which is an improvement.

at worst, the hostname is not a container ID, then it does not work, which is not worse than the actual behavior.

asottile: in case it's not clear I'm not going to accept this code for the same reasons I've rejected it in the past

jouve: that was not clear indeed !

PR #3481, opened 2025-06-07
fix: cargo install uses Cargo.lock for installation

hofbi:
According to the [cargo documentation](https://doc.rust-lang.org/cargo/commands/cargo-install.html#dealing-with-the-lockfile), cargo does not care about the lockfile by default but will recompute which versions of dependencies to use. This can lead to issues such as https://github.com/EmbarkStudios/cargo-deny/issues/771 where the release of a new upstream dependency breaks an existing unchanged pipeline.

Adding `--locked` to the cargo install to to force Cargo to use the packaged `Cargo.lock` file if it is available which should make rust hooks more reproducible.

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -155,6 +155,6 @@ def install_environment(
 
         for args in packages_to_install:
             cmd_output_b(
-                'cargo', 'install', '--bins', '--root', envdir, *args,
+                'cargo', 'install', '--locked', '--bins', '--root', envdir, *args,
                 cwd=prefix.prefix_dir,
             )
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -155,6 +155,7 @@ def install_environment(
 
         for args in packages_to_install:
             cmd_output_b(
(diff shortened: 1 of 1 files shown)


asottile:
there's a duplicate patch / proposal in progress already

hofbi:
I could not find anything related. The only open PR related to rust that I could find is #3235, which seems to be a new feature. Also, the last update was 10 months ago. Or is there anything you are working on which you did not push yet?

I am pretty sure, you understand the importance of lock files, so I wanted to make sure that this issue is fixed in time, being it with my PR or from someone else.

hofbi:
@asottile I think I was able to find the duplicate you are referring to which should be https://github.com/pre-commit/pre-commit/pull/3163 solving https://github.com/pre-commit/pre-commit/issues/3162. Now I am a bit confused why the solution proposed either by my PR or #3163 is not what you expect? What would be your proposal to solve #3162? I am happy to support here.

If you are in favor of https://github.com/pre-commit/pre-commit/pull/3235, the current solution would not add any additional challenges. And even with hard coding the locked attribute, I don't see any usecase where someone would not want to install the dependencies as in the lockfile.

PR #3484, opened 2025-06-18
Diagram visualization of pre-commit's codebase

ivanmilevtues:
This PR adds a Diagram visualization of the pre-commit's codebase. You can see how it will render here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/pre-commit/on_boarding.md

The diagram is designed to give high-level understanding of the full codebase. Then when a new contributor wants to join they usually start small - modifying just one sub-module. Having the diagram they can identify where the logic of what they want to improve is, then go one level deeper and again navigate the component of interest. Now they can start working as each component is linked with relevant source code files and the new person can edit them by roughly understanding how they fit in the bigger picture. Would love to hear what is your opinion on the matter!

I'd usually open a discussion first, but you don't have them enabled for this repo so I decided to go ahead and open a PR.

Any feedback is more than welcome!

Full transparency: we’re exploring this idea as a potential startup, but we’re still early and figuring out what’s actually useful to developers.

We’ve also released a free GitHub Action that keeps the diagrams automatically updated as the code changes, so there's no ongoing maintenance burden.

### DIFF
--- a/.codeboarding/Application_Core.md
+++ b/.codeboarding/Application_Core.md
@@ -1,0 +1,225 @@
+```mermaid
+
+graph LR
+
+    Application_Core["Application Core"]
+
+    Command_Modules["Command Modules"]
+
+    Store_Management["Store Management"]
+
+    Client_Utilities["Client Utilities"]
+
+    Git_Operations["Git Operations"]
+
+    Output_Logging["Output & Logging"]
+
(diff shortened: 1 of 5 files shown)


asottile:
nobody asked for this

please don't send spam PRs like this

PR #3509, opened 2025-08-11
feat: Adds just_warn option to allow commit when return value is not …

guionardo:
In some cases, I have tools returning values different of zero, but this is not an error. 

For example, in golang projects, I usually run `govulncheck` to check vulnerabilities. I need that the commit be allowed, but an warning should be shown to the user.

For that, I just added an option `just_warn` to .pre-commit-config.yaml file

```yaml
- repo: local
    hooks:
      - id: go-mod-tidy
        entry: go mod tidy
        language: system
        pass_filenames: false
        name: Ensures that the go.mod and go.sum files are in sync with the source code.
        fail_fast: true
      - id: go-vulncheck
        name: Ensures that no vulnerabilities are present in the codebase.
        entry: govulncheck ./...
        language: system
        pass_filenames: false
        just_warn: true
```


asottile:
please search the issue tracker and start with an issue next time

this has been rejected several times

adding warning noise, especially intentionally, is a negative user experience which will teach people to tune out important signal

PR #3528, opened 2025-08-25
Add fail-fast argument for run command

JulianMaurin:
Hello 👋

This is my first contribution to this repository. Please let me know if I missed any requirements. I’ll be glad to adjust.

Currently, fail-fast is enforced only via configuration. This pull request adds a `--fail-fast` argument to the `run` command, so it can be toggled at runtime (e.g. enabled in CI but disabled locally).


asottile (approved):
generally please create an issue to discuss first.  I'm not sure I see the value in this but it's simple enough and maybe could be useful.  it's just kinda weird that there's three slightly different ways to do this but shrugs

Issue #3529, opened 2025-08-25
Add the ability to customize the pre-commit logging level


rwarren:
### search you tried in the issue tracker

vscode unstaged

## Explanation

There is currently no way to adjust the logging level that I can see with pre-commit. This means that warning logs like this "_Unstaged files detected_" warning are _always_ output to stderr (at least in the extremely common case when a commit is done with unstaged files in the working dir):

https://github.com/pre-commit/pre-commit/blob/v4.3.0/pre_commit/staged_files_only.py#L70

This logging noise is of particular concern in IDEs like vscode, which unfortunately report the first observed line in the commit message in their dialog, like this (on Windows):

<img width="394" height="120" alt="Image" src="https://github.com/user-attachments/assets/aa1b3630-e9e9-4a62-b879-dec4be26c557" />

The fact that vscode does this is extremely unfortunate.  Even worse, the actual output from the commit hook is buried in a mostly hidden output window. Although this is unfortunate, it is an extremely common problem and it would be nice to be able to work around it by tuning pre-commit logging.

## Suggestion

An env var like `PRECOMMIT_LOGLEVEL` would be nice to have. All that it would need to do is something like this:

```python
logger = logging.getLogger("pre_commit")

lvl = os.getenv("PRECOMMIT_LOGLEVEL")
if lvl:
    try:
        logger.setLevel(int(lvl))
    except ValueError:
        logger.setLevel(logging.getLevelName(lvl.upper()))
```

Then someone can just set `PRECOMMIT_LOGLEVEL=ERROR` in their env, and avoid the logging noise, working around the problem with vscode (and other IDEs).

Even fancier would be if we could selectively mask out certain error messages with a regex, but that is probably a bit too much.

## Additional notes

ℹ️ This was actually already raised in #3467, but dismissed since there was no proposed solution or workaround.

### pre-commit --version

pre-commit 4.3.0

asottile:
there's an exact duplicate that you already found

PR #3530, opened 2025-08-25
Add configurable log level via PRECOMMIT_LOGLEVEL (#3529)

rwarren:
This PR makes the pre-commit logging output level configurable via a `PRECOMMIT_LOGLEVEL` env var.

Although this is generally useful, it also provides the world of vscode users a workaround for using pre-commit reliably inside the IDE (see #3529 for details, which has keywords to let others locate it).

### DIFF
--- a/pre_commit/logging_handler.py
+++ b/pre_commit/logging_handler.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import logging
+import os
 from collections.abc import Generator
 
 from pre_commit import color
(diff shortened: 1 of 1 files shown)


rwarren:
I see I missed tests... nice to see full coverage testing!!

However, I also see the driving issue (#3529) was closed (although the issue was not a true duplicate... just related so I linked it).  If the PR will be accepted I'll write test coverage. If not I won't waste my time. :)

If not accepted I'll continue with our monkey patching around this issue and post a workaround in the issue for future passers-by.

PR #3557, opened 2025-10-13
Bump `pre-commit-hooks` version in sample config to avoid warning

nathanjmcdougall:
The previous hard-coded version (v3.2.0) used stage names which are deprecated in the current version of `pre-commit`, and so emits a warning when used:

```
[WARNING] repo `https://github.com/pre-commit/pre-commit-hooks` uses deprecated stage names (commit, push) which will be removed in a future version.  Hint: often `pre-commit autoupdate --repo https://github.com/pre-commit/pre-commit-hooks` will fix this.  if it does not -- consider reporting an issue to that repo.
```

Bumping to `v6.0.0` will help avoid this warning being emitted and encourage the user to use a more up-to-date version if they use this sample config.


asottile:
please search for duplicates next time!

PR #3567, opened 2025-10-28
Cap hook output at terminal width to prevent wrapping

Zaczero:
Previously, _compute_cols() would calculate the ideal column width based on hook names and message lengths, with a minimum of 80 columns. However, it didn't consider the actual terminal width, causing the "Passed/Failed" status to wrap to the next line on narrower terminals.

This change:
- Adds terminal width detection using shutil.get_terminal_size()
- Caps the column width at the terminal width to prevent wrapping
- Maintains the minimum of 80 columns when terminal is wide enough
- Falls back to 80 columns if terminal size cannot be determined

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -6,6 +6,7 @@ import functools
 import logging
 import os
 import re
+import shutil
 import subprocess
 import time
 import unicodedata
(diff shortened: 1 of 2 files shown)


asottile:
look at the history of the function for a fun fact (and why I'm closing this)

Zaczero:
https://github.com/pre-commit/pre-commit/pull/408 I see. 

The idea was to make it follow the pytest style:



Comparing to the OG behavior, this PR only reduces max width and does not make it expand to full terminal width - that looks odd. Improving readability at small terminals.

asottile:
look I'm not going to argue with you.  next time please search the issue tracker

PR #3577, opened 2025-11-08
rename system and script languages to unsupported / unsupported_script

asottile:
to better represent that these "languages" aren't actually languages at all and do not provision environments.

a future version will show a deprecation warning and eventually the aliases will be removed!


qexat:
finally! :partying_face:

lightswitch05:
@asottile what is the recommendation for those of us using bash scripts here?

asottile:
what do you mean?

lightswitch05:
I've been using `language: script` in a few places for a couple custom lint checks. I see I can continue to upgrade using `unsupported_script`, but that is just a short term fix. What alternative 'approved' method should I be using to run my bash script?

asottile:
I think you're confused.  why do you think that's a short term fix?

lightswitch05:
Yeah, I guess I am confused. I think the long-term goal of the deprecations is to remove support for running bash scripts?

asottile:
why would you think that?

lightswitch05:
> a future version will show a deprecation warning and eventually the aliases will be removed!

Ok, I think I might be following along now. I think the above sentence is the source of my confusion.

The long-term goal isn't to remove `unsupported` and `unsupported_script`, but rather `script` & `system`. `unsupported` and `unsupported_script` will not be going anywhere. The change in naming is to better describe that dependencies need to be manage by the users and not something magically provided by pre-commit. 

Is that right? In that case, I just switch to `unsupported_script`, and I do not need to worry about any further deprecations and alias removals.

nikableh:
That's really confusing... What's the rationale behind renaming `system` to `unsupported`? It gives an impression that you are not supposed to use system executables and that this functionality is deprecated.

asottile:
that's the correct impression and exactly the intention.  it is the escape hatch and not a recommended way to use pre-commit

nikableh:
Why though? What if you need to execute something like `clang-format`? It is not shipped as a python package because it's not a python program at all.

asottile:
pre-commit is not just a tool for python tools. but also there is a clang-format: https://github.com/ssciwr/clang-format-wheel?tab=readme-ov-file#use-from-pre-commit

asottile:
also again it's not going away.  it's just making the naming more honest

nikableh:
> also there is a clang-format

If you actually look at the implementation of that hook, you can see that they still invoke a binary from PATH, but now also have to setup a dummy python package — <https://github.com/pre-commit/mirrors-clang-format>. This just looks like a very unnecessary extra step...

> also again it's not going away. it's just making the naming more honest

If anything, it feels like things are less honest now. People's first impression is that invoking binaries from PATH is "unsupported". Having a language called "system" was much more clear in its intentions.

Well, I guess is too late to argue about this now, since the PR is already merged :P

asottile:
the supported path is when pre-commit provisions the tools for you -- that's kind of the point.  the unsupported path is you need some random set of installation and setup instructions for your contributors to get every random tool set up and working

PR #3591, opened 2025-11-29
Fix xargs failure when os.sysconf is blocked in sandboxed environments

### LINKED ISSUE #3589: os.sysconf() call fails in highly restricted sandboxed environments (e.g., Cursor IDE)
### Searches I tried in the issue tracker

- os.sysconf
- SC_ARG_MAX

### Problem Description

The `pre-commit` tool is failing in highly secure, sandboxed execution environments, such as those used by the Cursor IDE agent. The sandbox intentionally blocks non-essential system calls for security reasons (like preventing sandbox escapes and host reconnaissance).

Specifically, the tool attempts to call `os.sysconf()` (or equivalent C library functions) to dynamically determine system limits, such as the maximum argument size (`_SC_ARG_MAX`). This system call is blocked by the sandbox, causing the hook execution to fail silently or with an unhandled exception in environments where system calls are filtered.

### Error message from cursor chat

```text
I can’t reliably run git commit for you here:
Your pre-commit hook calls os.sysconf('SC_ARG_MAX'), which is blocked in this sandbox, causing pre-commit (and thus git commit) to fail.
Please run this in your own shell (Cursor terminal / iTerm, etc.), where pre-commit can run normally:
```

### Suggested Solution

To make `pre-commit` more robust in restricted environments, please consider adding a fallback mechanism to use a reasonable, hardcoded POSIX default (e.g., `4096` bytes) for maximum argument size if the dynamic `os.sysconf()` call fails or is unavailable. This would allow the tool to function without violating strict security models.

See [_POSIX_SC_MAX](https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_chapter/libc_31.html#SEC648:~:text=the%20notifications%20asynchronously.-,_POSIX_ARG_MAX,-The%20value%20of) (4096) as an safe mimum if os.gesysconf('SC_ARG_MAX') is blocked.

### Sample config file

`.pre-commit-config.yaml`

````yaml
repos:
  - repo: https://github.com/igorshubovych/markdownlint-cli
    rev: v0.39.0
    hooks:
      - id: markdownlint-fix
````

### pre-commit --version

pre-commit 4.3.0

MichaelRWolf:
Fixes #3589

This PR fixes a bug where `pre_commit.xargs` fails in highly restricted sandboxed environments (like Cursor-IDE's AI-tool sandbox) when `os.sysconf('SC_ARG_MAX')` is blocked. This can manifest as import-time failures or runtime failures depending on when the sysconf call occurs.

**Changes:**
- Made `_max_length` parameter lazy-evaluated (defaults to `None`, computed at runtime)
- Added `try/except OSError` in `_get_platform_max_length()` to handle blocked sysconf calls
- Falls back to POSIX minimum (4096) when sysconf is unavailable
- Added comprehensive tests using subprocess approach to verify fix works at both import-time and runtime

**Testing:**
- All existing tests pass
- New tests verify import-time and runtime behavior when sysconf is blocked

### DIFF
--- a/tests/xargs_with_blocked_sysconf_test.py
+++ b/tests/xargs_with_blocked_sysconf_test.py
@@ -171,4 +171,3 @@ def test_xargs_with_explicit_max_length_when_sysconf_blocked_as_subprocess() -> 
         f'os.sysconf is blocked. This should always work since _get_platform_max_length() '
         f'is never called. stderr: {proc.stderr!r}'
     )
-


asottile:
I'm not really convinced this is a good idea.  there's literally zero reason to block read only access to sysconf.  I think cursor should fix their broken setup instead

not to mention if sysconf is blocked I can't imagine what else is blocked and I don't expect pre-commit to get any further

plus I can't accept ai generated patches and it's super clear this was not written by you

MichaelRWolf:
Thanks for your swift response.  I had hoped for a few days (or weeks), but got a response in a few _minutes_.  Thanks.

This was my largest attempt at patching such a public repo, so I'd appreciate some assistance to understand what I cold have done differently to propose a change that would have worked for you.

BTW.... I invested time in this patch because I found pre-commit hooks to be so helpful.   I was at the point of having to decide whether to drop pre-commit or Cursor.  It was tough.  I wanted to use them  BOTH.  I've been involved with open source for decades.  I figured I could make a difference here much easier than with OpenAI.  ;-}

Since the issue was triggered by a single line in pre-commit, I started there.

It worked.  The proposed change allows me to continue using pre-commit in a repo that has code I've been maintaing for about 20+ years -- my personal profile that works across many systems.

I can now ask Cursor to "analyze and commit", and it will create a commit message for staged changes, then commit (including pre-commit hooks).

I was actually doing a happy dance to celebrate this win when I received your comments.

Let me see if I can address each of them...

> I'm not really convinced this is a good idea. there's literally zero reason to block read only access to sysconf. I think cursor should fix their broken setup instead

I can't speak to their reasons for blocking `sysconf`, but my research convinced me that lead me to believe that POSIX_MAX_ARG is a safe default for `os.sysconf('SC_ARG_MAX')`, both at load-time and also at run-time.  Do you agree?  



> not to mention if sysconf is blocked I can't imagine what else is blocked and I don't expect pre-commit to get any further


I guess I don't understand what this is pointing at.  I only researched the problems that arose with os.sysconf().  There was only one 
             os.sysconf('SC_ARG_MAX')

Once this was overcome, pre-commit got much further.  All the way..  It completed beautifully.


> plus I can't accept ai generated patches and it's super clear this was not written by you

This PR was my work.  Human research.  Human review.  Human testing.  Human experience.

AI was my assistant, not the leader.  This was NOT vibe coding.

I started with background research.  Then we generated a failing test that would showcase the issue.  Only then did I modify the original code.  

Generating the code was actually the easy part.  I could have done that in significantly less time, but I wanted the test (and the understanding behind it) to give me the confidence that it was a robust change.  I'm happy to have had AI to assist me.  I'm especially grateful that it pointed out a problem with my testing strategy allowed `conftest.py` that pevented `pytest` from even running my tests.

MichaelRWolf:
TL;DR; -- I submitted this work because I value what pre-commit does for me.  It was my work.  Human work.  Human experience.  Human caring.  

It fixed my work-flow blockage.  I think it could be helpful to othersj, too.

PR #3601, opened 2025-12-24
perf: use git cat-file -e instead of git rev-list in _rev_exists

ptarjan:
Replace `git rev-list --quiet` with `git cat-file -e` for checking if a revision exists. This changes the operation from O(commits) to O(1).

Benchmarks (repo with ~1M commits, 5 runs each):

* git rev-list --quiet: 5159-5191ms
* git cat-file -e: 2-4ms

Both commands have the same semantics: return success if the revision exists in the local repo, failure otherwise.

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -111,7 +111,7 @@ def _ns(
 
 
 def _rev_exists(rev: str) -> bool:
-    return not subprocess.call(('git', 'rev-list', '--quiet', rev))
+    return subprocess.call(('git', 'cat-file', '-e', rev)) == 0
 
 
 def _pre_push_ns(
--- a/tests/commands/hook_impl_test.py
+++ b/tests/commands/hook_impl_test.py
@@ -91,6 +91,22 @@ def test_run_legacy_recursive(tmpdir):
             call()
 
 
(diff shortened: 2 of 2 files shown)


asottile:
sorry but I unfortunately can't accept ai submissions

asottile:
this piece of code also isn't performance critical -- are you solving an actual problem?

ptarjan:
> sorry but I unfortunately can't accept ai submissions

Would you prefer me to set the author to my account?

>  this piece of code also isn't performance critical -- are you solving an actual problem?

Absolutely. As per the summary, in our repo our users are seeing a large delay with our pre-push hooks. So much so that about 7% of users are skipping them entirely. I'm trying to reduce the performance footprint and this was one of the largest slowdowns.

ptarjan:
I updated the commit metadata to my name and username.

asottile:
> I updated the commit metadata to my name and username.

that's just being disingenuous then?  you didn't author this

ptarjan:
> that's just being disingenuous then? you didn't author this

I'm not sure what you mean by "author"? I mean I didn't open `vim` to make the changes to the text file, but I did instruct my terminal to "change git rev-list --quiet to git cat-file -e".

But I'd prefer not to split hairs if possible. What would you like me to do now so that we can accomplish this performance improvement?

PR #3618, opened 2026-02-09
implemented --tool flag to allow arbirary arguments to be passed

rmawatson:
We are using pre-commit as part of our python CI pipeline and often find we would like to use the really nice dependency management of pre-commit to include tools in various projects and invoke them in the terminal with arbitrary arguments.

Most of our tools are performing pure validation in the CI (we have a `--validate` flag we use with pre-commit) , but these tools expose other features useful to the developer.

Currently pre-commit only allows usage of the hook as a 'tool' in the terminal with 

`pre-commit run example-hook`

If we wanted different arguments passed to different invocations this requires defining each permutation as a separate entry in the `.pre-commit.config.yaml` with specific predefined arguments in `args:[...]`, which does not support using the hook as a tool with dynamic arguments.

Without having to have yet another solution to use our hooks as tools in the terminal, given pre-commit already provides 99% of the functionality we have added a `--tool` flag to our internal build of pre-commit.

`--tool` implies `allow_all_files=true` (this is a choice for our use cases but would work either way) and requires the hook is exposed for `stages:["manual"]`.

A hook can be invoked as a tool with any arbitrary arguments using:

`pre-commit run example-hook --tool -- --arg1 --arg2=4 ...`

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -147,6 +147,8 @@ def _run_single_hook(
         diff_before: bytes,
         verbose: bool,
         use_color: bool,
+        is_tool: bool = False,
+        extra_args: Sequence[str] = (),
 ) -> tuple[bool, bytes]:
     filenames = tuple(classifier.filenames_for_hook(hook))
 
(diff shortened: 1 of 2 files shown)


asottile:
there are quite a few duplicates in the issue tracker and rejection reasons already

PR #3623, opened 2026-02-19
feat: add pass_filenames_via_stdin for large changesets

sharmila-oai:
Today, pre-commit passes filenames via argv, which can hit command-length limits on large change sets and cause chunking into multiple hook invocations. Some hooks operate better with the full file list in one shot, but currently the only workaround is disabling filename passing and re-running custom git queries, which is slower and duplicates logic.

I added a flag `pass_filenames_via_stdin`: which solves this by sending the full filename set once (NUL-delimited) over stdin, avoiding chunking/re-runs

Why NUL-delimited stdin:
- safe for filenames containing spaces/newlines
- matches established -0 conventions in unix tooling

**Usage examples:**

- shell:

```shell
    while IFS= read -r -d '' filename; do
        ...
    done
```
- python:

```python
    data = sys.stdin.buffer.read()
    filenames = [os.fsdecode(p) for p in data.split(b'\0') if p]
```

**Implementation notes:**
- default is the same argv-based passing of files
- `pass_filenames`: false still disables filename passing entirely

Implementation includes schema/runtime wiring, shared NUL encode/decode helpers, and tests covering defaulting and runtime behavior.

### DIFF
--- a/tests/languages/pygrep_test.py
+++ b/tests/languages/pygrep_test.py
@@ -143,7 +143,9 @@ def test_grep_hook_matching(some_files, tmp_path):
 
 @pytest.mark.usefixtures('some_files')
 def test_main_reads_nul_delimited_filenames_from_stdin(cap_out):
-    with mock.patch.object(sys.stdin.buffer, 'read', return_value=b'f1\x00f2\x00'):
+    with mock.patch.object(
+        sys.stdin.buffer, 'read', return_value=b'f1\x00f2\x00',
+    ):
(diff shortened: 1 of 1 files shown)


elliot-google (approved):

asottile:
there's several duplicate issues rejecting this feature

ags-openai:
What if we say pretty please? 🙃

asottile:
lol hope you're doing well friend

ags-openai:
Ditto!

PR #3624, opened 2026-02-20
Remove a full stop in "Initializing environment" message

mathbunnyru:
This way links will work properly in GitHub UI / terminal, as dot won't be part of the link.


asottile:
there's a duplicate please search next time!

PR #3640, opened 2026-03-23
Delete .github/actions/pre-test/action.yml

mxr:
I think this is unneeded because the repo is py310+ ?


asottile:
ah actually the condition is just wrong.  we do actually want to test against the latest git version

mxr:
sg, for which python versions?

asottile:
whatever the minimum version is was the intention

PR #3656, opened 2026-04-11
Validate log_file paths to prevent arbitrary file writes

### LINKED ISSUE #3655: Remote hook manifests can write to arbitrary host paths via log_file
## Summary

`pre-commit` accepts the `log_file` field from remote hook manifests (`.pre-commit-hooks.yaml`) and later opens that path directly on the host filesystem in append mode when a hook produces output. Because the path is neither constrained to the repository root nor stripped for non-local manifests, a malicious hook repository can append attacker-controlled content to arbitrary relative or absolute host paths.

This is a tool-level trust-boundary issue — not simply "hooks execute code by design." The validated proof of concept uses `language: fail`, so **no hook code from the remote repository is executed**. The file write comes purely from manifest metadata.

**Severity:** Medium

### Manifest schema accepts `log_file` for remote repos

`pre_commit/clientlib.py`, line 261:

```python
MANIFEST_HOOK_DICT = cfgv.Map(
    'Hook', 'id',
    # ...
    cfgv.Optional('log_file', cfgv.check_string, ''),
    # ...
)
```

`log_file` is permitted in `MANIFEST_HOOK_DICT` (the schema used for `.pre-commit-hooks.yaml` in remote repos), with no path validation.

### Hook output is written to the manifest-controlled path

`pre_commit/commands/run.py`, line 232:

```python
if out.strip():
    output.write_line()
    output.write_line_b(out.strip(), logfile_name=hook.log_file)
    output.write_line()
```

### The path is opened directly in append mode

`pre_commit/output.py`, lines 14–29:

```python
def write_line_b(
        s: bytes | None = None,
        stream: IO[bytes] = sys.stdout.buffer,
        logfile_name: str | None = None,
) -> None:
    with contextlib.ExitStack() as exit_stack:
        output_streams = [stream]
        if logfile_name:
            stream = exit_stack.enter_context(open(logfile_name, 'ab'))
            output_streams.append(stream)
        # ...
```

No validation is performed on `logfile_name` — it is opened as-is with `open(logfile_name, 'ab')`.

## Steps to Reproduce

1. Create a malicious hook repository with `.pre-commit-hooks.yaml`:

```yaml
- id: evil-hook
  name: evil hook
  entry: |
    attacker-controlled content here
    this gets written to the target file
  language: fail
  files: ''
  always_run: true
  log_file: ../escape-to-parent.txt
```

2. Create a victim repository that uses this hook in `.pre-commit-config.yaml`:

```yaml
repos:
  - repo: https://github.com/<attacker>/malicious-hooks
    rev: v1.0.0
    hooks:
      - id: evil-hook
```

3. Run `pre-commit run --all-files` in the victim repository.

4. Observe that `../escape-to-parent.txt` is created one directory above the victim repo, containing the attacker-controlled `entry` string.

### Validation

I validated this in an isolated Docker environment:

- Created a local "malicious" hook repo with `log_file: ../manifest-log-file-write.txt` and `language: fail`
- Ran `pre-commit` from a separate victim repo directory
- Confirmed that `manifest-log-file-write.txt` was created **one directory above** the victim repo
- The file content was the attacker-controlled text from the hook's `entry` field

## Impact

A malicious hook repository can:

- **Write outside the repository** using relative traversal (`../target`) or absolute paths
- **Create new files** on the host filesystem
- **Append attacker-controlled bytes** to existing files
- The content is controlled through the manifest `entry` field, so the attacker has full control over what is written

This is achieved without any hook code execution — `language: fail` hooks never run an interpreter. The write primitive comes purely from manifest metadata being trusted as a file path.

## Suggested Fix

The least disruptive fix would be to make `log_file` config-only, i.e., only honor it from the user's `.pre-commit-config.yaml` and ignore it from remote manifests (`.pre-commit-hooks.yaml`).

Alternatively:

1. Remove `log_file` from `MANIFEST_HOOK_DICT` entirely (only keep it in the config schema).
2. If manifest support must remain, normalize the path against an allowed base directory and reject:
   - Absolute paths
   - `..` traversal components
   - Symlink escapes outside the repository root

eddieran:
## Summary

`log_file` in hook manifests (`.pre-commit-hooks.yaml`) accepts arbitrary strings and is passed directly to `open(logfile_name, 'ab')` in `output.write_line_b`. A malicious remote hook manifest can set `log_file` to an absolute path (e.g. `/etc/cron.d/malicious`) or use path traversal (`../../../etc/passwd`) to write hook output to arbitrary locations on the host filesystem.

## Fix

Add path validation for `log_file` in `clientlib.py` during manifest schema validation:
- Reject absolute paths
- Reject paths that normalize to a parent directory traversal (e.g. `../foo`)
- Empty string (the default) is still allowed

The validation is applied via `check_log_file` on the `MANIFEST_HOOK_DICT` schema, which also propagates to `CONFIG_HOOK_DICT` and `LOCAL_HOOK_DICT`.

## Test plan

- Added `test_check_log_file_valid` — verifies relative paths like `output.log`, `logs/hook.log`, and empty string pass validation
- Added `test_check_log_file_invalid` — verifies absolute paths and `..` traversal paths are rejected with `cfgv.ValidationError`

Fixes #3655

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -23,6 +23,22 @@ logger = logging.getLogger('pre_commit')
 
 check_string_regex = cfgv.check_and(cfgv.check_string, cfgv.check_regex)
 
+
+def _check_log_file(val: str) -> None:
+    if val == '':
+        return
+    if os.path.isabs(val):
+        raise cfgv.ValidationError(
+            f'log_file must be a relative path, got absolute path: {val!r}',
+        )
+    if os.path.normpath(val).startswith('..'):
+        raise cfgv.ValidationError(
+            f'log_file must not reference a parent directory: {val!r}',
+        )
+
+
+check_log_file = cfgv.check_and(cfgv.check_string, _check_log_file)
+
 HOOK_TYPES = (
     'commit-msg',
     'post-checkout',
@@ -258,7 +274,7 @@ MANIFEST_HOOK_DICT = cfgv.Map(
     cfgv.Optional('pass_filenames', cfgv.check_bool, True),
     cfgv.Optional('description', cfgv.check_string, ''),
     cfgv.Optional('language_version', cfgv.check_string, C.DEFAULT),
-    cfgv.Optional('log_file', cfgv.check_string, ''),
+    cfgv.Optional('log_file', check_log_file, ''),
     cfgv.Optional('require_serial', cfgv.check_bool, False),
     StagesMigration('stages', []),
     cfgv.Optional('verbose', cfgv.check_bool, False),
--- a/tests/clientlib_test.py
+++ b/tests/clientlib_test.py
@@ -7,6 +7,7 @@ import cfgv
 import pytest
 
 import pre_commit.constants as C
+from pre_commit.clientlib import _check_log_file
 from pre_commit.clientlib import check_type_tag
 from pre_commit.clientlib import CONFIG_HOOK_DICT
 from pre_commit.clientlib import CONFIG_REPO_DICT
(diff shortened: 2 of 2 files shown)


asottile:
there's no such thing as a "malicious remote hook" every part of pre-commit is opt in

please don't send ai slop prs

asottile:
also lol if they're truly evil they ALREADY HAVE ARBITRARY CODE EXECUTION

Issue #2530, opened 2022-09-28
ValueError on Windows when config is on a different drive than the git repo


SitiSchu:
### search tried in the issue tracker

is:issue relpath

### describe your issue

pre-commit will fail with `ValueError: path is on mount 'C:', start on mount 'D:'` when the config file is on `C:\` but the git repository is on `D:\`

Steps to reproduce:
- Create config file on your `C:` drive (for example `%USERPROFILE%\.pre-commit-config.yaml`)
- Create repo on `D:` drive
- Run pre-commit: `pre-commit run --config=%USERPROFILE%\.pre-commit-config.yaml`

Commenting out line 164 in `main.py` (`args.config = os.path.relpath(args.config)`) fixes this issue. Is it necessary that the config path is relative to the working directory?

### pre-commit --version

pre-commit 2.20.0

### .pre-commit-config.yaml

```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.3.0
    hooks:
    -   id: trailing-whitespace
```

### version information

```
pre-commit version: 2.20.0
git --version: git version 2.36.1
sys.version:
    3.9.9 (tags/v3.9.9:ccb0e6a, Nov 15 2021, 18:08:50) [MSC v.1929 64 bit (AMD64)]
sys.executable: D:\temp\prctest\venv\Scripts\python.exe
os.name: nt
sys.platform: win32
```

### error information

```
An unexpected error has occurred: ValueError: path is on mount 'C:', start on mount 'D:'
```

```
Traceback (most recent call last):
  File "D:\temp\prctest\venv\lib\site-packages\pre_commit\error_handler.py", line 73, in error_handler
    yield
  File "D:\temp\prctest\venv\lib\site-packages\pre_commit\main.py", line 343, in main
    _adjust_args_and_chdir(args)
  File "D:\temp\prctest\venv\lib\site-packages\pre_commit\main.py", line 164, in _adjust_args_and_chdir
    args.config = os.path.relpath(args.config)
  File "D:\local_software\PYTHON\3.9.9\Windows_NT.x64\lib\ntpath.py", line 703, in relpath
    raise ValueError("path is on mount %r, start on mount %r" % (
ValueError: path is on mount 'C:', start on mount 'D:'
```

asottile:
it's not necessary that it's a relative path, but it does make error messages better

feel free to send a patch

asottile:
[...]

PR #3658, opened 2026-04-16
fix: handle cross-drive ValueError in _adjust_args_and_chdir on Windows

terminalchai:
## Summary

Fixes #2530

On Windows, os.path.relpath() raises ValueError when the path and the current working directory are on different drives — e.g. a config at C:\Users\me\.pre-commit-config.yaml while the git repo lives on D:\. This causes an unhandled traceback instead of a helpful error message.

## Root cause

_adjust_args_and_chdir calls os.path.relpath() unconditionally on rgs.config, rgs.files, rgs.commit_msg_filename, and rgs.repo after doing os.chdir(toplevel). When the git root and the path being relativised are on different Windows drives, Python raises:

\\\
ValueError: path is on mount 'C:', start on mount 'D:'
\\\

## Fix

Introduce a small _try_relpath() helper (as suggested by @asottile in #2530 — option 2):

\\\python
def _try_relpath(path: str) -> str:
    try:
        return os.path.relpath(path)
    except ValueError:
        return path
\\\

All four os.path.relpath calls in _adjust_args_and_chdir are updated to use _try_relpath, so every cross-drive combination (config, files list, commit-msg filename, try-repo repo path) is covered. On POSIX and same-drive Windows the behaviour is identical to before.

## Tests added

- 	est_try_relpath_returns_relpath_when_same_drive — normal case, same drive
- 	est_try_relpath_returns_original_on_cross_drive_valueerror — mocks os.path.relpath to raise ValueError, asserts original path is returned unchanged
- 	est_adjust_args_config_on_different_drive — Windows-only integration test for the full _adjust_args_and_chdir path

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -172,6 +172,16 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
     )
 
 
+def _try_relpath(path: str) -> str:
+    """Return os.path.relpath(path), or path unchanged on Windows when path
+    and the current working directory are on different drives (which would
+    raise ValueError: path is on mount 'C:', start on mount 'D:')."""
+    try:
+        return os.path.relpath(path)
+    except ValueError:
+        return path
+
+
 def _adjust_args_and_chdir(args: argparse.Namespace) -> None:
     # `--config` was specified relative to the non-root working directory
     if os.path.exists(args.config):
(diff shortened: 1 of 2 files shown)


asottile:
sorry we don't accept ai slop prs

PR #3659, opened 2026-04-17
fix: handle ValueError from relpath on Windows cross-drive paths

bobo-xxx:
## Description

When the config file (or other paths) is on a different Windows drive than the git repository, `os.path.relpath` raises a `ValueError` because it cannot compute a relative path between paths on different drives.

This change wraps all `relpath` calls in `_adjust_args_and_chdir` with `ValueError` suppression, keeping the absolute path when `relpath` fails.

Fixes pre-commit/pre-commit#2530

## Testing

Added a regression test `test_adjust_args_and_chdir_relpath_valueerror_cross_drive` that simulates the cross-drive `ValueError` scenario.

All existing tests pass (29 passed, 1 skipped on Linux).

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+import contextlib
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


asottile:
ai slop is not acceptable

PR #3660, opened 2026-04-19
Fix ValueError on Windows when config is on different drive than git repo

avasis-ai:
Fixes #2530

On Windows, `os.path.relpath()` raises `ValueError` when the target path and the current working directory are on different drives (e.g., config on `C:` and repo on `D:`). This wraps all `os.path.relpath()` calls in `_adjust_args_and_chdir` with `try/except ValueError`, falling back to the absolute path when conversion is not possible.

As discussed in the issue, option 2 (try/except) was preferred by @asottile.


asottile:
ai slop is not acceptable

PR #3663, opened 2026-04-22
fix: handle ValueError when config path is on a different drive (Windows)

armorbreak001:
When `--config` points to a file on a different Windows drive than the
git repository (e.g. config at `C:\Users\...\.pre-commit-config.yaml`,
repo on `D:\`), `os.path.relpath` raises `ValueError: path is on mount
"C:", start on mount "D:"`.

Wrap the `relpath` call in a try/except so we fall back to using the
absolute path, which was already resolved earlier via `os.path.abspath`.
The absolute path works fine as a fallback.

**Before:** crash with unhandled ValueError
**After:** runs normally using absolute config path

Fixes #2530

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -188,7 +188,11 @@ def _adjust_args_and_chdir(args: argparse.Namespace) -> None:
     toplevel = git.get_root()
     os.chdir(toplevel)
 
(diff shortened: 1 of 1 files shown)


asottile:
ai slop

Issue #3664, opened 2026-04-24
Z40 = '0' * 40 hardcoding breaks `git push --delete` on SHA-256 repos


rlichtenwalter:
### search you tried in the issue tracker

`sha256`, `SHA-256 push`, `sha-256 delete`, `Z40`, `object-format sha256`, `"Invalid revision range"`, `"bad object"`, `zero SHA`, `pre-push delete`, `deletion push`

The closest related hit is #3434 (closed), but that is about fetching sha1-format hook repos into a sha256 local repo — a different codepath.

### describe your issue

`pre_commit/commands/hook_impl.py` hardcodes the SHA-1 zero OID:

```python
Z40 = '0' * 40
```

It is compared against `local_sha` / `remote_sha` to detect deletion-only pushes and new branches:

```python
if local_sha == Z40:
    continue
elif remote_sha != Z40 and _rev_exists(remote_sha):
    ...
```

In a repository initialized with `--object-format=sha256`, git's pre-push hook protocol emits 64-character zero OIDs rather than 40. The equality never matches, the deletion-only short-circuit never fires, and pre-commit falls through into `git diff --name-only ... <old>..<64-zeros>` (and then `<old>...<64-zeros>`). Both error out:

```
fatal: Invalid revision range 3dac4c61...6dce8d..000000...(×64)
fatal: Invalid symmetric difference expression 3dac4c61...6dce8d...000000...(×64)
```

pre-commit surfaces these as `An unexpected error has occurred: CalledProcessError` and exits non-zero, so `git push origin --delete <branch>` is refused on every sha256 repo. Branch deletion has to go through the forge's web UI or API to succeed.

Minimal reproducer:

```bash
git init --object-format=sha256 repro
cd repro
git config user.email t@e.com && git config user.name T
git commit --allow-empty -m init
cat > .pre-commit-config.yaml <<'EOF'
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v5.0.0
    hooks:
      - id: trailing-whitespace
EOF
git add .pre-commit-config.yaml && git commit -m "pc"
pre-commit install --hook-type pre-push

git checkout -b doomed
git commit --allow-empty -m x
SHA=$(git rev-parse HEAD)
ZEROS=$(printf '0%.0s' $(seq 1 64))

# Feed git's pre-push stdin format directly — simulates `git push origin --delete doomed`
[...]

PR #3665, opened 2026-04-24
fix: handle SHA-256 zero OID in pre-push hook

Jah-yee:
## Summary

Fix for issue #3664: Z40 hardcoding breaks `git push --delete` on SHA-256 repos.

### Problem

In repositories initialized with `--object-format=sha256`, git's pre-push hook protocol emits 64-character zero OIDs rather than the SHA-1 40-character `Z40`. This caused equality checks `local_sha == Z40` and `remote_sha != Z40` to never match, resulting in:

```
fatal: Invalid revision range ...000000...(x64)
fatal: Invalid symmetric difference expression ...000000...(x64)
```

Branch deletion (`git push origin --delete <branch>`) was refused on every sha256 repo.

### Solution

1. Added `ZERO_SHA_256 = '0' * 64` constant
2. Added `_get_object_format(remote_url)` function that detects SHA-256 via `git ls-remote --symref` reading the `object-format=` from HEAD symref
3. Modified `_pre_push_ns` to detect the remote's object format and use the appropriate zero SHA

### References

- Issue: pre-commit/pre-commit#3664

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -34,7 +34,6 @@ def _get_object_format(remote_url: str) -> str:
     return 'sha1'
 
 
-
 def _run_legacy(
         hook_type: str,
         hook_dir: str | None,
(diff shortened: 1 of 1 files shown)


Jah-yee:
Linking to issue #2530. The maintainer had indicated preference for option 2 (try/except approach). This PR implements that approach. The fix adds a _safe_relpath helper that wraps os.path.relpath in try/except, falling back to os.path.abspath when ValueError is raised (cross-drive on Windows).

PR #3666, opened 2026-04-24
fix: handle ValueError from os.path.relpath on Windows cross-drive paths

Jah-yee:
## Summary

Fixes ValueError when pre-commit config is on a different drive from the git repo on Windows (issue #2530).

## Root Cause

`os.path.relpath()` raises `ValueError` when the two paths are on different drives. The `_adjust_args_and_chdir` function in `main.py` unconditionally converts paths to relative, which fails on Windows cross-drive scenarios.

## Fix

Wrap `os.path.relpath` in a try/except inside `_adjust_args_and_chdir`. When `ValueError` is raised (cross-drive), fall back to `os.path.abspath` instead. This preserves functionality while fixing the crash.

```python
def _safe_relpath(path: str) -> str:
    try:
        return os.path.relpath(path)
    except ValueError:
        # On Windows, relpath raises ValueError when paths are on
        # different drives (e.g. config on C:, repo on D:). Fall back
        # to abspath so pre-commit still functions.
        return os.path.abspath(path)
```

## References

- Fixes pre-commit/pre-commit#2530
- Maintainer (@asottile) preference: option 2 (try/except approach)

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -188,15 +188,24 @@ def _adjust_args_and_chdir(args: argparse.Namespace) -> None:
     toplevel = git.get_root()
     os.chdir(toplevel)
 
-    args.config = os.path.relpath(args.config)
+    def _safe_relpath(path: str) -> str:
+        try:
+            return os.path.relpath(path)
+        except ValueError:
+            # On Windows, relpath raises ValueError when paths are on
+            # different drives (e.g. config on C:, repo on D:). Fall back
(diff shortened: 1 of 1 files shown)


Jah-yee:
Thank you for reviewing this PR. This implements the try/except approach for issue #2530 as suggested by @asottile. The fix wraps os.path.relpath and falls back to os.path.abspath when ValueError is raised on cross-drive paths in Windows.

asottile:
ai slop is not allowed

PR #3670, opened 2026-04-25
commands/run: skip staged files check for pre-push hooks (fixes #2486)

### LINKED ISSUE #2486: dotnet format pre-push hook allow push when correct file is in staged changes

Jah-yee:
## Summary
Fix pre-push hooks to not operate on staged files and not stash working tree changes.

**Problem (issue #2486):** When pre-push runs with staged files (e.g. `git commit -a` followed by `git push`), the `staged_files_only` context manager stashes ALL unstaged changes and then runs pre-push file-based hooks on the staged files instead of the pushed files. This causes:
1. Pre-push hooks to operate on the wrong set of files (staged vs pushed).
2. Unstaged changes to be stashed during pre-push, which can interfere with the push operation.
3. The warning about "unstaged changes" even when those changes are unrelated to the push.

**Fix:**
1. Add `pre-push` to the early-return list in `_all_filenames()` so file-based hooks don't try to operate on staged files.
2. Skip `staged_files_only` context for `pre-push` since the stash behavior is designed for commit hooks, not push hooks.

Per @asottile's comment: the "unstaged changes" check only makes sense for `commit` stage — this fix ensures it only applies there.

## Testing
`python -m py_compile` passes. CI tests will validate the behavior.

Fixes #2486

### DIFF
--- a/pre_commit/commands/run.py
+++ b/pre_commit/commands/run.py
@@ -254,7 +254,7 @@ def _all_filenames(args: argparse.Namespace) -> Iterable[str]:
     # these hooks do not operate on files
     if args.hook_stage in {
         'post-checkout', 'post-commit', 'post-merge', 'post-rewrite',
-        'pre-rebase',
+        'pre-rebase', 'pre-push',
     }:
         return ()
     elif args.hook_stage in {'prepare-commit-msg', 'commit-msg'}:
@@ -417,7 +417,7 @@ def run(
     environ['PRE_COMMIT'] = '1'
 
     with contextlib.ExitStack() as exit_stack:
-        if stash:
+        if stash and args.hook_stage != 'pre-push':
             exit_stack.enter_context(staged_files_only(store.directory))
 
         config = load_config(config_file)


asottile:
ai slop is not acceptable

PR #3671, opened 2026-04-27
Handle SHA-256 zero OIDs in pre-push

ShipItAndPray:
## Summary
- treat any all-zero object ID as the null push sentinel in `pre-push`
- keep branch deletion short-circuits working for SHA-256 repos
- add regression coverage for 64-zero delete lines in both `_run_ns()` and `hook_impl()`

Closes #3664.

## Validation
- baseline repro on current `main`: SHA-256 `pre-push` delete exited non-zero and raised `CalledProcessError` from `git diff ...<64 zeros>`
- `.venv/bin/python -m pytest tests/commands/hook_impl_test.py -k 'sha256_zero_oid' -q`
- `.venv/bin/python -m pytest tests/commands/hook_impl_test.py -k 'pre_push' -q`
- direct SHA-256 shell repro after fix: exit status `0`

## Gates
- `E1` baseline repro exists: yes
- `E2` new regression tests fail on baseline: yes
- `E3` new regression tests pass after fix: yes
- `E4` nearby targeted suite passes: yes
- `E5` diff localized to `pre_commit/commands/hook_impl.py` and `tests/commands/hook_impl_test.py`: yes

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -14,6 +14,10 @@ from pre_commit.store import Store
 Z40 = '0' * 40
 
 
+def _is_zero_oid(rev: str) -> bool:
+    return bool(rev) and set(rev) == {'0'}
+
+
 def _run_legacy(
         hook_type: str,
         hook_dir: str | None,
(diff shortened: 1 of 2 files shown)


asottile:
ai slop is not acceptable

PR #3674, opened 2026-04-28
fix: handle cross-drive ValueError in _adjust_args_and_chdir on Windows

terminalchai:
Fixes #2530

## Problem

On Windows, \os.path.relpath()\ raises \ValueError\ when the path and current working directory are on **different drives** (e.g. config lives on \C:\ but the repo is on \D:\):

\\\
ValueError: path is on mount 'C:', start on mount 'D:'
\\\

\_adjust_args_and_chdir\ calls \os.path.relpath\ unconditionally for four arguments — \config\, \iles\, \commit_msg_filename\, and \
epo\ — so any cross-drive combination crashes before a single hook runs.

## Fix

Add a small \_try_relpath()\ helper that wraps \os.path.relpath\ in a \	ry/except ValueError\ and falls back to the original absolute path when the drives differ:

\\\python
def _try_relpath(path: str) -> str:
    try:
        return os.path.relpath(path)
    except ValueError:
        return path
\\\

All four \os.path.relpath\ calls in \_adjust_args_and_chdir\ are updated to use this helper.

## Tests

Added 29 lines of tests in \	ests/main_test.py\ covering cross-drive scenarios for each of the four argument types.

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -172,6 +172,16 @@ def _add_run_options(parser: argparse.ArgumentParser) -> None:
     )
 
 
+def _try_relpath(path: str) -> str:
+    """Return os.path.relpath(path), or path unchanged on Windows when path
+    and the current working directory are on different drives (which would
+    raise ValueError: path is on mount 'C:', start on mount 'D:')."""
+    try:
+        return os.path.relpath(path)
+    except ValueError:
+        return path
+
+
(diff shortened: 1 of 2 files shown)


asottile:
https://github.com/pre-commit/pre-commit/issues/2530#issuecomment-4276626327

PR #3675, opened 2026-05-03
Add --quiet / -q flag to suppress output for passing and skipped hooks

### LINKED ISSUE #823: Question: Is there an option to only output on Failed hook

nitzan-shaked:
Closes #823.

## Summary

This adds a `--quiet` / `-q` flag to `pre-commit run` (and `try-repo`, which shares the same arg parser). In quiet mode:

- Per-hook output is **suppressed** for hooks that **pass** or are **skipped** (both `SKIP=` and no-files-to-check variants).
- **Failed** hooks still print their full output (hook name, exit code, stdout).
- Verbose detail lines (`- hook id:`, `- duration:`) are also suppressed for passing hooks when `--quiet` is set.

This is the minimal implementation discussed in the issue. The goal is to reduce noise in large monorepos where dozens of hooks are skipped on every commit.

## Changes

| File | What changed |
|---|---|
| `pre_commit/main.py` | Add `--quiet` / `-q` arg to `_add_run_options()` |
| `pre_commit/commands/run.py` | Propagate `quiet` through `_run_single_hook()`; suppress output for non-failing hooks |
| `pre_commit/commands/hook_impl.py` | Add `quiet=False` to the `_ns()` Namespace so the git-hook code path doesn't crash |
| `testing/util.py` | Add `quiet=False` to `run_opts()` helper |
| `tests/commands/run_test.py` | 5 new tests covering all quiet-mode cases |

## Test plan

```
pytest tests/commands/run_test.py -k "test_quiet"   # 5 new tests
pytest tests/commands/run_test.py                    # 102 total, 0 failures
pytest tests/commands/hook_impl_test.py              # 44 tests, 0 failures
```


asottile:
ai slop is not acceptable

PR #3676, opened 2026-05-04
Fix the docker language when using podman

swick:
Two changes:

* fix for the `podman system info` invocation which is required to have podman work when aliased to docker
* fallback from docker to podman for when podman is not aliased to docker (default on many distros)

### DIFF
--- a/pre_commit/languages/docker.py
+++ b/pre_commit/languages/docker.py
@@ -109,7 +109,7 @@ def install_environment(
 @functools.lru_cache(maxsize=1)
 def _is_rootless() -> bool:  # pragma: win32 no cover
     retcode, out, _ = cmd_output_b(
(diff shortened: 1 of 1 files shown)


asottile:
generally I just symlink an alternative container runtime to `docker`

swick:
You did not even take the time to read the comment.

asottile:
you did not even take the time to search the issue tracker

swick:
I certainly did, but you still didn't read it. Please take the 10 seconds it takes to actually read the 3 sentences I wrote. Literally less time than typing out a comment.

asottile:
I did read it.  and I read the whole changeset

and there are duplicates already discussing and denying what you changed

swick:
> I read the whole changeset

That is literally not true.

asottile:
promise I did.  good day

PR #3677, opened 2026-05-04
handle sha256 zero oid in pre-push

praneethhere:
Treat the SHA-256 zero object id as a deleted ref in pre-push

In SHA-256 repositories, Git sends 64 zeroes for a deleted ref. pre-commit was only checking the SHA-1 40-zero value, so branch deletion could continue into the diff logic

Fixes #3664

Tests:
- python -m pytest tests/commands/hook_impl_test.py -k "pre_push_deleting_branch or hook_impl_main_noop_pre_push" -q
- python -m pytest tests/commands/hook_impl_test.py -k pre_push -q
- python -m pytest tests/commands/hook_impl_test.py -q
- manual SHA-256 pre-push deletion repro

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -12,6 +12,8 @@ from pre_commit.parse_shebang import normalize_cmd
 from pre_commit.store import Store
 
 Z40 = '0' * 40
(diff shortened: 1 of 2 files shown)


asottile:
ai slop is not acceptable

PR #3680, opened 2026-05-10
Add opt-in hash-locked installs for Python hooks

### LINKED ISSUE #3679: [FR] Add opt-in hash-locked installs for `language: python` hooks

python_lockfile require-hashes additional_dependencies lockfile supply chain

I want to make `language: python` hook environments deterministic and hash-checked.

Today, Python hook environments are not fully deterministic from `.pre-commit-config.yaml`. Even when a hook `repo` and `rev` are pinned, Python dependencies can still be resolved at environment creation time. `additional_dependencies` can pin top-level packages, but it does not provide a complete, hash-checked dependency closure.

This matters more now because recent PyPI incidents show that legitimate package names can become temporarily unsafe after account, token, or release-pipeline compromise. For example, PyPI published an incident report for the LiteLLM/Telnyx supply-chain attacks in March 2026, and Sonatype reported malicious Lightning / PyTorch Lightning releases around April 30, 2026.

The broader ecosystem is also moving toward dependency cooldowns / minimum release age policies. pip now supports filtering candidates by upload time with `--uploaded-prior-to`, and uv supports dependency cooldowns via `exclude-newer`. Similar minimum-age settings are also appearing across JavaScript package managers such as npm, pnpm, and Yarn.

I would like to propose an opt-in feature for `language: python` hooks:

- Add `python_lockfile: <path>`
  - valid only for `language: python`
  - path is relative to `.pre-commit-config.yaml`
  - initially accepts a pip-compatible, hash-pinned requirements file
  - mutually exclusive with `additional_dependencies`
- pre-commit computes the SHA256 of the lockfile content internally
  - no user-provided hash field
  - the computed digest is included in the environment/cache key
- when creating the hook environment:
  - install the complete locked dependency set with hash enforcement:

    ```bash
    python -m pip install --require-hashes -r <python_lockfile>
    ```

  - then install the hook package itself without dynamic runtime or build dependency resolution:

    ```bash
    python -m pip install --no-deps --no-build-isolation .
    ```

  - the lockfile is expected to include both runtime dependencies and any build-system requirements needed to install the hook package with `--no-build-isolation`, such as `setuptools`, `wheel`, `hatchling`, or `flit-core`

The lockfile would be generated externally by the user, for example:

```bash
uv pip compile \
  --generate-hashes \
  --exclude-newer=P7D \
  -o pre-commit-lock.txt \
  requirements.in
```

or with pip-tools:

```bash
pip-compile \
  --generate-hashes \
  --allow-unsafe \
  --pip-args="--uploaded-prior-to=P7D" \
  -o pre-commit-lock.txt \
  requirements.in
```

The `--allow-unsafe` flag may be needed when the lockfile intentionally includes build-system requirements such as `setuptools`. In pip-tools terminology, “unsafe” refers to packaging/bootstrap tools such as `pip`, `setuptools`, and `distribute`; it does not mean the package is known to be malicious.

Example `requirements.in`:

```text
requests==2.33.1
setuptools
```

This would not change the default behavior. It would give users who care about reproducibility and supply-chain hardening a way to make Python hook environments deterministic and hash-checked.

This does not prevent every supply-chain attack: a bad lockfile still installs bad artifacts. The goal is narrower: make Python hook environments deterministic, hash-checked, and compatible with dependency-cooldown policies during lockfile generation.

References:

- PyPI incident report: LiteLLM/Telnyx supply-chain attacks: https://blog.pypi.org/posts/2026-04-02-incident-report-litellm-telnyx-supply-chain-attack/
- Sonatype report: malicious Lightning / PyTorch Lightning packages: https://www.sonatype.com/blog/malicious-pytorch-lightning-packages-found-on-pypi
- pip `--uploaded-prior-to`: https://pip.pypa.io/en/stable/cli/pip_install/
- uv dependency cooldowns / `exclude-newer`: https://docs.astral.sh/uv/concepts/resolution/
- Cross-ecosystem dependency cooldown overview: https://cooldowns.dev/

### pre-commit --version

pre-commit 4.6.0

bc-lee:
## Summary

Add opt-in `python_lockfile` support for `language: python` hooks.

When set, pre-commit installs Python hook environments by:

1. installing a pip-compatible, hash-pinned requirements file:
   `python -mpip install --require-hashes -r <python_lockfile>`
2. installing the hook package without dependency resolution:
   `python -mpip install --no-deps --no-build-isolation .`

The lockfile path is resolved relative to the config file, and its content SHA256 is included in the hook/cache identity. Default Python hook behavior is unchanged.

## Why

`additional_dependencies` is appended to the normal `pip install .` hook install. It can reference extra packages or a requirements file, but it does not provide the same behavior:

- lockfile contents are not part of the cache key
- the hook package can still trigger runtime/build dependency resolution
- pip hash-checking requires the full dependency closure to be pinned and hashed

This gives users an opt-in deterministic, hash-checked install path.

Fixes #3679

### DIFF
--- a/pre_commit/clientlib.py
+++ b/pre_commit/clientlib.py
@@ -404,6 +404,39 @@ class NotAllowed(cfgv.OptionalNoDefault):
             raise cfgv.ValidationError(f'{self.key!r} cannot be overridden')
 
 
+class PythonLockfile(NamedTuple):
+    key: str
+    default: str | None
+
+    def check(self, dct: dict[str, Any]) -> None:
+        if self.key not in dct:
+            return
+
+        with cfgv.validate_context(f'At key: {self.key}'):
+            cfgv.check_string(dct[self.key])
+
+        if dct.get('additional_dependencies'):
+            raise cfgv.ValidationError(
+                f'{self.key!r} cannot be combined with '
+                f"'additional_dependencies'",
+            )
+
+        if (
+                'language' in dct and
+                _translate_language(dct['language']) != 'python'
+        ):
+            raise cfgv.ValidationError(
+                f'{self.key!r} is only supported for language: python',
+            )
+
+    def apply_default(self, dct: dict[str, Any]) -> None:
+        if self.default is not None:
+            dct.setdefault(self.key, self.default)
+
+    def remove_default(self, dct: dict[str, Any]) -> None:
+        raise NotImplementedError
+
+
 _COMMON_HOOK_WARNINGS = (
     OptionalSensibleRegexAtHook('files', cfgv.check_string),
     OptionalSensibleRegexAtHook('exclude', cfgv.check_string),
@@ -454,12 +487,14 @@ CONFIG_HOOK_DICT = cfgv.Map(
     ),
     StagesMigrationNoDefault('stages', []),
     LanguageMigration('language', cfgv.check_one_of(language_names)),  # remove
+    PythonLockfile('python_lockfile', None),
     *_COMMON_HOOK_WARNINGS,
 )
 LOCAL_HOOK_DICT = cfgv.Map(
     'Hook', 'id',
 
     *MANIFEST_HOOK_DICT.items,
+    PythonLockfile('python_lockfile', ''),
     *_COMMON_HOOK_WARNINGS,
 )
 CONFIG_REPO_DICT = cfgv.Map(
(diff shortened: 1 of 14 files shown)


bc-lee:
Just leaving a short comment and closing the issue does not seem right to me, even though this is your repository. I believe this is a legitimate use case, and I hope we can discuss it here.

Issue #3687, opened 2026-05-15
Tests failures on macOS (26.5 tested, may affect others)


alsutton:
### search you tried in the issue tracker

macOS test

### describe your issue

**macOS 26.5 Tracker Issue for future PRs to reference**

After checking out the code, and running the tests, on a macOS 26.5 machine with Apple Command Line Tools for Xcode 26.5 and Python 3.14 from Homebrew, the in-repo test failure rate is around 4.5%. The following is the output from an environment setup and test run;

Full test output is at https://gist.github.com/alsutton/d959aede81742c41b92871bb89016b05

### pre-commit --version

commit edad96821ac495ce2a6e4fde26dd4e0b6916d72d

### .pre-commit-config.yaml

```yaml
Unmodified from commit edad96821ac495ce2a6e4fde26dd4e0b6916d72d in this repo
```

### ~/.cache/pre-commit/pre-commit.log (if present)

pytest execution did not provide a log.

alsutton:
Several of these are just missing asserts in the tests, where the process execution fails due to the required executable not being installed.

asottile:
I intentionally don't support other testing scenarios that are different from ci -- including macos.  there's several duplicates which outline this

PR #3689, opened 2026-05-15
Use a permanent directory for RUSTUP_HOME

alsutton:
`rustup` installs the toolchain in RUSTUP_HOME, so we need it to be a permanent directory, and not a temporary one, to ensure that the toolchain isn't deleted after it is installed.

### DIFF
--- a/pre_commit/languages/rust.py
+++ b/pre_commit/languages/rust.py
@@ -82,7 +82,7 @@ def _add_dependencies(
 
 def install_rust_with_toolchain(toolchain: str, envdir: str) -> None:
     with tempfile.TemporaryDirectory() as rustup_dir:
-        with envcontext((('CARGO_HOME', envdir), ('RUSTUP_HOME', rustup_dir))):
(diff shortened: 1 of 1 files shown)


asottile:
no tests failed before this or were changed

the tool chain isn't intended to be persisted only used as a builder

alsutton:
No tests failed because, as you explained in https://github.com/pre-commit/pre-commit/issues/3687 , you don't actually test for the scenario.

I know this may not land well with you, but I'm a bit disappointed that you're not open to improving situations under which pre-commit gets tested and works, but I appreciate you being fast with your response, which gives a clear signal of what your priorities are.

asottile:
if you read the duplicate you'd find my explanation about not wanting to boil the ocean exhausting every scenario possible (coupled with the extreme difficulty of keeping up with the shifting sands of macos completely) 

the scenario is tested.  you deleted the assert for it in the other branch (and coverage remains at 100%)

PR #3690, opened 2026-05-16
Fix config path handling across Windows drives

hansu650:
Fixes #2530.

## Summary

This fixes a Windows path handling error when `--config` points to a file on a different drive from the repository root.

`os.path.relpath` raises `ValueError` for paths on different Windows drives, so this keeps the original absolute path when a relative path cannot be computed.

## Tests

Added a Windows regression test for `--config` on another drive.

Validation run locally:

```text
python -m pytest tests/main_test.py -k different_drive
python -m pytest tests/main_test.py -k adjust_args_and_chdir
python -m pytest tests/main_test.py
python -m py_compile pre_commit/main.py tests/main_test.py
git diff --check

### DIFF
--- a/pre_commit/main.py
+++ b/pre_commit/main.py
@@ -47,6 +47,13 @@ COMMANDS_NO_GIT = {
 }
 
 
+def _relpath(path: str) -> str:
+    try:
+        return os.path.relpath(path)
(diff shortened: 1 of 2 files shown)


asottile:
ai slop is not acceptable

PR #3691, opened 2026-05-16
fix: handle SHA-256 zero OIDs in pre-push hook

karlhillx:
## Fix SHA-256 zero OID handling in pre-push hook

In repositories initialized with `--object-format=sha256`, git's pre-push hook protocol emits 64-character zero OIDs instead of the SHA-1 40-character variety. The previous hardcoded check

```python
Z40 = '0' * 40
if local_sha == Z40:
```

never matched in SHA-256 repos, causing `git push --delete` to fail with:

```
fatal: Invalid symmetric difference expression <sha>...<64-zeros>
```

Add `_is_zero_oid()` as a hash-format-agnostic helper that accepts zero OIDs of either supported length (40 or 64), replacing the hardcoded `Z40` constant in `_pre_push_ns`.

Also adds `test_run_ns_pre_push_deleting_branch_sha256` as a regression test for the SHA-256 deletion push case.

Fixes #3664

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -11,6 +11,12 @@ from pre_commit.envcontext import envcontext
 from pre_commit.parse_shebang import normalize_cmd
 from pre_commit.store import Store
 
+
+def _is_zero_oid(oid: str) -> bool:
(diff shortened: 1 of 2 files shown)


asottile:
sorry we didn't accept ai contributions

PR #3694, opened 2026-05-18
support sha256 zero oid in pre-push hook

D0NMEGA:
Resolves #3664.

The pre-push hook compared `local_sha` and `remote_sha` against a hardcoded 40-char Z40 sentinel. On `sha256` object-format repositories git's pre-push protocol emits 64-char zero oids instead, so the deletion-only and new-branch short-circuits never fired. `git push origin --delete <branch>` fell through to a `git diff <old>...<64-zeros>` invocation that produced `fatal: Invalid revision range` and pre-commit exited non-zero.

Replace the equality comparisons with an `_is_zero_oid` predicate that accepts all-zero oids of either supported length (40 or 64). `Z40` stays around for backward compatibility with the existing tests.

Reproducer from the issue body confirmed locally on git 2.53; added `test_run_ns_pre_push_sha256_deleting_branch` which fails on `main` and passes with this change. Also added `test_run_ns_pre_push_sha256_new_branch` for forward compatibility on the creation path. Unit tests for the new `_is_zero_oid` predicate cover both supported lengths and a parametrized set of rejection cases. The new SHA-256 fixture skips with `pytest.skip` if the running git lacks `--object-format=sha256` support, keeping the test backward compatible with git < 2.29.

AI assistance (Claude) was used while drafting parts of this change. The diff was reviewed manually before submission.

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -14,6 +14,14 @@ from pre_commit.store import Store
 Z40 = '0' * 40
 
 
+def _is_zero_oid(oid: str) -> bool:
+    # git's pre-push protocol uses an all-zero object id to indicate
+    # branch creation (old) or deletion (new).  the length depends on the
+    # repository's object format: 40 for sha1 (default) and 64 for sha256
+    # (`git init --object-format=sha256`).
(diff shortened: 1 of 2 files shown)


asottile:
we don't accept ai slop contributions

Issue #3695, opened 2026-05-18
more clarity for `pre-commit autoupdate --repos`


mxr:
### search you tried in the issue tracker

autoupdate silent

### describe your actual problem

`pre-commit autoupdate --repos` matches against the passed-in repos here:
https://github.com/pre-commit/pre-commit/blob/edad96821ac495ce2a6e4fde26dd4e0b6916d72d/pre_commit/commands/autoupdate.py#L190

If no repos match, then no futures are created, so the command exits silently.

```console
$ pre-commit autoupdate --repo https://github.com/pre-commit/pre-comimt-hooks
# exits with 0 and prints nothing
```

Wondering if it would be clearer to log which passed-in repos don't match any repos in the pre-commit config to make this easier to debug.

### pre-commit --version

pre-commit 4.6.0

asottile:
it might (?) be better -- though the usual program usually outputs _something_ on success anyway

mxr:
To be precise it prints out a log line for each future only. 

```console
$ pre-commit autoupdate --freeze --repo https://github.com/zizmorcore/zizmor-pre-commit --repo https://github.com/sqlfluff/sqlfluff --repo https://github.com/pre-commit/pre-commit-hooks --repo doesnotexist
[https://github.com/pre-commit/pre-commit-hooks] already up to date!
[https://github.com/sqlfluff/sqlfluff] updating ec5ed61d83ae336c65c4c15396a23772ce9d2fcc -> 4.2.1 (frozen)
[https://github.com/zizmorcore/zizmor-pre-commit] updating a4727cbbcd26d7098e96b9cb738169b59711ae51 -> v1.25.2 (frozen)
```

(doesnotexist has no output since no future is created)

asottile:
slept on this and I think the following should happen (yeah technically a breaking change but tbh it's just an oversight that this was allowed to pass)

any specified `--repo` that doesn't match any in the config should be an error

mxr:
sg I opened a PR to log a warning (in the short-term) and another PR to return an error (for 6-12 months later?)

asottile:
I'm fine to just go straight to error

mxr:
sure

asottile:
via #3701

PR #3701, opened 2026-05-28
Return an error for invalid --repo

mxr:
Related to #3695

### DIFF
--- a/tests/commands/autoupdate_test.py
+++ b/tests/commands/autoupdate_test.py
@@ -260,8 +260,8 @@ def test_autoupdate_out_of_date_repo_with_correct_repo_name(
     assert 'local' in after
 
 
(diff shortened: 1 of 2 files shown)


tests/commands/autoupdate_test.py
```diff
@@ -260,7 +260,7 @@ def test_autoupdate_out_of_date_repo_with_correct_repo_name(
     assert 'local' in after
 
 
-def test_autoupdate_out_of_date_repo_with_wrong_repo_name(
+def test_autoupdate_missing_repo_name(
```
asottile, line 263: this test should validate the output of the new error message as well

mxr: Sure

asottile (approved):

PR #3710, opened 2026-06-15
feat: Add `--no-fail-fast` CLI option to override fail-fast behavior

anthonyhaussman:
Introduce a mutually exclusive `--no-fail-fast` CLI argument alongside `--fail-fast`. 

This option allows disabling fail-fast behavior even if it has been enabled globally in the config or at the individual hook level. 
Also update hook-impl to safely support the no_fail_fast attribute.

### DIFF
--- a/pre_commit/commands/hook_impl.py
+++ b/pre_commit/commands/hook_impl.py
@@ -110,6 +110,7 @@ def _ns(
         verbose=False,
         show_diff_on_failure=False,
         fail_fast=False,
+        no_fail_fast=False,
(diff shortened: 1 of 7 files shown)


asottile:
I don't think this is a good idea.  if it's misconfigured just fix the config

anthonyhaussman:
Well, the goal was to enable `fail-fast` in the configuration file for local execution, but run the CI steps with `--no-fail-fast`.

But I'll find another way. 🙏

PR #3711, opened 2026-06-16
set `JULIA_DEPOT_PATH`

damonbayer:
This PR pins Julia's package depot to the per-hook environment directory by setting
`JULIA_DEPOT_PATH` in `get_env_patch`, instead of leaving it at its default (`~/.julia`).

I believe this is required for `julia` to fully qualify as a second-class language:
https://github.com/pre-commit/pre-commit/blob/1553b465fd7ea42321ae0d04d1b41e706b89ae45/CONTRIBUTING.md#L68-L70

This surfaced in a review from @asottile of https://github.com/pre-commit-ci/runner-image/pull/335.

Details about `JULIA_DEPOT_PATH` are available [here](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH).


damonbayer:
@asottile are you able to review?

asottile (approved):

PR #3714, opened 2026-06-24
fix: fix capitalization and grammar in CONTRIBUTING.md

Dev9269:
Fixed sentence starting with lowercase 'here are the apis' to proper capitalization 'Here are the APIs'.


asottile:
what a waste of CO2 and tokens

PR #3725, opened 2026-07-21
Don't use identify if not necessary

vries:
If we're not filtering on types, there's no need to call identify.


asottile:
seems like unnecessary complexity and unlikely to improve anything.  please start with an issue explaining next time

vries:
> seems like unnecessary complexity and unlikely to improve anything. please start with an issue explaining next time

I've filed an issue ( https://github.com/pre-commit/pre-commit/issues/3729 ).

Issue #2366, opened 2022-04-27
Autoupdate inconsistency with one repo


tspearconquest:
### describe your issue

2 days ago when I ran `pre-commit autoupdate` on one of my projects, the pre-commit hook repo `https://github.com/crate-ci/typos` was updated to a `rev` string with a completely different pattern than what we had been using

We were using `v1.5.0` and attempting to upgrade to `v1.7.1`, but pre-commit picked `typos-v0.9.1`.
Today when I ran `pre-commit autoupdate` on a different project, the same repo was updated to `typos-dict-v0.8.1`.

The `v1.7.1`, `typos-v0.9.1`, and `typos-dict-v0.8.1` releases were made on the same date and have the same commit sha, but could have different contents depending on how the CI that built it was designed. For my case, all 3 releases appear to be identical, but this could be used in malicious ways because of the unpredictability.

If I picked `v1.5.0` as my `rev` string, I don't want `pre-commit autoupdate` to change the format of that (just update to the highest version which matches the same pattern), unless the owners of the hook repo tell pre-commit they have changed the format of the version string intentionally somehow, or they decide to stop publishing updates for an old format and tell users to manually change to the new format in order to continue receiving updates to the hook. In other words, changing version string pattern should be a breaking change and users should have notice about it. It should not be changed arbitrarily by `pre-commit autoupdate`.

The repo I mentioned above published 3 releases on the same day with the same commit hash but different names, and the official Github Release version is v1.7.1 as evidenced here: https://api.github.com/repos/crate-ci/typos/releases/latest so my goal with this issue is to get pre-commit to stick with the format already in use in my local `.pre-commit-config.yml` or at the very least check against the official release API from Github for the right version string to use when updating the hooks.

### pre-commit --version

pre-commit 2.18.1

### .pre-commit-config.yaml

```yaml
repos:
- repo: https://github.com/crate-ci/typos
  rev: v1.5.0
  hooks:
  - id: typos
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v4.1.0
  hooks:
  - id: check-added-large-files
  - id: check-shebang-scripts-are-executable
  - id: check-json
  - id: pretty-format-json
    args:
    - --autofix
  - id: check-merge-conflict
  - id: check-symlinks
  - id: destroyed-symlinks
  - id: check-yaml
  - id: detect-private-key
  - id: trailing-whitespace
  - id: end-of-file-fixer
  - id: check-docstring-first
[...]

PR #3732, opened 2026-08-02
Choose a semver tag over the last tag in a repo

pjaselin:
Problem: I'm (and it looks like others) are seeing this pop up with pre-commit hooks coming from repos like: https://github.com/crate-ci/typos where a tag like "v1" is maintained for the latest release along with a semver tag like "v1.Y.Z". When running `pre-commit autoupdate`, you always get the "v1" tag instead of the latest "v1.Y.Z" tag. This is reported with a workaround here: https://github.com/crate-ci/typos/issues/390 but it seems like the fix should actually be at the pre-commit level and not to go around mirroring repos to get a clean set of tags.

Solution: I believe the intention of the git.get_best_candidate_tag function is to choose a semver-compliant tag instead of necessarily the most recent one, where these tags may be off the same commit SHA. This PR gets the commit for the last found tag, then gets all tags associated with that commit SHA, and finally allows for returning that commit if there's a period in it. I've added tests for the git and autoupdate module to support the changes.

Edit: I'm reading your comments in #2366 and I'm seeing how there's a tradeoff between repeatability here in getting the right tags so this might not be an improvement even if it appears to fix what I've experienced. Happy to close this if so and take it as an opportunity to learn about the internals of pre-commit.

### DIFF
--- a/pre_commit/git.py
+++ b/pre_commit/git.py
@@ -236,8 +236,11 @@ def get_best_candidate_tag(rev: str, git_repo: str) -> str:
     Multiple tags can exist on a SHA. Sometimes a moving tag is attached
     to a version tag. Try to pick the tag that looks like a version.
     """
+    commit = cmd_output(
+        'git', 'rev-list', '-n', '1', rev, cwd=git_repo,
+    )[1].strip()
     tags = cmd_output(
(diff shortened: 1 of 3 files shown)


pre_commit/git.py
pjaselin, line 1: Does NO_FS_MONITOR need to be in L240?

asottile:
search the issue tracker next time

PR #3745, opened 2026-08-15
Update virtualenv to min. 20.36.1

nkaaf:
This version update fixes the disclosed TOCTOU vulnerability of `virtualenv` (https://github.com/pypa/virtualenv/security/advisories/GHSA-597g-3phw-6986).


asottile:
pre-commit works with any version above the one listed.  there's no reason to change the minimum

PR #3754, opened 2026-09-09
add `--keep` to `pre-commit gc`

kristofferkoch:
this avoids cache churn when CI jobs use different revisions of the same repo, e.g. with PRs lagging a updated main branch.

referenced revisions are always kept and count toward the limit.

This creates a new table, so I'm not sure if this is OK, but it's a feature that would make my life better :)


asottile:
I suspect you're using your cache incorrectly.  please start with an issue

